PR #231, opened 2014-11-09
TAJO-838: Improve query planner to utilize index 

jihoonson:
In this work, the previous work is extended to support indexes for multiple columns or even expressions. Even though we don't provide any methods to estimate the query selectivity, users can use the index scan by configuring the session variables. Finally, I have added more unit tests including performing join, aggregation, sort after index scans.

As summarized below, some necessary features are still lacked in this work.
However, IMHO, this work is ready to be merged into master branch as a prototype of query execution using index. Please leave any comments.
Following features should be added.
* Query optimization based on selectivity estimation
* Support index for partitioned tables


jihoonson:
Updated against the master branch.

jihoonson:
Updated against the master branch.

hyunsik:
+1
It seems to be ready to be committed to index_support branch. Even through the test results in two failures, I don't think that they are not related to your patch. They are related to some blocker issues TAJO-1282 and TAJO-1278. They will be resolved in other Jiras.

jihoonson:
@hyunsik thanks for the review!

There still remains many things for efficient query processing with indexes. Here are the significant issues.
1. Printing appropriate messages after creating an index. If the index is successfully created, cli should show a success message. Otherwise, it should show a failure message. 
2. Index metadata backup and restore. In order to support the index restore without recreation, we need to provide a new DDL statement, ```CREATE INDEX [name] ON [path]```.
3. Index support for partitioned tables.
4. New index structure.
5. Other experimental issues, such as prefetching indexes, for the performance improvement.

IMHO, after resolving the issues 1 and 2, it would be sufficient to be merged into the master branch.
What do you think about it?

hyunsik:
@jihoonson ,

I agree with your plan. Even through this feature is still experimental, it seems to be ready to merge this feature to master branch due to maintenance cost. It also works well for CSV files.

jihoonson:
@hyunsik thanks for your agreement.
I'll implement remaining things according to the plan.

PR #246, opened 2014-11-15
TAJO-1180: digitValue should throw Exception when char is not in valid range

charsyam:
digitValue in TextUtils
it returns numeric value for hex data.
and. it should handle [0-9a-zA-Z].
but it returns some digit for other ranges.
so. I think it is better throw exception if it is not in valid range.


eminency:
+1

I think it looks good for detecting unexpected error.

hyunsik:
+1

PR #251, opened 2014-11-18
TAJO-1033: Implement a FileAppender for ElasticSearch.

blrunner:
I implemente ElasticSearchAppender. You can load tajo table data to elasticsearch index as follows:

 1. Create an elasticsearch index and index type.
ex) index: radio, index type: artitists

 2. Create a tajo for match to an elasticsearch index. For reference, you must use elasticsearch for storage type.
ex) CREATE TABLE estable1 (id int, name text, score float, type text)
USING elasticsearch
with ('elasticsearch.cluster'='elasticsearch',
'elasticsearch.nodes'='localhost:9300',
'elasticsearch.resources'='radio/artists',
'elasticsearch.replication'='1',
'elasticsearch.bulk.item.size'='1000');

 3. Load tajo data to an elasticsearch index with sql.
ex) insert overwrite into estable1 select * from table1;

This patch is not complete.I'll write a documentation and implement a few methods to improve performance. Also we need to allow for users to change last task numbers.This is because lots of tasks for storage will cause low performance of an elasticsearch cluster.


blrunner:
This issue relates to as following issue:
https://issues.apache.org/jira/browse/TAJO-1404

Watching TAJO-1404, I'll start it over again.

PR #301, opened 2014-12-15
TAJO-1226: add json and parquet validator

charsyam:
Result of this patch.

```c
default> create external table table1 (       id int,       name text,       score float,       type text, mytime date)       using parquet location 'file:/Users/charsyam/tajo/table';
ERROR: PARQUET doesn't support mytime:DATE
```

### DIFF
--- a/tajo-plan/src/main/java/org/apache/tajo/plan/DataTypeValidatorFactory.java
+++ b/tajo-plan/src/main/java/org/apache/tajo/plan/DataTypeValidatorFactory.java
@@ -1,0 +1,46 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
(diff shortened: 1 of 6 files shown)


hyunsik:
I had have thought of some design consideration. Our direction is to make storages pluggable. BTW, in this implementation, type validator is implemented separately from storage format implementation. Especially, type validator is in tajo-core. It would be better if validation and storage formation implementations are located in some same place.

The way is to implement some class or interface to give file format properties. But, it causes some breaking changes and takes more time. So, I think that your patch is a nice approach right now. 

Also, I'll create an issue to improve it to be on the our roadmap.

hyunsik:
Here is my +1. I'll commit it shortly.

hyunsik:
This patch creates new package 'plan/validator'. But, it's duplicated to 'plan/verifier'. So, I moved the classes to 'plan/verifier'.

PR #307, opened 2014-12-18
TAJO-1257: ORDER BY with NULL FIRST misses some data

sirpkt:
I changed setMaxRangeIfNull() method to set NULL as minimum value when null first is set


hyunsik:
I'm sorry for late review. I tested various cases. It seems to work correctly. 

BTW, I found another bug, I don't think that your patch causes this problem. But, you may have interested in this problem.

https://issues.apache.org/jira/browse/TAJO-1283

I'll finish your review by tomorrow.

hyunsik:
@sirpkt 

+1 LGTM

I'm sorry for late review. I missed the review for few days. Ship it. Thank you for your contribution.

sirpkt:
Thank you for the review, @hyunsik .
I just committed the patch.

PR #315, opened 2014-12-22
TAJO-1265: min(), max() does not handle null properly

sirpkt:
min() and max() handle null value separately from non-null values.
It computes min() or max() among non-null values first.
If no non-null value exists, they return null value.
Implementation style of min() and max() is also changed that 
there are abstract classes of Min and Max common for all the data types
and only type specific methods are implemented for each type.


sirpkt:
I also modified sum() and avg() to handle null properly.

hyunsik:
I'll review it tonight. Thank you for your contribution.

hyunsik:
The patch looks good to me. It corrects the wrong implementation. There are some trivial things that I need to check. I'll finish the review by tomorrow.

tajo-core/src/main/java/org/apache/tajo/engine/function/builtin/AvgDouble.java
```diff
@@ -18,6 +18,7 @@
 
 package org.apache.tajo.engine.function.builtin;
 
+import org.apache.commons.lang.ObjectUtils;
```
hyunsik: Unused import

tajo-core/src/main/java/org/apache/tajo/engine/function/builtin/AvgLong.java
```diff
@@ -18,6 +18,7 @@
 
 package org.apache.tajo.engine.function.builtin;
 
+import org.apache.commons.lang.ObjectUtils;
```
hyunsik: unused import

tajo-core/src/main/java/org/apache/tajo/engine/function/builtin/SumDouble.java
```diff
@@ -25,6 +25,7 @@
 import org.apache.tajo.datum.Datum;
 import org.apache.tajo.datum.DatumFactory;
 import org.apache.tajo.datum.Float8Datum;
```
hyunsik: unused import

tajo-core/src/main/java/org/apache/tajo/engine/function/builtin/SumLong.java
```diff
@@ -18,6 +18,7 @@
 
 package org.apache.tajo.engine.function.builtin;
 
+import org.apache.commons.math.stat.descriptive.summary.Sum;
```
hyunsik: unused import

hyunsik:
+1
Thank you for finding these bugs. It seems to correct wrong null handling behavior of aggregation operators.

hyunsik:
Hi @sirpkt,

I leave some trivial comments. Because they are trivial, you can commit after you remove them.

sirpkt:
I just committed the patch.

PR #317, opened 2014-12-24
TAJO-1228: TajoClient should communicate with only TajoMaster without TajoWorker.

hyunsik:
See the description at https://issues.apache.org/jira/browse/TAJO-1228.


blrunner:
+1

I tested the patch on tpc-h cluster. Queries ran successfully and codes look good to me.

PR #324, opened 2014-12-31
TAJO-1274: Merge separate pages of getting started document into a singl...

hyunsik:
...e page.

Please see the issue https://issues.apache.org/jira/browse/TAJO-1274.
I've merged the multiple small pages into a single page. 
Here is the generated page. http://people.apache.org/~hyunsik/TAJO-1274/getting_started.html


jihoonson:
+1

PR #325, opened 2015-01-01
TAJO-967: Don't store exit command in history

gyias:


jihoonson:
+1
thanks for your contribution!

PR #327, opened 2015-01-02
TAJO-1275: Optimizer pushs down non-equi filter as theta join qualifier

sirpkt:
This patch prevents non-equi filter condition becomes join qualifier of child nodes.
It can avoid theta join is created as a result of filter push down.

in FilterPushDownRule.java, visitJoin() additionally checks all the join qualifier candidates from pushing down filters whether it is non-equi join qualifier or not.


tajo-core/src/test/java/org/apache/tajo/engine/planner/TestLogicalPlanner.java
```diff
@@ -644,7 +644,7 @@ public final void testJoinWithMultipleJoinQual4() throws IOException, PlanningEx
         , new FieldEval(new Column("default.t.n_nationkey", Type.INT4))
         , new FieldEval(new Column("default.s.s_suppkey", Type.INT4))
     );
-    joinQualMap.put(joinQual, Boolean.FALSE);
+    //joinQualMap.put(joinQual, Boolean.FALSE);
```
jihoonson: It would be better if you note that this line must be restored when https://issues.apache.org/jira/browse/TAJO-742 is resolved.
The ```TODO``` annotation will be helpful for the future restoration.

tajo-plan/src/main/java/org/apache/tajo/plan/rewrite/rules/FilterPushDownRule.java
```diff
@@ -210,6 +210,20 @@ public LogicalNode visitJoin(FilterPushDownContext context, LogicalPlan plan, Lo
       }
     }
 
+      // non-equi filter should not be push down as a join qualifier until theta join is implemented
```
jihoonson: It would be better if you note that this line must be restored when https://issues.apache.org/jira/browse/TAJO-742 is resolved.
The ```TODO``` annotation will be helpful for the future restoration.

jihoonson:
Hi @sirpkt
This patch looks good to me.
+1 and thanks for your contribution.

sirpkt:
Thank you for the review, @jihoonson.
I added the comments about TODO and just committed the patch.

PR #329, opened 2015-01-05
TAJO-1271: Improve memory usage in HashShuffleFileWriteExec.

jinossy:


jinossy:
This patch just improve the memory usage. 
So I will improve the tuple-copy in new jira issue

jinossy:
This patch is not yet ready for review

PR #330, opened 2015-01-06
TAJO-1225: Fix wrong schema name in JDBC driver

blrunner:
This patch fixes wrong default schema name used in JDBC driver. So far, we have used 'public' in a default schema name. This patch will change it to an empty string (i.e., "");


hyunsik:
Could you trigger TravisCI?

blrunner:
Hi @hyunsik 

I've just merged the master branch to TAJO-1225 branch. :)

hyunsik:
+1
Thanks. ship it!

PR #331, opened 2015-01-06
TAJO-1279: Cleanup TajoAsyncDispatcher and interrupt stop events.

jinossy:


jihoonson:
I'm reviewing the patch.

jihoonson:
@hyunsik thanks for your review. I also think that we should resolve this issue as soon as possible. 

All changes look good to me, too. The only thing I concern is, as commented above, the frequent hangs during unit tests. Whenever a zookeeper client fails to establish a connection, the test waits for 10 ~ 50 seconds. I think that it will significantly bother us.

I cannot find this error in the current master branch. Thus, I suspect that an existing, but hidden error is exposed due to the changed HBase-related configurations. Maybe we can fix it in another issue.

What do you think about it?

jinossy:
@jihoonson I will remove the HBase-related codes

```
+    conf.setInt(HConstants.HBASE_CLIENT_IPC_POOL_SIZE, 5);
+    conf.setInt(HConstants.REGION_SERVER_HANDLER_COUNT, 5);
+    conf.setInt(HConstants.REGION_SERVER_META_HANDLER_COUNT, 2);
+    conf.setInt(HConstants.MASTER_HANDLER_COUNT, 5);
+    conf.setInt("hbase.hconnection.threads.max", 5);
+    conf.setInt("hbase.hconnection.threads.core", 5);
```

jihoonson:
@jinossy thanks. We can commit the patch immediately. 
Anyway, we must resolve the zookeeper-related problem in another issue.

jinossy:
@jihoonson 
I don’t think so. Hbase was merge by TAJO-1233. and This patch is not related miniZookeeperCluster binding issue.
Have your problem is in this patch ?

jihoonson:
@jinossy, I have tested the latest patch. It worked well.
Here is my +1.
Thanks!

jinossy:
@jihoonson Thanks! 
I'll commit it shortly.

PR #332, opened 2015-01-06
TAJO-1280: Update the roles of Hyoungjun and Jihun in web site.

hyunsik:


hyunsik:
Thank you for the quick review. I'll commit it soon.

PR #335, opened 2015-01-07
TAJO-1260: Add ALTER TABLE ADD/DROP PARTITION statement to parser

blrunner:
I plan to implement ALTER TABLE ADD/DROP PARTITION statement stage by stage. For the first time, I'll add the statement to parser.

* Syntax

ALTER TABLE table_name [IF NOT EXISTS] ADD PARTITION (key1 = 'val2', key2 = 'val2', ...) [LOCATION 'partition_path_of_directory'];

ALTER TABLE table_name [IF NOT EXISTS] DROP PARTITION (key1 = 'val2', key2 = 'val2', ...);


tajo-core/src/test/java/org/apache/tajo/engine/parser/TestSQLAnalyzer.java
```diff
@@ -367,6 +364,59 @@ public void testCreateTablePartitionByColumn() throws IOException {
   }
 
   @Test
+  public void testAlterTableAddPartition1() throws IOException {
+    String sql = FileUtil.readTextFile(new File("src/test/resources/queries/default/alter_table_add_partition_1.sql"));
```
hyunsik: could you use FileUtil.readTextFileFromResource instead of direct paths? Direct path can cause errors without setting a base directory.

blrunner:
Hi @hyunsik 

Thanks for your quick review. I used FileUtil.readTextFileFromResource instead of direct paths and I added more unit test cases.

hyunsik:
Could you trigger the travisCI job?

blrunner:
Hi @hyunsik 

I've just merged the master branch to TAJO-1260 branch.

hyunsik:
+1
The patch looks good to me. 

BTW, you need to consider inequality expression in DROP PARTITION. But, the current patch is good starting point. We can improve it later in another jira.
```sql
ALTER TABLE table_name [IF EXISTS] DROP PARTITION  (key1 [=|<|<=|>|>=|!=] 'val1', key2 ...,);
```

hyunsik:
``ALTER TABLE .. [ADD | DROP] PARTITION`` statement will cause error if this statement is given in JDBC driver or tsql. So, could you add some exception handling code to prevent ugly errors?

blrunner:
Hi @hyunsik 

I added the exception handling code. If possible, could you check the pathch again?

hyunsik:
Sure, I'll review it soon.

hyunsik:
@blrunner Could you trigger TravisCI?

blrunner:
@hyunsik 

Sure, I've just triggered it. :)

hyunsik:
+1 ship it. Thanks.

It should be committed to only master branch.

PR #336, opened 2015-01-07
TAJO-1285 Refactoring Magic Number to HAConstants

charsyam:
HAConstants defines types.
But in HAServiceUtil, just using Magic Numbers 1,2,3,4,5
so this patch just changed them as HAConstants.


blrunner:
Hi @charsyam 

+1
Thanks for your contribution.

PR #337, opened 2015-01-08
Tajo-1287: Repeated using of the same order by key in multiple window clauses should be supported

sirpkt:
I fixed a bug in WindowAggExec by removing duplicated Columns in sort column list.

"mvn clean install" is checked.


hyunsik:
The same workaround is applied to groupby and order-by planning. I agree with this bug fix. I'll finish the review after the TravisCI gives green sign.

hyunsik:
+1 ship it!

sirpkt:
Thank you for the fast review, @hyunsik.
I just committed the patch.

PR #339, opened 2015-01-08
TAJO-1251: Query is hanging occasionally by shuffle report.

jinossy:


hyunsik:
I have some suggestion:
 * StageState.FINALIZE should be StageState.FINALIZING because it's a continuing state while receiving shuffle reports.
 * lastContactTime does not need to be AtomicLong because lastContactTime update will be serialized by event handler.

hyunsik:
Others all look good to me.

hyunsik:
The event name ``SQ_STAGE_FINALIZE`` is not intuitive regarding what this event does. How about ``SQ_SHUFFLE_REPORT`` instead of ``SQ_STAGE_FINALIZE``?

jinossy:
Thank you for your suggestion.
I will update the patch.

PR #341, opened 2015-01-09
TAJO-1286: Remove netty dependency from tajo-jdbc

ykrips:
Though JdbcConnection class does not use parameters on its class, this QueryStringDecoder class was implemented for future use.


hyunsik:
+1 ship it. The patch looks good to me. 

Even though TestTajoClient.testKillQuery causes one test failures, it is not related to this issue. Actually, I'm digging this problem. I'll fix this problem in another jira. Your patch seems to work correctly.

ykrips:
Hello @hyunsik,
Thank you for reviewing this issue. This pull request will be merged into master branch.

PR #345, opened 2015-01-09
TAJO-1295: Remove legacy worker.dataserver package and its unit tests.

hyunsik:


jihoonson:
Would you check some failures during tests?

hyunsik:
I believe that the failure is a occasionally happened problem that we are still finding in another jira. It seems to be not related to my patch because all unit tests are passed in my machine. So, I just rebased and trigger the TravisCI.

jihoonson:
+1
You are right. Ship it!

PR #347, opened 2015-01-10
TAJO-1297: Tajo Web UI does not work after TAJO-1291

jihoonson:


hyunsik:
+1 LGTM

jihoonson:
Thanks for the quick review!

PR #349, opened 2015-01-14
TAJO-1304: Can not found TextFile in catalog.

jinossy:


hyunsik:
+1

The patch looks good to me.

PR #351, opened 2015-01-14
TAJO-919: Implement LAG and LEAD window functions

sirpkt:
This patch contains following changes
- lead, lag are implemented. lag is implemented as window function as it only sees previous rows while lead is implemented as aggregation function since it should see next rows.
- visitWindowFunction() in ExprAnnotator.java and TypeDeterminant.java is modified to support multiple arguments because current implementation only supports one argument for window function.

There are some arguable points 
- It assumes that offset value is non-negative.
- Way to handle aggregation function in WindowAggExec.java is modified to call terminate for every Tuple, and it provides way to return different value for each terminate call, which is essential to implement lead function within current window supporting structure. However, it adds additional burden for other aggregation functions calculation because their terminate calls always induce the same calculation and return the same results.
- lead implementation temporarily holds all the rows in the same partition in memory.

Any idea about above points and about other parts of the patch, please. 

I checked 'mvn clean install'


tajo-core/src/main/java/org/apache/tajo/engine/function/builtin/LeadDate.java
```diff
@@ -0,0 +1,49 @@
...
+import org.apache.tajo.catalog.CatalogUtil;
+import org.apache.tajo.catalog.Column;
+import org.apache.tajo.common.TajoDataTypes;
+import org.apache.tajo.common.TajoDataTypes.Type;
+import org.apache.tajo.engine.function.annotation.Description;
+import org.apache.tajo.engine.function.annotation.ParamTypes;
+
+@Description(
+    functionName = "lead",
+    description = "the nth previous row value of current row",
```
hyunsik: It should be 'nth next row'  instead of 'nth previous ...'.

tajo-core/src/main/java/org/apache/tajo/engine/function/builtin/LeadDouble.java
```diff
@@ -0,0 +1,49 @@
...
+import org.apache.tajo.catalog.CatalogUtil;
+import org.apache.tajo.catalog.Column;
+import org.apache.tajo.common.TajoDataTypes;
+import org.apache.tajo.common.TajoDataTypes.Type;
+import org.apache.tajo.engine.function.annotation.Description;
+import org.apache.tajo.engine.function.annotation.ParamTypes;
+
+@Description(
+    functionName = "lead",
+    description = "the nth previous row value of current row",
```
hyunsik: It should be 'nth next row' instead of 'nth previous ...'.

tajo-core/src/main/java/org/apache/tajo/engine/function/builtin/LeadFloat.java
```diff
@@ -0,0 +1,49 @@
...
+import org.apache.tajo.catalog.CatalogUtil;
+import org.apache.tajo.catalog.Column;
+import org.apache.tajo.common.TajoDataTypes;
+import org.apache.tajo.common.TajoDataTypes.Type;
+import org.apache.tajo.engine.function.annotation.Description;
+import org.apache.tajo.engine.function.annotation.ParamTypes;
+
+@Description(
+    functionName = "lead",
+    description = "the nth previous row value of current row",
```
hyunsik: It should be 'nth next row' instead of 'nth previous ...'.

tajo-core/src/main/java/org/apache/tajo/engine/function/builtin/LeadInt.java
```diff
@@ -0,0 +1,49 @@
...
+import org.apache.tajo.catalog.CatalogUtil;
+import org.apache.tajo.catalog.Column;
+import org.apache.tajo.common.TajoDataTypes;
+import org.apache.tajo.common.TajoDataTypes.Type;
+import org.apache.tajo.engine.function.annotation.Description;
+import org.apache.tajo.engine.function.annotation.ParamTypes;
+
+@Description(
+    functionName = "lead",
+    description = "the nth previous row value of current row",
```
hyunsik: It should be 'nth next row' instead of 'nth previous ...'.

tajo-core/src/main/java/org/apache/tajo/engine/function/builtin/LeadTime.java
```diff
@@ -0,0 +1,49 @@
...
+import org.apache.tajo.catalog.Column;
+import org.apache.tajo.common.TajoDataTypes;
+import org.apache.tajo.common.TajoDataTypes.Type;
+import org.apache.tajo.engine.function.annotation.Description;
+import org.apache.tajo.engine.function.annotation.ParamTypes;
+
+@Description(
+    functionName = "lead",
+    description = "the nth previous row value of current row",
+    example = "> SELECT lag(column, n) OVER ();",
```
hyunsik: The example explains lag, but this function is lead. Others also have same some copy-and-paste errors. It may be trivial and common mistakes :) 

Could you check them all?

hyunsik:
I'll review it soon. This feature is very desired in many applications. It would be great if we add this feature to 0.10.

hyunsik:
According to https://wiki.postgresql.org/wiki/SQL2008_windowing_queries, the offset is a positive integer (offset > 0). So, your assumption would be right. If offset or default value is omitted, default offset will be 1 and default value will be null.

hyunsik:
> lead implementation temporarily holds all the rows in the same partition in memory.

Do you mean a LinkedList in LeadContext?

sirpkt:
Thank you for the review, @hyunsik.
I'll fixed the wrong description. 

Yes, I mean that LinkedList.
It does not keep the copy of all rows but it does keep the reference of all rows and it prevents from freeing those resources during lead operation I think (is it right?).

sirpkt:
I just fixed the wrong description of lead functions.

hyunsik:
Thank you @sirpkt for your description and your fixes. I'm also sorry for late response. I took flight, and I just get back into harness.

I just wondered that you mentioned 'all rows' because in my view LinkedList<Datum> seems to all  column values.

sirpkt:
Sorry for making you confused.
I mean all the values of the given column.

hyunsik:
Thank you for correcting that. I'll finish the review as soon as possible.

hyunsik:
+1 ship it. The patch looks good to me.

hyunsik:
@sirpkt Could you add this patch to both master and 0.10.0 branch?

sirpkt:
Thank you for the review, @hyunsik 
I just committed the patch to both master and 0.10.0 branch.

dongjoon-hyun:
Hi, @hyunsik, @sirpkt. 
This issue was closed in GitHub, but not in https://issues.apache.org/jira/browse/TAJO-919. 
Could you close that too, please?

PR #353, opened 2015-01-16
TAJO-1305: With metadata storage of MySQL, columns with the same character but difference case are not allowed

ykrips:
It has been tested on MySQL 5.1.73 and MariaDB 5.5.41.


sirpkt:
+1
'mvn clean install' passed and I tested in hadoop-2.6.0 cluster with 10.0.15-MariaDB.

ykrips:
Thank you for reviewing this patch. It will be merged into master branch.

PR #355, opened 2015-01-21
TAJO-1309: Add missing break point in physical operator.

jinossy:


blrunner:
+1

It looks good overall and all unit test cases finished successfully.
Push it. :)

PR #356, opened 2015-01-22
TAJO-1289: History reader fails to get the query information after a successful query execution

jinossy:
Main problem is  some corrupted history files in hdfs. TajoClient are to get a query status in finished list when a query is finished. but the finished list is a stored files
So I’ve add LRU map for a finished query caching, and fix hidden race condition bug.

### DIFF
--- a/tajo-core/src/main/java/org/apache/tajo/master/QueryManager.java
+++ b/tajo-core/src/main/java/org/apache/tajo/master/QueryManager.java
@@ -20,6 +20,7 @@ package org.apache.tajo.master;
 
 import com.google.common.collect.Lists;
 import com.google.common.collect.Maps;
+import com.google.common.collect.Sets;
 import org.apache.commons.collections.map.LRUMap;
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
@@ -42,6 +43,7 @@ import org.apache.tajo.util.history.HistoryReader;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.Map;
+import java.util.Set;
 import java.util.concurrent.atomic.AtomicLong;
 
 /**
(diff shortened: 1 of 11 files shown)


blrunner:
+1

It looks good to me and I found that history server ran as expected on my testing server. :)

tajo-core/src/main/java/org/apache/tajo/master/QueryManager.java
```diff
@@ -121,19 +126,27 @@ public EventHandler getEventHandler() {
 
   public synchronized Collection<QueryInfo> getFinishedQueries() {
```
hyunsik: In some cases, it may result in lots of histories. In that case, It will cause long latency. Later, we need to improve the paging feature in new jira.

tajo-core/src/main/java/org/apache/tajo/util/history/HistoryWriter.java
```diff
@@ -49,8 +56,10 @@
   public static final String QUERY_LIST = "query-list";
   public static final String QUERY_DETAIL = "query-detail";
   public static final String HISTORY_FILE_POSTFIX = ".hist";
+  private static short REPLICATION = 1;
```
hyunsik: Since history may be not critical data in many cases, one replication would be enough. But, in some cases, it can be very important. How about improve it to be some configurable?

hyunsik:
The patch looks nice to me. I leave some trivial comments.

tajo-core/src/main/java/org/apache/tajo/master/QueryManager.java
jinossy: I agree with you, we should make to pagination

tajo-core/src/main/java/org/apache/tajo/util/history/HistoryWriter.java
jinossy: Sure!

jinossy:
Thanks guys for the review. I’ll commit it shortly.
If you have more review, please let me know

jinossy:
I've add two configuration following:
```
tajo.history.query.replication
tajo.history.task.replication
```
Please review again
Thanks

PR #357, opened 2015-01-22
TAJO-1311: Enable Scattered Hash Shuffle for CTAS statement.

blrunner:
Currently, Tajo just use hash shuffle for CTAS statement. If you use large partitioned table for target table, you will see poor performance of wrong task number. Thus, we need to calculate the number for the partitioned table with scattered hash shuffle.

For reference, CTAS includes two bugs as follows:
- Tajo doesn't verify target column number for CTAS statement
- If target table and source table are column partitioned table, Tajo doesn't support CTAS statement.

I'll create another jira issues at once.


tajo-core/src/main/java/org/apache/tajo/engine/planner/PhysicalPlannerImpl.java
```diff
@@ -854,6 +855,14 @@ private PhysicalExec createSortBasedColumnPartitionStorePlan(TaskAttemptContext
           sortSpecs[i++] = new SortSpec(insertNode.getProjectedSchema().getColumn(id), true, false);
         }
       }
+    } else if (storeTableNode.getType() == NodeType.CREATE_TABLE) {
+      int i = 0;
+      for (int j = 0; j < partitionKeyColumns.length; j++) {
+        int id = storeTableNode.getOutSchema().getColumns().size() + j;
```
jinossy: In my opinion, we should consider the qualified name for CTAS with defined column. 
@blrunner @hyunsik , How do you think?

tajo-core/src/main/java/org/apache/tajo/engine/planner/global/GlobalPlanner.java
```diff
@@ -1130,6 +1130,18 @@ private void setShuffleKeysFromPartitionedTableStore(StoreTableNode node, DataCh
       }
       channel.setShuffleKeys(shuffleKeys);
       channel.setShuffleType(SCATTERED_HASH_SHUFFLE);
+    } else if (node.getType() == NodeType.CREATE_TABLE) {
```
hyunsik: The newly added code is the mostly same as those of INSERT.  It would be nice if some code to get ProjectSchema and InSchema are distinguished.

tajo-core/src/test/java/org/apache/tajo/engine/query/TestTablePartitions.java
```diff
@@ -45,54 +45,72 @@
...
+    if (nodeType == NodeType.INSERT) {
+      res = executeString(
         "create table " + tableName + " (col1 int4, col2 int4) partition by column(key float8) ");
-    res.close();
+      res.close();
 
-    assertTrue(catalog.existsTable(DEFAULT_DATABASE_NAME, tableName));
-    assertEquals(2, catalog.getTableDesc(DEFAULT_DATABASE_NAME, tableName).getSchema().size());
-    assertEquals(3, catalog.getTableDesc(DEFAULT_DATABASE_NAME, tableName).getLogicalSchema().size());
+      assertTrue(catalog.existsTable(DEFAULT_DATABASE_NAME, tableName));
```
hyunsik: The assert statements for CREATE or INSERT are duplicated. This kind of assert statements in other unit test methods are also duplicated. You need to keep only one.

tajo-core/src/main/java/org/apache/tajo/engine/planner/PhysicalPlannerImpl.java
hyunsik: In 'CREATE TABLE', it is to hard to use qualified names of partition keys.

CreateTableNode's in/out schema only contains physically stored columns and does not contain logical columns (i.e., partition key columns). The way described in the patch will work correctly. 

But, since this way is not intuitive, we need to refactor CreateTableNode and InsertTableNode to provide a better way to access physical columns corresponding to partition keys.

hyunsik:
Any progress?

blrunner:
Sorry guys, I have been very busy.
I'll restart it shortly. :)

blrunner:
@hyunsik 

I'm sorry to give the feedback lately.
I cleaned up duplicated codes.
Could you check the patch again?

blrunner:
It seems as if there are some problems on commit logs.
I'll create new pull request soon.

PR #360, opened 2015-01-27
TAJO-1318: Unit test failure after miniDFS cluster restart

jinossy:


blrunner:
+1

It looks good to me and all unit test cases finished successfully.
Push it~ :)

PR #361, opened 2015-01-27
TAJO-1319: Tajo can't find HBase configuration file.

blrunner:
Currently, Tajo can't find HBase configuration file as follows:

default> CREATE EXTERNAL TABLE hbase_blog (rowkey text, author text, register_date text, title text)
 USING hbase WITH (
   'table'='blog'
   , 'columns'=':key,info:author,info:date,content:title');

ERROR: HBase mapped table is required a 'hbase.zookeeper.quorum' attribute.

We need to check tajo shell script file.


jinossy:
+1 LGTM
Ship it!

PR #363, opened 2015-01-28
TAJO-1320: HBaseStorageManager need to support Zookeeper Client Port.

blrunner:
Currently, HBaseStorageManager just use Zookeeper Quorum host names. But I think that users may want to change their Zookeeper client port. Thus, we need to support that users can set the port. For reference, HBase defines the port to ZOOKEEPER_CLIENT_PORT in HBaseConstants.

### DIFF
--- a/tajo-storage/tajo-storage-hbase/src/main/java/org/apache/tajo/storage/hbase/HBaseStorageConstants.java
+++ b/tajo-storage/tajo-storage-hbase/src/main/java/org/apache/tajo/storage/hbase/HBaseStorageConstants.java
@@ -27,6 +27,7 @@ public interface HBaseStorageConstants {
   public static final String META_SPLIT_ROW_KEYS_KEY = "hbase.split.rowkeys";
   public static final String META_SPLIT_ROW_KEYS_FILE_KEY = "hbase.split.rowkeys.file";
   public static final String META_ZK_QUORUM_KEY = "hbase.zookeeper.quorum";
+  public static final String META_ZK_CLIENT_PORT = "hbase.zookeeper.property.clientPort";
   public static final String META_ROWKEY_DELIMITER = "hbase.rowkey.delimiter";
 
   public static final String INSERT_PUT_MODE = "tajo.hbase.insert.put.mode";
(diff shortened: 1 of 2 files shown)


tajo-storage/tajo-storage-hbase/src/main/java/org/apache/tajo/storage/hbase/HBaseStorageConstants.java
```diff
@@ -27,6 +27,7 @@
   public static final String META_SPLIT_ROW_KEYS_KEY = "hbase.split.rowkeys";
   public static final String META_SPLIT_ROW_KEYS_FILE_KEY = "hbase.split.rowkeys.file";
   public static final String META_ZK_QUORUM_KEY = "hbase.zookeeper.quorum";
+  public static final String META_ZK_CLIENT_PORT = "hbase.zookeeper.property.clientPort";
```
jihoonson: We traditionally avoid the Camel style for configuration key names.
And, we need to make the naming style consistent.
So, how about ```hbase.zookeeper.client.port```?

tajo-storage/tajo-storage-hbase/src/main/java/org/apache/tajo/storage/hbase/HBaseStorageManager.java
```diff
@@ -312,6 +312,17 @@ public static Configuration getHBaseConfiguration(Configuration conf, TableMeta
           HBaseStorageConstants.META_ZK_QUORUM_KEY + "' attribute.");
     }
 
+    String zkPort = hbaseConf.get(HConstants.ZOOKEEPER_CLIENT_PORT);
```
jihoonson: I wonder why we don't use the default value.
Should users always specify the zookeeper client port?

jihoonson:
Hi @blrunner, your patch looks good.
I left some minor comments.
Please consider them.

tajo-storage/tajo-storage-hbase/src/main/java/org/apache/tajo/storage/hbase/HBaseStorageConstants.java
blrunner: Thank you for your review.
The name borrowed with HBase Constant class. I think that HBase users may be familiar with it compare than another name.

tajo-storage/tajo-storage-hbase/src/main/java/org/apache/tajo/storage/hbase/HBaseStorageManager.java
blrunner: We already use the default value.
If users doesn't set zookeeper hosts or client port, Tajo uses properties of HBase configuration file. 
The above code just be used to check empty value.

jihoonson:
Thanks for your comment.
Those changes look good to me.
+1!

PR #364, opened 2015-01-28
TAJO-1283: ORDER BY with the first descending order causes wrong results

sirpkt:
As BaseTupleComparator already handles both ascending and descending sort keys, remove unnecessary DescendingTupleRangeComparator class.    
- We do not need to consider whether the first sort key is ascending or descending in scheduleRangeShuffledFetches() so related codes are removed.
- Add TEST_MIN_TASK_NUM support in getNonLeafTaskNum() of Stage.java for Test purpose.

'mvn clean install' passed.


tajo-core/src/main/java/org/apache/tajo/querymaster/Repartitioner.java
```diff
@@ -725,13 +725,8 @@ public static void scheduleRangeShuffledFetches(TaskSchedulerContext schedulerCo
       }
     }
 
-    boolean ascendingFirstKey = sortSpecs[0].isAscending();
     SortedMap<TupleRange, Collection<FetchImpl>> map;
-    if (ascendingFirstKey) {
-      map = new TreeMap<TupleRange, Collection<FetchImpl>>();
-    } else {
-      map = new TreeMap<TupleRange, Collection<FetchImpl>>(new TupleRange.DescendingTupleRangeComparator());
```
jinossy: This patch looks good to me.  
@hyunsik and @jihoonson Do we have any reason  to keep the descending ranges ?

tajo-core/src/main/java/org/apache/tajo/querymaster/Repartitioner.java
```diff
@@ -740,7 +735,7 @@ public static void scheduleRangeShuffledFetches(TaskSchedulerContext schedulerCo
         fetchSet = new HashSet<FetchImpl>();
         for (FetchImpl fetch: fetches) {
           String rangeParam =
-              TupleUtil.rangeToQuery(ranges[i], ascendingFirstKey ? i == (ranges.length - 1) : i == 0, encoder);
```
jinossy: @sirpkt 
I have heard from hyunsik that descending order consider the sampling and limit clause.
In my opinion, main bug fixing  is good. 
The range array is ascending. so {{ascendingFirstKey}} checking do not need
```
TupleUtil.rangeToQuery(ranges[i], i == (ranges.length - 1), encoder);
```

tajo-core/src/main/java/org/apache/tajo/querymaster/Repartitioner.java
hyunsik: I'm sorry for late response. Descending order is for the future feature. So, it is not necessary right now. Later, it can be added again when necessary.

tajo-core/src/main/java/org/apache/tajo/querymaster/Repartitioner.java
hyunsik: Please refer to the above comment I leaved.

hyunsik:
Hi @sirpkt, 
Could you trigger this unit test?

hyunsik:
+1 ship it!

The patch includes the correct bug fix and enough unit tests.

hyunsik:
@sirpkt Could you please commit this patch to both branch-0.10.0 and master branch?

sirpkt:
Thank you for the comments, @hyunsik, @jinossy 
I just committed to master and branch-0.10.0.

PR #366, opened 2015-01-29
TAJO-1135: Implement queryable virtual table for cluster information

ykrips:
This virtual table will be shown like following.
```
information_schema> select * from workers;
host,  port,  type,  status,  total_cpu,  used_mem,  total_mem,  free_heap,  max_heap,  used_diskslots,  total_diskslots,  running_tasks,  heartbeat
hostnameA,  28093,  QueryMaster,  RUNNING,  ,  ,  ,  130004408,  932118528,  ,  ,  0,  2015-01-29 17:09:18.859999
hostnameA,  28091,  Worker,  RUNNING,  1,  0,  1073741824,  130004408,  932118528,  0.0,  1.0,  0,  2015-01-29 17:09:18.859999
(2 rows, 0.025 sec, 0 B selected)
```


jihoonson:
Hi @ykrips, thanks for your work.
In overall, the patch looks good to me.
Only one thing I want to suggest is the table name.
Since the newly added table contains the information of query masters as well as workers, I think that ```cluster``` is the more proper name than ```workers```.
What do you think about it?

ykrips:
Good idea. ```cluster``` would be better. fix up and post the commit soon.

hyunsik:
I'm sorry for the late review. I'll review it soon.

jihoonson:
Sorry for late review.
The latest patch looks good to me.
Here is my +1, but it will be good to wait for Hyunsik's review.

ykrips:
No problems. I'll commit it after hyunsik finishes the review.

hyunsik:
I'll finish the review by tomorrow.

hyunsik:
I'm sorry for late review. The patch looks good to me. Even though the patch is conflict with the latest revision, the conflict is trivial. So, I tested it easily. 

One suggestion is one column name 'heartbeat. In my view, the 'heartbeat' is slightly ambiguous. I think that it would be better if we use 'last_heartbeat' or 'last_heartbeat_ts'.

ykrips:
That would be good. This column name will be changed in a next commit.

PR #367, opened 2015-01-30
TAJO-1322: Invalid stored caching on StorageManager.

jinossy:


blrunner:
+1

It looks good to me. 
Ship it. :)

PR #368, opened 2015-01-30
TAJO-1323: Cleanup the unstable test case.

jinossy:


blrunner:
+1

You resolved our endemic test problems and all unit test cases finished successfully.

Push it. :)

PR #371, opened 2015-02-03
TAJO-1325: Invalid history cleaner timeout

jinossy:


tajo-core/src/main/java/org/apache/tajo/querymaster/QueryMaster.java
```diff
@@ -525,7 +524,13 @@ private void cleanExpiredFinishedQueryMasterTask(long expireTime) {
       synchronized(finishedQueryMasterTasks) {
         List<QueryId> expiredQueryIds = new ArrayList<QueryId>();
         for(Map.Entry<QueryId, QueryMasterTask> entry: finishedQueryMasterTasks.entrySet()) {
-          if(entry.getValue().getStartTime() < expireTime) {
+          long finishedTime = entry.getValue().getStartTime();
```
jihoonson: Would you leave some comments?

tajo-core/src/main/java/org/apache/tajo/worker/TaskRunnerManager.java
```diff
@@ -240,7 +240,9 @@ private void cleanExpiredFinishedQueryMasterTask(long expireTime) {
       synchronized(taskRunnerHistoryMap) {
         List<String> expiredIds = new ArrayList<String>();
         for(Map.Entry<String, TaskRunnerHistory> entry: taskRunnerHistoryMap.entrySet()) {
-          if(entry.getValue().getStartTime() > expireTime) {
+
+          long finishedTime = Math.max(entry.getValue().getStartTime(), entry.getValue().getFinishTime());
```
jihoonson: Would you leave some comments?

jihoonson:
+1 LGTM.
Please add some explanations before you commit.

tajo-core/src/main/java/org/apache/tajo/worker/TaskRunnerManager.java
jinossy: Sure!

PR #374, opened 2015-02-04
TAJO-1269: Separate cli from tajo-client.

hyunsik:
This work separates tajo-cli from tajo-client. As a result, tajo-cli includes only cli tools and tsql. tajo-client contains only client API itself. In addition, TajoHAClientUtil checks the connectivity between client and master. This patch move the connectivity check to SessionConnection. So, I removed TajoHAClientUtil.


jinossy:
You should add tajo-cli-module in tajo-dist maven plugin

hyunsik:
@jinossy Thanks! I missed it. I'll add it.

hyunsik:
@jinossy I just updated the tajo-dist/pom.xml.

jinossy:
+1 LGTM
I'll fix the Unit test failure

PR #377, opened 2015-02-06
TAJO-1339: Incorrect handling of tables with custom delimiter when their data contain '|'

sirpkt:
meta options of scanned tables are passed between Stages through DataChannel
    
- DataChannel contains meta options of source Execution block, collected by visiting all the ScanNodes in the block. 
  - When multiple ScanNodes have different options about the same key, options of last visited ScanNode overwrite previous ones.
- Each Execution block reflects incoming DataChannel in buildInputExecutor.
- Meta options of the Stage are recalculated before materializing the result, which reflect all the options from DataChannels and also new ScanNode in the given Stage.
    
Changes:
- DataChannel contains all the meta options from the source Execution block, which means protocol buffer definition is also modified.
- PlannerUtil provides new static method getScanOptions(), which visit all the ScanNode in the Execution block and get options.
- buildInputExecutor() is modified to reflect meta options of incoming DataChannel of the given Execution block.
- 'csvfile.delimiter' option sets CSVFILE_DELIMITER (not TEXT_DELIMITER) because TEXT_DELIMITER is set as '|' even when there is no explicit 'csvfile.delimiter' so that explicit 'csvfile.delimiter' should be prior to default delimiter. 
  - Without this, when two tables, one is 'csvfile.delimiter'=';' and the other has no option, are used in the same query, one is set as TEXT_DELIMITER = ';' and the other is set as TEXT_DELIMITER = '|' and compete for the meta option of the Execution block. 
  - If we give high prioritiy to the explicit meta setting, it always wins over the default value and I think it is desirable.
    
I think some changes may be controversial.


hyunsik:
I'm sorry for late review. I'll review it soon.

jihoonson:
Sorry for late review. I'll review today.

jihoonson:
@sirpkt thanks for your work.
I have a couple of comments.
First, ```CSVFILE_DELIMITER``` is deprecated. If there are any bugs related to ```TEXT_DELIMITER```, we must fix it instead of using ```CSVFILE_DELIMITER```.
In addition, I have a question about the following your description.
```
Without this, when two tables, one is 'csvfile.delimiter'=';' and the other has no option, are used in the same query, one is set as TEXT_DELIMITER = ';' and the other is set as TEXT_DELIMITER = '|' and compete for the meta option of the Execution block.
```
If only one table is configured with an explicitly different delimiter, two tables must be parsed with different delimiters. If I misunderstand anything, please let me know.

Second, it is hard to unserstand for me why ```CSVFILE_DELIMITER``` or ```TEXT_DELIMITER``` is required to store intermediate results. As you know, the default file type of intermediate results is ```RAW```, not ```CSV``` or ```TEXT```.
Would you mind explaining more detailed backgrounds?

hyunsik:
Hi @sirpkt,

I think that this patch can be even more simplified. It may be enough to change only few points. If you are ok, I'll do this work.

sirpkt:
I feel very sorry for late response, @jihoonson , @hyunsik 
I completely forgot this issue due to my recent trip to Europe.

Previously, I thought this issue could be resolved with TAJO-1340.
However, if there is simpler way, please feel free to take the issue, @hyunsik

hyunsik:
Thank you for quick response.

Actually, the real problem is that TEXT is a default output file format, which is prone to cause delimiter conflicts. As you mentioned, TAJO-1340 would be an essential solution. Nevertheless, your issue would be a nice temporary workaround now. I think that it can be included in 0.10.1 bug fix release. I'll take it.

PR #378, opened 2015-02-09
TAJO-1298: Improve the printing message after an execution of an index creation statement

jihoonson:
In this issue, I improved the printing message after an index creation.
In addition, I changed the writing process of ```create index``` query to exploit the result staging feature. Instead of directly writing the results in the output directory, Tajo writes them in the staging directory and moves them to the final output directory on the successful query execution.

The following is an example of the improved message.

```
default> create index l_orderkey_idx on lineitem (l_orderkey);
Progress: 0%, response time: 1.169 sec
Progress: 0%, response time: 1.17 sec
...
Progress: 83%, response time: 17.446 sec
Progress: 100%, response time: 17.758 sec
(0 rows, 17.758 sec, 70.4 MiB inserted)
default>
```

As shown in this example, the number of inserted bytes represents the total index size.
However, the number of rows is 0. I think this is right because index files contain the binary information about the index structure, so the number of inserted rows cannot be computed.

Alternatively, we can show the number of rows of source table instead of the inserted number of rows.
In this case, I'm concerning the inconsistency of the printing messages.

What do you think?
Or, any other good idea?

[Review on Reviewable](https://reviewable.io/reviews/apache/tajo/378)


jihoonson:
I fixed the test failure.

hyunsik:
```
Progress: 100%, response time: 17.758 sec
(0 rows, 17.758 sec, 70.4 MiB inserted)
default>
```

It does look the result about index creation because the message is the same to general CTASs. Especially, it would be better if the word 'inserted' is changed.

jihoonson:
Thanks for your comment.
I'll change 'inserted' to 'written'.

jihoonson:
I changed the message according to @hyunsik's comment.

hyunsik:
How about 'Index written'? I think that just 'written' may look 'table written' or 'data written'.

jihoonson:
Actually, I used 'written' as the meaning of 'data written'. 
But, as you said, it would be better if users can see what type of query is finished.
So, I'd like to suggest adding the finished query type as follows.
```
Progress: 100%, response time: 17.758 sec
(0 rows, 17.758 sec, 70.4 MiB written)
CREATE INDEX OK
default>
```
We also can show this kind of message after ```CREATE TABLE``` or ```INSERT``` statements.
What do you think?

hyunsik:
+1
The change looks good to me. This patch seems to be ready to be committed.

jihoonson:
@hyunsik, thanks for your comment.
Actually, I have just updated my patch according to my suggestion.
Please review again.

hyunsik:
Could you trigger the unit test?

jihoonson:
@hyunsik, I made some changes in GlobalEngine, especially DDL execution part. They cause some problems with HBaseStorqageManager. I already created an issue (https://issues.apache.org/jira/browse/TAJO-1373). After resolving that issue, I'll trigger again.

jihoonson:
@hyunsik, this patch is finally ready for review. Please review when you have sufficient free slots.

PR #379, opened 2015-02-10
TAJO-1277: GreedyHeuristicJoinOrderAlgorithm sometimes wrongly assumes associativity of joins

sirpkt:
Basically, it limits the range of join ordering until it meets outer join operations.
For example, in the case of (((((a inner join b) inner join c) outer join d) inner join e) inner join f),
join ordering is partitioned as three parts as
1) (a inner join b) inner join c
2) (result of 1) outer join d
3) (((result of 2) inner join e) inner join f) 

Following modifications are included:
- findBestOrder() is changed to partition join ordering
- getBestPair() and findJoin() are changed to return the corresponding JoinEdges of the selected join because those JoinEdges should be removed before next join ordering

It passes 'mvn clean install' and several join cases I tested,
however, I'm not sure this is good approach.
Please, leave me comments about the patch.

### DIFF
--- a/tajo-plan/src/main/java/org/apache/tajo/plan/joinorder/GreedyHeuristicJoinOrderAlgorithm.java
+++ b/tajo-plan/src/main/java/org/apache/tajo/plan/joinorder/GreedyHeuristicJoinOrderAlgorithm.java
@@ -28,10 +28,7 @@ import org.apache.tajo.plan.expr.AlgebraicUtil;
 import org.apache.tajo.plan.logical.*;
 import org.apache.tajo.util.TUtil;
 
-import java.util.LinkedHashSet;
-import java.util.Set;
-import java.util.SortedSet;
-import java.util.TreeSet;
+import java.util.*;
 
 /**
  * This is a greedy heuristic algorithm to find a bushy join tree. This algorithm finds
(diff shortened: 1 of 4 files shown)


tajo-plan/src/main/java/org/apache/tajo/plan/joinorder/GreedyHeuristicJoinOrderAlgorithm.java
```diff
@@ -57,17 +54,69 @@ public FoundJoinOrder findBestOrder(LogicalPlan plan, LogicalPlan.QueryBlock blo
...
+          }
+        }
+
+        // check if the relation is the last piece of outer join
+        boolean endInnerRelation = false;
+        for (JoinEdge joinEdge: joinEdges) {
+          switch(joinEdge.getJoinType()) {
+            case LEFT_OUTER:
+            case RIGHT_OUTER:
+            case FULL_OUTER:
```
jihoonson: We should consider other join types such as ```anti join``` or ```semi join``` as well as ```inner join``` and ```outer join```.
So, I think that the iteration must be stopped when finding ```anti join``` or ```semi join```.

tajo-plan/src/main/java/org/apache/tajo/plan/joinorder/GreedyHeuristicJoinOrderAlgorithm.java
```diff
@@ -57,17 +54,69 @@ public FoundJoinOrder findBestOrder(LogicalPlan plan, LogicalPlan.QueryBlock blo
...
+        for (String relationString: relationStrings) {
+          joinEdgesForGiven = joinGraph.getIncomingEdges(relationString);
+          if (joinEdgesForGiven != null) {
+            joinEdges.addAll(joinEdgesForGiven);
+          }
+        }
+
+        // check if the relation is the last piece of outer join
+        boolean endInnerRelation = false;
+        for (JoinEdge joinEdge: joinEdges) {
```
jihoonson: Here, joinEdges seem to be sorted in the order of their occurrence in the user query.
So, I wonder how we can guarantee that joinEdges are always sorted in that order.

jihoonson:
@sirpkt, thanks for your nice work.
The basic approach looks good to me.
I left some comments.

Sometimes, outer joins can be associative under some conditions. (Please refer to http://stackoverflow.com/questions/20022196/are-left-outer-joins-associative, https://cwiki.apache.org/confluence/display/Hive/LanguageManual+Joins)
So, we should handle them, but I think that it should be done in another issue.

tajo-plan/src/main/java/org/apache/tajo/plan/joinorder/GreedyHeuristicJoinOrderAlgorithm.java
sirpkt: Thank you for the comment, @jihoonson .
I'll add other join types.

tajo-plan/src/main/java/org/apache/tajo/plan/joinorder/GreedyHeuristicJoinOrderAlgorithm.java
sirpkt: Actually, I thought that the order of JoinEdges in joinEdges list does not matter.
Since the given relation is the right relation for all the JoinEdges,
I just check the existence of JoinEdge that is outer join.
And, in my thought, the result does not vary regardless of the order of JoinEdge's occurrence.

jihoonson: You are right. It was my misunderstanding.

tajo-plan/src/main/java/org/apache/tajo/plan/joinorder/GreedyHeuristicJoinOrderAlgorithm.java
```diff
@@ -57,17 +54,69 @@ public FoundJoinOrder findBestOrder(LogicalPlan plan, LogicalPlan.QueryBlock blo
...
+
+      for (LogicalNode relation : remainRelations) {
+        Collection <String> relationStrings = PlannerUtil.getRelationLineageWithinQueryBlock(plan, relation);
+        List<JoinEdge> joinEdges = new ArrayList<JoinEdge>();
+        String relationCollection = TUtil.collectionToString(relationStrings, ",");
+        List<JoinEdge> joinEdgesForGiven = joinGraph.getIncomingEdges(relationCollection);
+        if (joinEdgesForGiven != null) {
+          joinEdges.addAll(joinEdgesForGiven);
+        }
+        for (String relationString: relationStrings) {
```
jihoonson: Would you mind explaning these lines?
I found that the same join edge is added twice to the ```joinEdgesForGiven``` list.

sirpkt: Oh, it's my mistake.
When relationStrings has only one entry, this code may adds that entry twice.

When a LogicalNode contains two relations, for example, A and B,
above code first finds joinEdges whose right relation is "A, B", which is obtained by TUtil.collectionToString().
Next, it finds joinEdges whose right relation is "A" or "B", which is obtained by 'for (String relationString: relationStrings)'.
So, if a LogicalNode contains just one relation, this code may adds that relation repeatedly.

Duplicated relation does not affect the result but I'll patch not to have duplicated relations.

jihoonson: Thanks for your explanation.

sirpkt:
I updated the patch to reflect the comments from @jihoonson .
- add semi and anti join cases
- avoid duplicated JoinEdges

jihoonson:
+1
The latest patch looks good to me.
Thanks for your work!

hyunsik:
+1
The patch contains a workaround code which can fix the bug right now.

I have some comments. Your workaround code seems to try to firstly find the order of joins whose join types is same. In my opinion, this approach may miss some of join order combinations. It may cause sub-optimal join orders in some cases. I think that the right way is to improve the findBestPair() to find only associative join orders. It would be great if we do it in another jira.

jihoonson:
@hyunsik, thanks for your review.
I created an issue to handle the potentially missed cases.
https://issues.apache.org/jira/browse/TAJO-1352

PR #381, opened 2015-02-12
TAJO-1310: Maintaining join filters in join operators

jihoonson:
Summary of changes
* Add ```AbstractJoinExec``` which is a super class of all physical join executor classes.
* Add the join filter to the logical ```JoinNode``` and ```AbstractJoinExec```.


tajo-core/src/main/java/org/apache/tajo/engine/planner/physical/HashLeftOuterJoinExec.java
```diff
@@ -211,12 +169,16 @@ protected void loadRightToHashTable() throws IOException {
...
       List<Tuple> newValue = tupleSlots.get(keyTuple);
-      if (newValue != null) {
-        newValue.add(tuple);
-      } else {
-        newValue = new ArrayList<Tuple>();
-        newValue.add(tuple);
-        tupleSlots.put(keyTuple, newValue);
+      try {
+        if (newValue != null) {
+          newValue.add(tuple.clone());
```
jihoonson: I found that this cloning increases the memory space usage significantly, which causes frequent GC.
Currently, there is a problem of that the reference of the ```tuple``` value is not changed.
That is, the underlying operator returns the ```tuple``` with the same reference but the different contents.
I'll fix it.

jihoonson: This problem is quite complicated. So, I create a new issue (https://issues.apache.org/jira/browse/TAJO-1343) to handle this problem.
Please review the patch.

tajo-core/src/main/java/org/apache/tajo/engine/planner/physical/AbstractJoinExec.java
```diff
@@ -0,0 +1,157 @@
...
+
+  public void setJoinFilter(EvalNode joinFilter) {
+    this.joinFilter = joinFilter;
+  }
+
+  public EvalNode getJoinFilter() {
+    return joinFilter;
+  }
+
+  public boolean evalFilter() {
```
hyunsik: The method is different to our naming convention. It should be hasEvalFilter().

tajo-core/src/main/java/org/apache/tajo/engine/planner/physical/AbstractJoinExec.java
```diff
@@ -0,0 +1,157 @@
...
+    filterSatisfied = hasJoinFilter() ?
+        joinFilter.eval(inSchema, frameTuple).isTrue() : true;
+    return filterSatisfied;
+  }
+
+  public boolean evalQual() {
```
hyunsik: It also the same issue to above comment.

tajo-core/src/main/java/org/apache/tajo/engine/planner/physical/BNLJoinExec.java
```diff
@@ -190,15 +152,12 @@ public Tuple next() throws IOException {
...
       }
 
-      frameTuple.set(leftTuple, rightIterator.next());
-      if (hasJoinQual) {
-        if (joinQual.eval(inSchema, frameTuple).isTrue()) {
-          projector.eval(frameTuple, outputTuple);
-          return outputTuple;
+      updateFrameTuple(leftTuple, rightIterator.next());
+
+      if (evalQual()) {
```
hyunsik: Except for BNLJoinExec, all join operators (hash, and merge joins) must have join conditions.

BNLJoinExec can be used for cross join, meaning that there is no join condition. Nevertheless, it can have join filter as follows:

```
SELECT x, y, z FROM A, B WHERE concat (x,y) = 'keyword'
```

```concat (x,y) = 'keyword'``` will be play a join filter in BNL join operator. So, we need to consider join condition and filter separately in BNLJoinExec.

File: tajo-plan/src/main/java/org/apache/tajo/plan/rewrite/rules/FilterPushDownRule.java [hunk 5]
```diff
@@ -358,7 +353,21 @@ public LogicalNode visitJoin(FilterPushDownContext context, LogicalPlan plan, Lo
       context.pushingDownFilters.removeAll(matched);
     }
 
-    context.pushingDownFilters.addAll(outerJoinFilterEvalsExcludePredication);
+    matched.clear();
+    matched.addAll(outerJoinFilterEvalsExcludePredication);
```
hyunsik: I know that you didn't name the variables ```outerJoinPredicationEvals``` and ```outerJoinFilterEvalsExcludePredication```.

Their name is not intuitive. So, it would be great if we rename them to more better ones.

tajo-plan/src/main/java/org/apache/tajo/plan/rewrite/rules/FilterPushDownRule.java
```diff
@@ -358,7 +353,21 @@ public LogicalNode visitJoin(FilterPushDownContext context, LogicalPlan plan, Lo
       context.pushingDownFilters.removeAll(matched);
     }
 
-    context.pushingDownFilters.addAll(outerJoinFilterEvalsExcludePredication);
+    matched.clear();
```
hyunsik: Could add some brief comment about the purpose of the following block? It would be helpful for other code readers.

hyunsik:
The patch looks good to me. The changes clean up join operators and and join plans. This separation between join filter and join condition would make more opportunities to clean up other join related codes later. I leave some comments. After your answer, I'll finish the review.

In addition, I think that FilterPushDownRule::visitJoin has a long single method. We need to clean up and refactor the method into several small well-defined methods. For it, we may need another jira.

hyunsik:
I created https://issues.apache.org/jira/browse/TAJO-1350 for visitJoin() refactoring issue.

tajo-core/src/main/java/org/apache/tajo/engine/planner/physical/BNLJoinExec.java
jihoonson: Thanks for your comment.
I missed it.

tajo-plan/src/main/java/org/apache/tajo/plan/rewrite/rules/FilterPushDownRule.java
jihoonson: I'll change those variable names in https://issues.apache.org/jira/browse/TAJO-1350.

tajo-plan/src/main/java/org/apache/tajo/plan/rewrite/rules/FilterPushDownRule.java
jihoonson: Ok. I'll add.

jihoonson:
Thanks for your review. I'll update the patch, soon.

hyunsik:
@jihoonson Could you trigger TravisCI?

jihoonson:
@hyunsik,
As I commented at https://issues.apache.org/jira/browse/TAJO-1310, it would be better to proceed after resolving other issues (https://issues.apache.org/jira/browse/TAJO-1350 and https://issues.apache.org/jira/browse/TAJO-1352.)
I think that many code lines must be changed after resolving them.
So, I'll close this PR, and make another one.

PR #382, opened 2015-02-12
TAJO-1284: Add alter partition method to CatalogStore.

blrunner:
This patch will implement ALTER TABLE ADD PARTITION method and ALTER TABLE DROP PARTITION method to CatalogStore.


blrunner:
I removed existing addPartition and addPartitions at CatalogStore. Because it seemed like an inconvenient methods to implement and maintain. So, I implemented it to alter table method. For reference, I implemented it to HCatalogStore, But I didn't implement getAllPartitions and get Partitions because of HiveMetastore performance.

blrunner:
I'll open new PR.

PR #383, opened 2015-02-16
TAJO-1351: Resolve findbug warnings on Tajo Common Module

ykrips:
Removed findbugs warnings except 1 item, unread field on org.apache.tajo.util.datetime.DateTimeFormat.java.


hyunsik:
+1
The patch looks good to me. Even though there remains one warning, findbug rules occasionally is too strict. We also need to exclude some of them. It would be great if we remove most of warnings firstly and we handle few cumbersome ones later.

PR #388, opened 2015-02-27
TAJO-1314: Documentation for the support of the swift

jihoonson:


hyunsik:
Could you add some example for CREATE TABLE? It may be similar to those of other storage types. Nevertheless, the example would be helpful to users.

jihoonson:
Sure. I'll add it.

jihoonson:
Added an example of the table creation.

hyunsik:
+1
The patch looks great to me. Ship it!

PR #394, opened 2015-03-02
TAJO-1367: Tajo Python Native Client

charsyam:
Implementation of Tajo Python Native Client(with Protocol Buffer)

currently it just supports executeQuery. But I think this is useful


hyunsik:
According to my test, it works well. Thank you for your contribution. This is the first trial to put python code to Tajo repository. I'm still considering wha the best source tree layout is. I'll review it soon.

PR #395, opened 2015-03-03
TAJO-1366: The timestamp type conversion occasionally leads to wrong results

jihoonson:
Let me suppose an example query as follows:
```
select 'Mon Nov 03 00:03:00 +0000 2014'::timestamp;
```
The problem is that 'Mon' is identified as the ```UNITS``` type, but its actual type is ```DOW``` in this query.
This wrong parsing result is led by the duplicated key when managing DateTimeConstants in a map.
So, I removed all items of ```deltatktbl```, which causes the duplication problem, from the constants map.
This table is used when parsing ```interval```, but we don't support it yet.


jinossy:
+1 LGTM!

jihoonson:
@jinossy thanks for the quick review!

PR #396, opened 2015-03-05
TAJO-1370 Some stack trace information is missed in error/fail logging

navis:
Trivial patch for getting stack trace from log. I doubt I'm doing this in right way.


hyunsik:
+1
The change looks straightforward.

hyunsik:
This patch is for TAJO-1369 rather than TAJO-1370. So, I'll change the commit log before committing it.

PR #397, opened 2015-03-05
TAJO-1370 TUtils.checkEquals() is not consistent with description in jav...

navis:
IntelliJ said it's strange. :)


hyunsik:
+1

The change looks straightforward. Thank you for nice finding.

PR #398, opened 2015-03-05
TAJO 1348 Parser error during create partitioned table

navis:
Appearance order can be changeable between using/partition-spec/table-param. There would be better way I guess.


blrunner:
I'll review tomorrow.

PR #400, opened 2015-03-06
TAJO-1374 Support multi-bytes delimiter for CSV file

navis:
Simple patch


hyunsik:
Hi @navis,
Thank you for your contribution. There may be mistake in the title. Could you rename the title? 

In these days, TravisCI has been too slow. There have been many pending pull requests for this reason. So, it would be better if you submit the patch to the jira and mark the jira as PATCH AVAILABLE. Please see this mailing thread too.

http://markmail.org/message/itfwyhcoppqmaouu

jinossy:
@navis Good to see you
The CSVFile will be deprecated. please refer to DelimitedTextFile

navis:
@jinossy If it will be not deprecated in near future, can we still get some usefulness in this? Fixed test fails, anyway.

hyunsik:
The patch looks nice to me. Of course, your work is useful. 

It would be great if this feature is applied to DelimitedTextFile too because DelimitedTextFile is new replacement to CSVFile. DelimitedTextFile's performance is really great. According to some benchmark result, it can parse more than 500MB CSV data sets per second. It also can boost up query response times in many cases, especially I/O intensive workloads.

I think that the work for DelimitedTextFile does not need to be done in this issue. We can do in another jira.

Anyway, could you fix some test failure? It still has one test failure.

tajo-storage/tajo-storage-common/src/main/java/org/apache/tajo/storage/LazyTuple.java
```diff
@@ -134,7 +134,7 @@ else if (textBytes.length <= fieldId) {
       }
       textBytes[fieldId] = null;
     } else {
-      //non-projection
+      values[fieldId] = NullDatum.get(); //non-projection
```
ykrips: This code fills the NullDatum to values array of LazyTuple when user does not insert any data on this index. However, VTuple class does not initialize the values array on itself.

jinossy:
@navis 
OK, I've create new jira issue TAJO-1381.
If you fix test failure, we can start review.

Thanks

tajo-storage/tajo-storage-common/src/main/java/org/apache/tajo/storage/LazyTuple.java
navis: reverted this part. I've hardly noticed tajo expects null for not-picked and byte[0] for not-existing picked.

navis:
Fixed test fails

jinossy:
+1 LGTM I'll commit it shortly.
Thank you for your contribution

PR #402, opened 2015-03-09
TAJO-1368: Exceptions during processing nested union queries

ykrips:
Please feel free to leave any comments on this patch.


jihoonson:
I'll review tomorrow.

tajo-core/src/main/java/org/apache/tajo/engine/planner/global/GlobalPlanner.java
```diff
@@ -1371,32 +1373,57 @@ public LogicalNode visitUnion(GlobalPlanContext context, LogicalPlan plan, Logic
...
+          (((TableSubQueryNode)leftChild).getSubQuery().getType() == NodeType.UNION);
+      boolean rightUnion = (rightChild.getType() == NodeType.TABLE_SUBQUERY) &&
+          (((TableSubQueryNode)rightChild).getSubQuery().getType() == NodeType.UNION);
       if (leftChild.getType() == NodeType.UNION) {
         unionBlocks.add(leftBlock);
       } else {
-        queryBlockBlocks.add(leftBlock);
+        if (leftUnion) {
+          node.setLeftChild(
+              createScanNodeWithTableSubQuery(masterPlan,
```
jihoonson: Please check whether I understand your patch well.
When an union has a child of subquery that has another union child, the subquery and its child union should be removed. Let me suppose an example union ```u1```. It has a child subquery ```s1``` which has another union child, ```u2```. Finally, ```u2``` has a scan child, ```scan1```. In this case, ```u1``` should be directly connected with ```scan```.

However, according to your patch, the example query forms (```scan1``` - ```u2``` - ```s1```) - (```scan2``` - ```u1```). Parentheses mean query blocks, and ```scan2``` is created at this line. In this case, I think that it is difficult to avoid unnecessary materialization of intermediate data between query blocks.

If I misunderstand something, please tell me.

jihoonson:
@ykrips thanks for your patch. I left a comment. 
In addition, it is difficult to understand details of some changes. 
Would you mind telling me a brief summarization of the changes?
Also, it would be great if you add some comments.

tajo-core/src/main/java/org/apache/tajo/engine/planner/global/GlobalPlanner.java
ykrips: You're right. First of all, I eliminated these child subqueries, but plan threw exceptions that plan does not have sufficient data channels. I feel that it would be better to change the implementation. therefore, I have created these even unnecessary data operations exist.
I agree with your thought, and I will try again for this optimization work.

ykrips:
@jihoonson , Thank you for review on this patch. I will add some comments on this patch, and will upload patches in a short time.

tajo-core/src/main/java/org/apache/tajo/engine/planner/global/GlobalPlanner.java
```diff
@@ -1371,32 +1373,57 @@ public LogicalNode visitUnion(GlobalPlanContext context, LogicalPlan plan, Logic
...
+        } else {
+          queryBlockBlocks.add(rightBlock);
+        }
       }
 
       ExecutionBlock execBlock;
       if (unionBlocks.size() == 0) {
-        execBlock = context.plan.newExecutionBlock();
+        if (leftUnion || rightUnion) {
+          execBlock = (!leftUnion && rightUnion)? rightBlock: leftBlock;
```
jihoonson: Would you add some descriptions for this line?

ykrips: This line is no more necessary. I will upload changes for this patch.

jihoonson:
Sorry for delayed review.
I'll review tonight.

jihoonson:
Hi @ykrips thanks for your nice patch.
It looks good, but I have another suggestion even though it is not directly related with this issue.

I think that ```union``` queries are popularly used for ETLs. So, I tested the following query. (It is a CTAS statement of the case 1 of TAJO-1368.)
```
create table test using parquet as 
  select * from 
    (select c_custkey, c_nationkey from customer where c_nationkey < 0 
    union all 
    select c_custkey, c_nationkey from customer where c_nationkey > 0 
    ) a 
  union all 
  select * from 
    (select c_custkey, c_nationkey from customer where c_nationkey < 0 
    union all 
    select c_custkey, c_nationkey from customer where c_nationkey > 0 
    ) b 
```
And, I found the following error.
```
2015-03-17 17:29:42,257 ERROR: org.apache.tajo.querymaster.QueryMasterTask (startQuery(368)) - 
java.lang.NullPointerException
	at org.apache.tajo.engine.planner.global.GlobalPlanner.buildNoPartitionedStorePlan(GlobalPlanner.java:1113)
	at org.apache.tajo.engine.planner.global.GlobalPlanner.buildStorePlan(GlobalPlanner.java:1037)
	at org.apache.tajo.engine.planner.global.GlobalPlanner.access$600(GlobalPlanner.java:68)
	at org.apache.tajo.engine.planner.global.GlobalPlanner$DistributedPlannerVisitor.visitCreateTable(GlobalPlanner.java:1570)
	at org.apache.tajo.engine.planner.global.GlobalPlanner$DistributedPlannerVisitor.visitCreateTable(GlobalPlanner.java:1141)
	at org.apache.tajo.plan.visitor.BasicLogicalPlanVisitor.visit(BasicLogicalPlanVisitor.java:126)
	at org.apache.tajo.plan.visitor.BasicLogicalPlanVisitor.visitRoot(BasicLogicalPlanVisitor.java:151)
	at org.apache.tajo.engine.planner.global.GlobalPlanner$DistributedPlannerVisitor.visitRoot(GlobalPlanner.java:1146)
	at org.apache.tajo.engine.planner.global.GlobalPlanner$DistributedPlannerVisitor.visitRoot(GlobalPlanner.java:1141)
	at org.apache.tajo.plan.visitor.BasicLogicalPlanVisitor.visit(BasicLogicalPlanVisitor.java:60)
	at org.apache.tajo.engine.planner.global.GlobalPlanner.build(GlobalPlanner.java:147)
	at org.apache.tajo.querymaster.QueryMasterTask.startQuery(QueryMasterTask.java:360)
	at org.apache.tajo.querymaster.QueryMasterTask.start(QueryMasterTask.java:175)
	at org.apache.tajo.querymaster.QueryMaster$QueryStartEventHandler.handle(QueryMaster.java:416)
	at org.apache.tajo.querymaster.QueryMaster$QueryStartEventHandler.handle(QueryMaster.java:403)
	at org.apache.hadoop.yarn.event.AsyncDispatcher.dispatch(AsyncDispatcher.java:173)
	at org.apache.hadoop.yarn.event.AsyncDispatcher$1.run(AsyncDispatcher.java:106)
	at java.lang.Thread.run(Thread.java:745)
```
This error seems to be caused by misfunctioning of ```GlobalPlanner.hasUnionChild()```.
I think it would be great if you fix this bug, too.

ykrips:
Good finding!
I also figure out this error.

jihoonson:
Thanks a lot!
The patch looks great.
I'll give +1 after CI test is finished.
It will be good to put your latest patch on Jira because Travis is too slow.

PR #404, opened 2015-03-10
TAJO-1383: Improve broadcast table cache

jinossy:
I’ve change to keep single broadcast table in memory


hyunsik:
It causes severe memory consumption problem and performance degradation if queries involve broadcast, especially in high-end servers which allow many task concurrency. I'll review it today night.

jihoonson:
The patch looks good in overall. However, it is difficult to understand details of some changes. 
Would you mind to give me a brief summarization of the changes?
In addition, it would be great if you add some comments.

jinossy:
@jihoonson Sure,
If a right table can broadcast, it does not share memory. so I’ve change to share the right table
In addition, I’ve fix a hidden bugs.
* Fix memory leak
* Add missing property in protobuf idl

Can you tell me lines ? I will be easy to add comments

tajo-core/src/main/java/org/apache/tajo/engine/planner/physical/ScanExec.java
```diff
@@ -0,0 +1,70 @@
...
+import org.apache.tajo.catalog.proto.CatalogProtos;
+import org.apache.tajo.engine.planner.enforce.Enforcer;
+import org.apache.tajo.ipc.TajoWorkerProtocol;
+import org.apache.tajo.worker.TaskAttemptContext;
+
+import java.io.IOException;
+import java.util.List;
+
+public abstract class ScanExec extends PhysicalExec {
+  private boolean canBroadcast;
```
jihoonson: Would you add some descriptions on this variable?
It makes me confusing.
Does it mean this scan is broadcasted or is a broadcast candidate?

tajo-core/src/main/java/org/apache/tajo/engine/utils/CacheHolder.java
```diff
@@ -0,0 +1,81 @@
...
+import java.util.List;
+import java.util.Map;
+
+public interface CacheHolder<T> {
+
+  T getData();
+  TableStats getTableStats();
+  void release();
+
+  public static class BroadcastCacheHolder implements CacheHolder<Map<Tuple, List<Tuple>>> {
```
jihoonson: Would you add some descriptions for this class?

tajo-core/src/main/java/org/apache/tajo/engine/utils/TableCache.java
```diff
@@ -0,0 +1,81 @@
...
+import com.google.common.collect.Lists;
+import com.google.common.collect.Maps;
+import org.apache.commons.logging.Log;
+import org.apache.commons.logging.LogFactory;
+import org.apache.tajo.ExecutionBlockId;
+
+import java.util.List;
+import java.util.Map;
+
+public class TableCache {
```
jihoonson: Would you add some descriptions on this class?

jihoonson:
@jinossy thanks for your comment. It is very helpful.
I left some comments for descriptions on classes and a variable.
If a table is broadcasted, it seems that all tasks in a node share the same cache. It would be great if you can give detailed descriptions on this cache mechanism.
Thanks!

jinossy:
I will add comments. 
Thanks

jinossy:
@jihoonson 
I've add some comments. Could you review again?

jihoonson:
+1 
Thanks for great work!

hyunsik:
This seems to be a performance improved issue. Intuitively, the changes look better than before. Nevertheless, it would be great if you share the benchmark results.

jinossy:
I was ran simple benchmark

* Cluster : 1Master + 4Worker
* TPC-H 100GB part of Q16

```
select 
  p_brand, p_type, p_size, ps_suppkey
from 
  partsupp ps join part p 
  on 
    p.p_partkey = ps.ps_partkey and p.p_brand <> 'Brand#45' 
    and not p.p_type like 'MEDIUM POLISHED%'
  join supplier_tmp s 
  on 
    ps.ps_suppkey = s.s_suppkey;
    
```

* This is only join benchmark (broadcast table is "supplier_tmp")

Broadcast | execution time
------------ | -------------
false | 45 sec
true | 33 sec
improved | 22 sec

hyunsik:
Here is my +1. The results look nice.

jihoonson:
The evaluation result looks great!

PR #405, opened 2015-03-10
TAJO-1380: Update JDBC documentation for new JDBC driver

dongjoon-hyun:


tajo-docs/src/main/sphinx/jdbc_driver.rst
```diff
@@ -9,11 +9,15 @@ In this section, we explain how to get JDBC driver and an example code.
 How to get JDBC driver
 =======================
 
+From Tajo Homepage
```
hyunsik: I have one trivial suggestion. How about 'Direct Download'? In my view, It would be better.

hyunsik:
Thank you for your work. The patch looks good to me. I leaved one trivial comment.

dongjoon-hyun:
Thank you for your feedback. No problem. Do you mean the following links? Which one do you prefer?

1. http://apache.org/dyn/closer.cgi/tajo/tajo-0.10.0/tajo-jdbc-0.10.0.jar

2. http://www.us.apache.org/dist/tajo/tajo-0.10.0/tajo-jdbc-0.10.0.jar

By the way, those have a version number in the link, so we should update carefully.

hyunsik:
I just meant the title name. The link 'http://tajo.apache.org/downloads.html' would be best because this page can track the number of downloads for each release.

dongjoon-hyun:
Oh, I see. You mean literally 'Direct Download' instead of 'From Tajo Homepage'.
Okay, I will change soon.

hyunsik:
Thank you :)

dongjoon-hyun:
Oh, I updated my pull request here (GitHub) and forgot to upload in Jira issue. I will update the attachement file in Jira issue soon.

PR #406, opened 2015-03-10
TAJO-1384: Duplicated output file path problem

jihoonson:


hyunsik:
The change looks straightforward. The second patch that eliminates unnecessary path list and unnecessary sort looks much better than the first one.

If all unit tests are passed, I'll finish the review.

PR #409, opened 2015-03-10
TAJO-1387: Correct error message for EXISTS clause

dongjoon-hyun:


hyunsik:
+1

The fix is straightforward.

PR #411, opened 2015-03-11
TAJO-1206 : implements thrift server

jerryjung:
For other service integration, Tajo must support multiple interface, Thrift Interfaces can be support multiple language such as python, cpp, ruby etc.
I implemented Thrift Server on Tajo. User can running Thrift server and monitoring it.


hyunsik:
Welcome,

We need to keep the code size small. It is very important because the code size is highly relative to the maintenance cost. BTW, this patch includes too many duplicated codes. You should try to find the minimum branch point to suppress code duplicates. 

Please refer this comment. I think that this approach is the best for us.
https://issues.apache.org/jira/browse/TAJO-1206?focusedCommentId=14229312

PR #413, opened 2015-03-11
TAJO-1360: VALUES_ field in OPTIONS table of catalog store should be longer.

charsyam:
I suggest OPTIONS size as 4000, because max size of  oracle varchar2 is just 4000
I researched docs for these patch.

maria: https://mariadb.com/kb/en/mariadb/varchar/
mysql: http://dev.mysql.com/doc/refman/5.7/en/char.html
postgres: http://www.postgresql.org/docs/8.3/static/datatype-character.html
derby: https://db.apache.org/derby/docs/10.7/ref/rrefsqlj41207.html
oracle: https://docs.oracle.com/cd/B28359_01/server.111/b28320/limits001.htm#i287903
(here varchar2 limit is just 4000)


hyunsik:
+1

Even though I suggested 8000 chars, 4000 chars also look enough. When we face the same problem due to limited length, we need to investigate new solution.

If there are no objection until tomorrow, I'll commit it.

dongjoon-hyun:
+1

ykrips:
Since I did move sql statements to xml files for Oracle and PostgreSQL DBMS, but these deprecated sql files for Oracle and PostgreSQL are still remained. Sorry for that. Can you apply these changes to /tajo-catalog/tajo-catalog-server/src/main/resources/schemas/oracle/oracle.xml and /tajo-catalog/tajo-catalog-server/src/main/resources/schemas/postgresql/postgresql.xml? I will create another issue to delete these deprecated sql files. Without this issue, I don't have any concerns on this patch.

hyunsik:
Thank you @ykrips for your comment. I missed them.

hyunsik:
Hi @charsyam,

Any progress? If you are busy now, I can finish this issue.

charsyam:
@hyunsik I will patch it soon ASAP.
actually, I forgot it, Thank you for reminding. :)

charsyam:
There is something wrong. so I will send it again.

charsyam:
@hyunsik @dongjoon-hyun @ykrips 
Please see https://github.com/apache/tajo/pull/436
I'm sorry to make you confused

PR #414, opened 2015-03-12
TAJO-1394 Support reconnect on tsql

navis:
Just reconnect on "invalid session" message and update session vars in local stash to server.


hyunsik:
+1

the patch looks good to me. I tested the patch in the case where TajoMaster is restarted. It works well. tsql reconnected to TajoMaster and got new session while keeping cached session variables.

hyunsik:
I also verified 'mvn clean install'.

PR #415, opened 2015-03-12
TAJO-1395: Remove deprecated sql files for Oracle and PostgreSQL

ykrips:
These sql files are still remained, and users can be confused. These files could be better to remove.


hyunsik:
+1
I've manually tested this patch with PostgreSQL. It works well.

PR #416, opened 2015-03-12
TAJO-1396 Unexpected IllegalMonitorStateException can be thrown in QueryInProgress

navis:
Simple / trivial patch

### DIFF
--- a/tajo-core/src/main/java/org/apache/tajo/master/QueryInProgress.java
+++ b/tajo-core/src/main/java/org/apache/tajo/master/QueryInProgress.java
@@ -95,7 +95,7 @@ public class QueryInProgress {
         queryMasterRpcClient.killQuery(null, queryId.getProto(), NullCallback.get());
       }
     } catch (Throwable e) {
-      catchException(e);
+      catchException("Failed to kill query " + queryId + " by exception " + e, e);
     } finally {
       writeLock.unlock();
     }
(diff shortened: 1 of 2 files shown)


tajo-core/src/main/java/org/apache/tajo/master/QueryInProgress.java
```diff
@@ -163,6 +168,12 @@ public void submmitQueryToMaster() {
 
     try {
       writeLock.lockInterruptibly();
+    } catch (Exception e) {
```
jihoonson: I have a question. 
Would you explain why writeLock.lockInterruptibly() is called subsequently?

navis: It just excluded locking from try/finally clause, avoiding releasing lock which is not held by the thread.

tajo-core/src/main/java/org/apache/tajo/master/QueryInProgress.java
```diff
@@ -163,6 +168,12 @@ public void submmitQueryToMaster() {
...
+      LOG.error(e.getMessage(), e);
+      return;
+    }
+    try {
+      writeLock.lockInterruptibly();
```
navis: Ah, you are saying on this. Forgot to remove it, sorry

navis:
Added more log messages

jinossy:
+1 This PR will remove the possible IllegalMonitorStateException

PR #417, opened 2015-03-13
TAJO-1393: Resolve findbug warnings on Tajo Cli Module

dongjoon-hyun:
Pass the following two steps.

* cd tajo-cli; mvn compile findbugs:findbugs
* mvn clean install -Pdist -Dtar -Dhadoop.version2.6.0

### DIFF
--- a/tajo-cli/src/main/java/org/apache/tajo/cli/tools/TajoGetConf.java
+++ b/tajo-cli/src/main/java/org/apache/tajo/cli/tools/TajoGetConf.java
@@ -119,6 +119,9 @@ public class TajoGetConf {
     } else if (hostName == null && port == null) {
       tajoClient = new TajoClientImpl(ServiceTrackerFactory.get(tajoConf));
     }
+    if (!tajoClient.isConnected()) {
+      System.out.println("ERROR: cannot connect Tajo server");
+    }
 
     processConfKey(writer, param);
     writer.flush();
--- a/tajo-cli/src/main/java/org/apache/tajo/cli/tools/TajoGetConf.java
+++ b/tajo-cli/src/main/java/org/apache/tajo/cli/tools/TajoGetConf.java
@@ -40,7 +40,6 @@ public class TajoGetConf {
   }
 
   private TajoConf tajoConf;
-  private TajoClient tajoClient;
   private Writer writer;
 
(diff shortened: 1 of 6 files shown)


dongjoon-hyun:
Hi, Travis CI build is done~

tajo-cli/src/main/java/org/apache/tajo/cli/tools/TajoGetConf.java
```diff
@@ -119,6 +119,9 @@ public void runCommand(String[] args, boolean tsqlMode) throws Exception {
     } else if (hostName == null && port == null) {
       tajoClient = new TajoClientImpl(ServiceTrackerFactory.get(tajoConf));
     }
+    if (!tajoClient.isConnected()) {
```
hyunsik: Actually, this class does not need TajoClient. I think that removing tajoClient is a better solution. So, you don't need the below ``System.out.print`` statement.

tajo-cli/src/main/java/org/apache/tajo/cli/tools/TajoHAAdmin.java
```diff
@@ -131,6 +131,9 @@ public void runCommand(String[] args) throws Exception {
     } else if (hostName == null && port == null) {
       tajoClient = new TajoClientImpl(ServiceTrackerFactory.get(tajoConf));
     }
+    if (!tajoClient.isConnected()) {
```
hyunsik: I think that this class also does not require TajoClient connection. Removing TajoClient would be a better solution to fix 'unused field' warning.

hyunsik:
I leave some comments. TajoHAAdmin seems to need refactoring. Its code is somewhat messy.

dongjoon-hyun:
Thank you. I will rebase and remove them.

dongjoon-hyun:
I've done and the followings are passed.
cd tajo-cli; mvn compile findbugs:findbugs
mvn clean install -Pparallel-test,hcatalog-0.12.0 -DLOG_LEVEL=ERROR -Dmaven.fork.count=2

hyunsik:
The patch looks good to me. I'll finish the review when CI is completed.

dongjoon-hyun:
Thanks, @hyunsik

PR #421, opened 2015-03-13
TAJO-1400: Add TajoStatement::setMaxRows method support

hys9958:
I implemented the setMaxRows and getMaxRows method support.
Please review my patch.

thanks.

### DIFF
--- a/tajo-client/src/main/java/org/apache/tajo/client/QueryClient.java
+++ b/tajo-client/src/main/java/org/apache/tajo/client/QueryClient.java
@@ -45,6 +45,10 @@ public interface QueryClient extends Closeable {
   public Map<String, String> getClientSideSessionVars();
 
   public String getBaseDatabase();
+  
+  public void setMaxRows(int maxRows);
+  
+  public int getMaxRows();
 
   @Override
   public void close();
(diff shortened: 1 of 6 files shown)


blrunner:
I'll review the patch today. :)

jihoonson:
Would you add some unit tests?
TestTajoJDBC will be good to add tests.

hys9958:
Ok, jihoonson,
I'll update the TestTajoJDBC for unit tests of this issue soon.

jihoonson:
+1
Thanks for your contribution!

jinossy:
@hys9958 
How do you think that move the setMaxRows() to TajoStatement ?

jihoonson:
@jinossy thanks for the good suggestion.
It would be a better implementation.

hys9958:
@jinossy Ok, It's better idea.
I'll re-patch the implementation of maxRows(). :)

jihoonson:
@hys9958 thanks. I'll wait for the next patch.

hys9958:
Hi @jihoonson ,
Query return value is 'java.sql.ResultSet' interfase in TajoStatement. It need to many refactoring.
And i'll also another patch about getMoreResults. When the getMoreResults patch time, i update maxRows.

jihoonson:
Ok. Here is my +1.
If there are no objections, I'll commit.

PR #422, opened 2015-03-14
TAJO-1359: Add nested field projector and language extension to project nested record.

hyunsik:
This work is still on going. So, please just review the direction. 

Actually, all features seem to work well. But, I still need to clean up and refactor the changes in the patch. Also, I'll add more unite tests to verify more use cases of complex types.  

For tests, I'm using a twitter example data set at https://dev.twitter.com/rest/reference/get/search/tweets. Projecting all fields including nested primitive fields works well.

Here is an example DDL and a query statement.
*DDL*
```sql
CREATE EXTERNAL TABLE tweets (
  coordinates TEXT,
  favorited BOOL,
  truncated BOOL,
  created_at TIMESTAMP,
  id_str TEXT,
  /*entrities RECORD (
    urls ARRAY<TEXT>
  )*/
  in_reply_to_user_id_str TEXT,
  contributors TEXT,
  text TEXT,
  metadata RECORD (
    iso_language_code TEXT,
    result_type TEXT
  ),
  retweet_count INTEGER,
  in_reply_to_status_id_str TEXT,
  id TEXT,
  geo TEXT,
  retweeted BOOL,
  in_reply_to_user_id TEXT,
  place TEXT,
  user RECORD (
    profile_sidebar_fill_color TEXT,
    profile_sidebar_border_color TEXT,
    profile_background_tile TEXT,
    name TEXT,
    profile_image_url TEXT,
    created_at TIMESTAMP,
    location TEXT,
    follow_request_sent TEXT,
    profile_link_color TEXT,
    is_translator BOOL,
    id_str TEXT,
    /*
    entities RECORD (
      url RECORD (
      ),
      description RECORD (
      )
    ), */
    default_profile BOOL,
    contributors_enabled BOOL,
    favourites_count INTEGER,
    url TEXT,
    profile_image_url_https TEXT,
    utc_offset INTEGER,
    id BIGINT,
    profile_use_background_image BOOL,
    listed_count INTEGER,
    profile_text_color TEXT,
    lang TEXT,
    followers_count INTEGER,
    protected BOOL,
    notifications TEXT,
    profile_background_image_url_https TEXT,
    profile_background_color TEXT,
    verified TEXT,
    geo_enabled TEXT,
    time_zone TEXT,
    description TEXT,
    default_profile_image TEXT,
    profile_background_image_url TEXT,
    statuses_count INTEGER,
    friends_count INTEGER,
    following TEXT,
    show_all_inline_media BOOL,
    screen_name TEXT
  ),
  in_reply_to_screen_name TEXT,
  source TEXT,
  in_reply_to_status_id TEXT
) USING JSON LOCATION ${table.path};
```

*DML*
```sql
SELECT
  coordinates,
  favorited,
  truncated,
  created_at,
  id_str,
  in_reply_to_user_id_str,
  contributors,
  "text",
  metadata.iso_language_code,
  metadata.result_type,
  retweet_count,
  in_reply_to_status_id_str,
  id,
  geo,
  retweeted,
  in_reply_to_user_id,
  place,
  user.profile_sidebar_fill_color,
  user.profile_sidebar_border_color,
  user.profile_background_tile,
  user.name,
  user.profile_image_url,
  user.created_at,
  user.location,
  user.follow_request_sent,
  user.profile_link_color,
  user.is_translator,
  user.id_str,
  user.default_profile,
  user.contributors_enabled,
  user.favourites_count,
  user.url,
  user.profile_image_url_https,
  user.utc_offset,
  user.id,
  user.profile_use_background_image,
  user.listed_count,
  user.profile_text_color,
  user.lang,
  user.followers_count,
  user.protected,
  user.notifications,
  user.profile_background_image_url_https,
  user.profile_background_color,
  user.verified,
  user.geo_enabled,
  user.time_zone,
  user.description,
  user.default_profile_image,
  user.profile_background_image_url,
  user.statuses_count,
  user.friends_count,
  user.following,
  user.show_all_inline_media,
  user.screen_name,
  in_reply_to_screen_name,
  source,
  in_reply_to_status_id
FROM
  tweets;
```

[Review on Reviewable](https://reviewable.io/reviews/apache/tajo/422)


hyunsik:
This patch also includes some refactoring of Schema class as follows:
 * Rename Schema::getColumns to Schema::getRootColumns.
   * It just returns only root columns in a schema.
 * Add Schema::getAllColumns, returning all flatten columns including nested fields.

hyunsik:
I used this site to convert jsons to json lines.
http://www.freeformatter.com/json-formatter.html

I think that it would be helpful to make test data sets.

jihoonson:
Thanks for the great work!
I'll review soon.

jihoonson:
I looked over the above queries. 
Basically, it looks good to support the nested schema. 
I have a question. 
It looks that a record column and a table have the same name. In this case, how can we identify them exactly? 
Additionally, I have one suggestion. 
As for tables, it would be good if we support the alias names for record types. It will be helpful when the record column has a long name. 
What do you think?

hyunsik:
Hi @jihoonson,

> It looks that a record column and a table have the same name. In this case, how can we identify them exactly? 

Here is how to resolve names.
https://github.com/apache/tajo/pull/422/files#diff-23481c6f9849708ac4a57d2252714846R312

There is a priority about resolving names. If a record column and a table, a table has a higher priority. To resolve this problem, users should use full qualified names.

Could you explain what alias for nested record is? Is it different from just column alias?

jihoonson:
@hyunsik thanks. It looks good.
On alias, I was missed it. My suggestion is exactly same with column alias.

jihoonson:
@hyunsik truly sorry for delayed review.
I'm reviewing, but it requires to be rebased.

tajo-catalog/tajo-catalog-common/src/main/java/org/apache/tajo/catalog/NestedPathUtil.java
```diff
@@ -0,0 +1,109 @@
...
+
+  /**
+   * Make a nested field path
+   *
+   * @param parts path parts
+   * @param startIndex startIndex
+   * @param depth Depth
+   * @return Path
+   */
+  public static String make(String [] parts, int startIndex, int depth) {
```
jihoonson: It would be nice if the method name is more intuitive.

tajo-core/src/main/java/org/apache/tajo/engine/planner/physical/SeqScanExec.java
```diff
@@ -186,6 +195,7 @@ private void initScanner(Schema projected) throws IOException {
     // set system default properties
     PlannerUtil.applySystemDefaultToTableProperties(context.getQueryContext(), meta);
 
+    // TODO - If fragments is null, we don't need to make any task. We should investigate it later
```
jihoonson: How about logging a debug message for later investigation?

tajo-plan/src/main/java/org/apache/tajo/plan/nameresolver/NameResolver.java
```diff
@@ -51,30 +70,45 @@
...
+      // if a table name is qualified
+      if (relation.getCanonicalName().equals(tableName) || relation.getTableName().equals(tableName)) {
+        found.add(relation);
+
+      // if a table name is not qualified
+      } else if (CatalogUtil.extractSimpleName(relation.getCanonicalName()).equals(tableName) ||
           CatalogUtil.extractSimpleName(relation.getTableName()).equals(tableName)) {
-        // obtain the database name
-        found.add(CatalogUtil.extractQualifier(relation.getTableName()));
+        // obtain the a qualified table name
```
jihoonson: This comment looks unnecessary.

tajo-plan/src/main/java/org/apache/tajo/plan/nameresolver/NameResolver.java
```diff
@@ -162,18 +210,22 @@ static Column resolveFromCurrentAndChildNode(LogicalPlan.QueryBlock block, Colum
...
    * @param block The current query block
-   * @param columnRef The column reference to be found
+   * @param columnName The column reference to be found
    * @return The found column
    */
-  static Column resolveFromAllRelsInBlock(LogicalPlan.QueryBlock block,
-                                          ColumnReferenceExpr columnRef) throws VerifyException {
+  static Column lookupColumnFromAllRelsInBlock(LogicalPlan.QueryBlock block,
+                                               String columnName) throws VerifyException {
+    Preconditions.checkArgument(CatalogUtil.isSimpleIdentifier(columnName));
```
jihoonson: It would be nice if you add an appropriate error message for pre-condition checking failure.

tajo-plan/src/main/java/org/apache/tajo/plan/nameresolver/NameResolver.java
```diff
@@ -31,14 +34,30 @@
...
+ * defined in different ways. Some column name indicates just a column in a relation.
+ * Another column name includes alias table name or alias column name, derived from some expression.
+ *
+ * This utility ensures that each column name is derived from valid and accessible column, and
+ * it also finds the exact data type of the column.
+ *
+ * Terminology:
+ * <ul>
+ *   <li>Qualifier:  database name, table name, or both included in a column name</li>
+ *   <li>Simple name: just column name without any qualifier</li>
```
jihoonson: I have a question. When a column ```C``` has an alias ```A```, is its simple name ```C``` or ```A```?

tajo-plan/src/main/java/org/apache/tajo/plan/nameresolver/NameResolver.java
```diff
@@ -240,33 +292,94 @@ static Column resolveAliasedName(LogicalPlan.QueryBlock block, ColumnReferenceEx
...
+    if (qualifierParts.length >= 1) {
+      RelationNode rel = lookupTable(block, qualifierParts[0]);
+      if (rel != null) {
+        guessedRelations.add(rel);
+        columnNamePosition = 1;
+      }
+    }
+
+    // column.nested_fieldX...
+    if (guessedRelations.size() == 0 && qualifierParts.length == 1) {
```
jihoonson: I have a question for the condition ```guessedRelations.size() == 0```. 
It looks to find every candidate relation, but stop finding if any relations are found for flat fields. I wonder the reason.

jihoonson: Oh, I missed the below part that throws an AmbiguousFieldException when there are two ore more candidate relations.

jihoonson:
@hyunsik, thanks for your nice work. I've left some trivial comments.
In addition, I met an error while running an example in the unit test as follows.
```
default> create external table test (title TEXT,   name RECORD (     first_name TEXT,     last_name TEXT   ) ) USING JSON LOCATION 'hdfs://localhost:7020/test/';
OK
default> \d
test
default> \d test

table name: default.test
table path: hdfs://localhost:7020/test
store type: JSON
number of rows: unknown
volume: 257 B
Options: 
	'text.serde'='org.apache.tajo.storage.json.JsonLineSerDe'

schema: 
title	TEXT
name	RECORD (first_name TEXT, last_name TEXT)


default> select * from test;
ERROR: java.lang.NullPointerException
```
Here is the log.
```
2015-05-08 01:47:37,239 INFO org.apache.tajo.master.GlobalEngine: Query: select * from test
2015-05-08 01:47:37,323 ERROR org.apache.tajo.master.GlobalEngine:
Stack Trace:
java.lang.NullPointerException
        at org.apache.tajo.catalog.DDLBuilder.buildSchema(DDLBuilder.java:76)
        at org.apache.tajo.catalog.SchemaUtil.toDisplayString(SchemaUtil.java:161)
        at org.apache.tajo.catalog.TypeDesc.toString(TypeDesc.java:66)
        at org.apache.tajo.catalog.Column.toString(Column.java:159)
        at java.lang.String.valueOf(String.java:2847)
        at java.lang.StringBuilder.append(StringBuilder.java:128)
        at org.apache.tajo.catalog.Schema.toString(Schema.java:503)
        at org.apache.tajo.plan.logical.ScanNode.getPlanString(ScanNode.java:246)
        at org.apache.tajo.plan.visitor.ExplainLogicalPlanVisitor.visitScan(ExplainLogicalPlanVisitor.java:195)
        at org.apache.tajo.plan.visitor.ExplainLogicalPlanVisitor.visitScan(ExplainLogicalPlanVisitor.java:32)
        at org.apache.tajo.plan.visitor.BasicLogicalPlanVisitor.visit(BasicLogicalPlanVisitor.java:108)
        at org.apache.tajo.plan.visitor.ExplainLogicalPlanVisitor.visitUnaryNode(ExplainLogicalPlanVisitor.java:130)
        at org.apache.tajo.plan.visitor.ExplainLogicalPlanVisitor.visitProjection(ExplainLogicalPlanVisitor.java:88)
        at org.apache.tajo.plan.visitor.ExplainLogicalPlanVisitor.visitProjection(ExplainLogicalPlanVisitor.java:32)
        at org.apache.tajo.plan.visitor.BasicLogicalPlanVisitor.visit(BasicLogicalPlanVisitor.java:69)
        at org.apache.tajo.plan.visitor.ExplainLogicalPlanVisitor.visitRoot(ExplainLogicalPlanVisitor.java:81)
        at org.apache.tajo.plan.visitor.ExplainLogicalPlanVisitor.visitRoot(ExplainLogicalPlanVisitor.java:32)
        at org.apache.tajo.plan.visitor.BasicLogicalPlanVisitor.visit(BasicLogicalPlanVisitor.java:60)
        at org.apache.tajo.plan.visitor.ExplainLogicalPlanVisitor.getBlockPlanStrings(ExplainLogicalPlanVisitor.java:74)
        at org.apache.tajo.plan.LogicalPlan.getLogicalPlanAsString(LogicalPlan.java:344)
        at org.apache.tajo.plan.LogicalPlan.getQueryGraphAsString(LogicalPlan.java:334)
        at org.apache.tajo.plan.LogicalPlan.toString(LogicalPlan.java:366)
        at org.apache.tajo.master.GlobalEngine.createLogicalPlan(GlobalEngine.java:275)
        at org.apache.tajo.master.GlobalEngine.executeQuery(GlobalEngine.java:186)
        at org.apache.tajo.master.TajoMasterClientService$TajoMasterClientProtocolServiceHandler.submitQuery(TajoMasterClientService.java:287)
```

hyunsik:
The main cause is that my patch does not support direct projection of record type. In other words, the current work only support projecting primitive type fields in nested records. I'll support this feature and flatten operator in another jira issue. Now, I added logical planner to prohibit the direct projection of record type.

---





<sup>**[tajo-core/src/main/java/org/apache/tajo/engine/planner/physical/SeqScanExec.java, line 198 \[r6\]](https://reviewable.io:443/reviews/apache/tajo/422#-Jo8-AMBbSEZOQS3d0Cy-r6-198)** ([raw file](https://github.com/apache/tajo/blob/c5263f4e276abe3e42dbf99bf9dd87e142f67539/tajo-core/src/main/java/org/apache/tajo/engine/planner/physical/SeqScanExec.java#L198)):</sup>
This was investigated at https://issues.apache.org/jira/browse/TAJO-1422. So, I'll remove this TODO comment.

---

<sup>**[tajo-plan/src/main/java/org/apache/tajo/plan/nameresolver/NameResolver.java, line 56 \[r6\]](https://reviewable.io:443/reviews/apache/tajo/422#-Jo8-AME6RwC628sqVQV-r6-56)** ([raw file](https://github.com/apache/tajo/blob/c5263f4e276abe3e42dbf99bf9dd87e142f67539/tajo-plan/src/main/java/org/apache/tajo/plan/nameresolver/NameResolver.java#L56)):</sup>
LogicalPreProcessor and LogicalPlanner changes all column names derived from relations are full qualified names. LogicalPlanner temporarily makes some short names during processing. So, the situation you mentioned won't occur in the current planner. 

---

<sup>**[tajo-plan/src/main/java/org/apache/tajo/plan/nameresolver/NameResolver.java, line 104 \[r6\]](https://reviewable.io:443/reviews/apache/tajo/422#-Jo8-AME6RwC628sqVQV-r6-104)** ([raw file](https://github.com/apache/tajo/blob/c5263f4e276abe3e42dbf99bf9dd87e142f67539/tajo-plan/src/main/java/org/apache/tajo/plan/nameresolver/NameResolver.java#L104)):</sup>
Done.

---

<sup>**[tajo-plan/src/main/java/org/apache/tajo/plan/nameresolver/NameResolver.java, line 223 \[r6\]](https://reviewable.io:443/reviews/apache/tajo/422#-Jo8-AME6RwC628sqVQV-r6-223)** ([raw file](https://github.com/apache/tajo/blob/c5263f4e276abe3e42dbf99bf9dd87e142f67539/tajo-plan/src/main/java/org/apache/tajo/plan/nameresolver/NameResolver.java#L223)):</sup>
This is only for developers misuse instead of users.

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/apache/tajo/422)

jihoonson:
Thanks. I've verified that join and aggregation queries work well. Really nice patch.
The one thing I'd like to comment is that it needs to improve the error message.
Here is an example case.
```
default> select title, name.first_name, name.last_name from test t1, test t2 where t1.name.last_name = t2.name.last_name;
ERROR: column "title" does not exist
default> select name.first_name, name.last_name from test t1, test t2 where t1.name.last_name = t2.name.last_name;
ERROR: column name name.first_name is ambiguous
```
It would be great if the ambiguous exception occurs for the first case.

---







Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/apache/tajo/422)

jihoonson:
+1 thanks for your nice work!

PR #423, opened 2015-03-16
TAJO-1130 Concurrent execution of independent execution blocks

navis:


dongjoon-hyun:
+1.
This will improve Tajo more faster!

navis:
Fixed problems in shuffle join.

jihoonson:
I'll review today.

tajo-common/src/main/java/org/apache/tajo/conf/TajoConf.java
```diff
@@ -315,6 +315,9 @@ public static int setDateOrder(int dateOrder) {
 
     $GROUPBY_MULTI_LEVEL_ENABLED("tajo.dist-query.groupby.multi-level-aggr", true),
 
+    // WARN "tajo.yarn-rm.parallel-task-runner-launcher-num" should be set enough to avoid deadlock
+    $QUERY_EXECUTE_PARALLEL_MAX("tajo.query.execute.parallel.max", 2),
```
jihoonson: There are a lot of remaining issues to fully support concurrent stage execution. One of the representative issues is resource scheduling. However, as you already know, our new resource scheduler which can schedule resources for multiple queries and stages is on-going work. So, we cannot handle this issue now.
Thus, even though this feature is still nice, it should be disabled by default. In other words, the default value of parallelism should be 1.

Additionally, this is related to the concurrent execution of multiple stages. So, it would be better if the name is intuitive. How about ```tajo.querymaster.stage.parallel-execution.max-num```?

jihoonson:
@navis thanks for your nice work!
In overall, it looks good. 
However, since this is a new feature, we need to test the patch thoroughly.
First of all, it is required to add some tests for this issue.
I think that you can enable your concurrent execution option for query tests. It will be easier than adding new test functions.

In addition, I wonder how much Tajo's performance is improved with this feature. Can you share any performance test results?

navis:
Yes, it's enabled just for test. I'm not implying this should be enabled by default. We can fix things that could be caused by this in SNAPSHOT term and disable it for release version, of course. 
For performance numbers, I should admit I've not done any serious tests with this. But when TAJO-1410 is to be included, I expect we can test this thoroughly.

jihoonson:
Sorry. There are some misunderstandings. I mean that unit tests should be added for this issue because it is difficult to expect what kinds of errors we will meet. 

However, I think it will be also difficult to write some new test functions to test only concurrent execution of multiple stages. So, I suggest adding a parameter to existing tests to test this feature. If so, we can test Tajo's query processing with and without concurrent execution. 
I think you already know the parameterized test, but you can find an example at TestJoinQuery. 

What do you think?

navis:
I don't want to double the test time, but you insists..

jihoonson:
It's just my suggestion. 
If you have any better ideas, please feel free to tell me.

hyunsik:
Hi guys,

I agree with all you guys. This feature still requires unit tests, yet I also don't want to double test time. We can compromise it.

First of all, this feature will completely work after we have a better scheduler. So, we need to mark this feature as experimental for a while. As navis said, we need to keep it disabled by default, and if necessary, users can enable this feature with session variables. 

With this approach, we may not need to verify all unit tests with this feature right now. We won't double the unit test time. Nevertheless, since this feature will require unit tests, it would be great if we test this this feature in only some unit tests like TestJoinQuery and TestUnionQuery, which mostly have independent execution blocks.

After we have a better scheduler, Tajo will use this feature by default. At that time, all unit tests verify this feature.

jihoonson:
Thanks. It looks a good compromising point. 

In addition, I think this feature should be disabled by default even in SNAPSHOT term because most developers don't want to be bothered by potential bugs which this feature can newly introduce.

navis:
ok. got it.

hyunsik:
I missed one point. If you want to enable it during SNAPSHOT term, I agree with navis's suggestion.

jihoonson:
Thanks. 
I had a direct chat with @hyunsik, and finally agree on enabling this feature by default during SNAPSHOT term. We can aggressively find hidden bugs.

hyunsik:
We need to be eager to fix the problems for SNAPSHOT term. If we face some very critical issues to interrupt other developments, we can disable it by default at that time.

Also, I think that this feature will have the following problem in the current implementation:
 * Resource allocation will be performed once when ExecutionBlock starts, and the allocated resources won't be changed during the execution block.
 * It means that some execution block can be slow if the execution starts with few resource capacity.

TAJO-1397 (https://issues.apache.org/jira/browse/TAJO-1397) will solve this problem, enabling continuous resource allocation while an execution block is running. I'll rush this issue.

dongjoon-hyun:
Hi, all. Is it ready to be included in SNAPSHOT now? :-)

jihoonson:
Uufortunately, not yet.
As discussed above, unit tests for this feature are required.
As @navis update his patch, I'll review as soon as possible.

hyunsik:
Didn't we agree to put this feature during SNAPSHOT in order to find aggressively hidden bugs? I remember like that.

jihoonson:
Sorry. You are right. It was my mistake.
If there aren't any further objections, I'll finish my review.

jinossy:
Guys, I think that we can disable this feature for a while in default. and we can enable this feature in TajoTestingCluster.
If this feature is stabled, we should enable this feature in default.

jihoonson:
It looks a good idea. We can find simple bugs with unit tests.
After making this feature more stable, we can enable it by default.

navis:
Ok, whatever. What should I do?

jihoonson:
As @jinossy said, it would be better if this feature is enabled only in unit tests by default. ```TajoTestingCluster::initPropertiesAndConfigs()``` looks good to enable it for unit tests. Additionally, you also should increase ```WORKER_RESOURCE_AVAILABLE_MEMORY_MB``` because only a single task can be executed with a query master due to the small size of the default available memory (1024 MB). Please note that a query master and a task use 512 MB by default, respectively.

navis:
done

jihoonson:
Thanks. I'll test.

jihoonson:
Travis CI seems to be failed due to the test hang.

navis:
Yes, seemed some part of tajo is not ready to launch two block simultaneously. I'll check that.

navis:
It seemed caused by tajo.yarn-rm.parallel-task-runner-launcher-num(I've commented this once and took half of a day to remind it, damn), which is also used for resource release. we should separate this on dedicate thread. I'll book this on lira. 
Confirmed in local. Let's see what travis says.

jihoonson:
I run a test with TPC-H Q10, and found a problem of the wrong query result.
Please see the following.
```
default> \set QUERY_EXECUTE_PARALLEL 1
default> select count(*) from (select c_custkey, c_name, sum(l_extendedprice * (1 - l_discount)) as revenue, c_acctbal, n_name, c_address, c_phone, c_comment from customer as c join nation as n on c.c_nationkey = n.n_nationkey join orders as o on c.c_custkey = o.o_custkey and o.o_orderdate >= '1993-10-01' and o.o_orderdate < '1994-01-01' join lineitem as l on l.l_orderkey = o.o_orderkey and l.l_returnflag = 'R' group by c_custkey, c_name, c_acctbal, c_phone, n_name, c_address, c_comment order by revenue desc) t;
Progress: 16%, response time: 2.185 sec
Progress: 16%, response time: 2.186 sec
Progress: 16%, response time: 2.588 sec
Progress: 16%, response time: 3.394 sec
Progress: 37%, response time: 4.399 sec
Progress: 100%, response time: 4.946 sec
?count
37967
(1 rows, 4.946 sec, 6 B selected)
default> \set QUERY_EXECUTE_PARALLEL 3
default> select count(*) from (select c_custkey, c_name, sum(l_extendedprice * (1 - l_discount)) as revenue, c_acctbal, n_name, c_address, c_phone, c_comment from customer as c join nation as n on c.c_nationkey = n.n_nationkey join orders as o on c.c_custkey = o.o_custkey and o.o_orderdate >= '1993-10-01' and o.o_orderdate < '1994-01-01' join lineitem as l on l.l_orderkey = o.o_orderkey and l.l_returnflag = 'R' group by c_custkey, c_name, c_acctbal, c_phone, n_name, c_address, c_comment order by revenue desc) t;
Progress: 25%, response time: 1.627 sec
Progress: 25%, response time: 1.628 sec
Progress: 25%, response time: 2.034 sec
Progress: 25%, response time: 2.837 sec
Progress: 35%, response time: 3.841 sec
Progress: 100%, response time: 4.398 sec
?count
1159
(1 rows, 4.398 sec, 5 B selected)
```

navis:
isExecutable() should be also checked for first() call. Took time to make env for TPC-H. Anyway I'm off today but dealing this in kids cafe. ^^

jihoonson:
Thanks for your effort for fast bug fix! I know that working in the kids cafe is not easy. 
Anyway, I'm happy with the valid query result.
Since this is an issue for the performance improvement, I'd like to see how fast Tajo is with this feature. So, I'll conduct a simple performance evaluation on a real cluster within the next few days.
If possible, would you mind testing, too?

babokim:
@jihoonson, I'll test on a real cluster.

navis:
Performance gain seemed hard to see if it's not running on very big cluster.

jihoonson:
I dont think so. If a query involves two or more independent ebs and there are sufficient amount of resources in the cluster, the query is expected to be faster with the concurrent execution of ebs.

jihoonson:
+1. This patch looks good to me.
I've simply evaluated performance gain with this patch as follows.

### Configurations and environment
* Broadcast join is disabled.
* The test is conducted on my desktop which equips 8 cores, 32 GB memory, and 2 HDDs.

### Data
* Input: TPC-H scale factor of 1

### Query
```
tpch> select sleep(1) from region, nation where n_regionkey = r_regionkey;
```

### Result
* with QUERY_EXECUTE_PARALLEL 1: 30.399 sec
* with QUERY_EXECUTE_PARALLEL 2: 25.571 sec
 * The performance is improved due to the two input relations are read concurrently.

PR #426, opened 2015-03-17
TAJO-1407 Minor performance improvement of MemSortExec

navis:
Simple micro benchmark shows 15~30% performance improvement, but this uses more memory. (Also can make little difference on appearance order of non-key values, caused by different sort implementation)

### DIFF
--- a/tajo-core/src/main/java/org/apache/tajo/engine/planner/physical/VectorizedSorter.java
+++ b/tajo-core/src/main/java/org/apache/tajo/engine/planner/physical/VectorizedSorter.java
@@ -1,0 +1,198 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.tajo.engine.planner.physical;
+
+import com.google.common.primitives.Booleans;
+import com.google.common.primitives.Doubles;
+import com.google.common.primitives.Floats;
+import com.google.common.primitives.Ints;
+import com.google.common.primitives.Longs;
+import com.google.common.primitives.Shorts;
+import org.apache.hadoop.util.IndexedSortable;
+import org.apache.hadoop.util.QuickSort;
+import org.apache.tajo.catalog.SortSpec;
+import org.apache.tajo.common.TajoDataTypes;
+import org.apache.tajo.datum.Datum;
+import org.apache.tajo.datum.TextDatum;
+import org.apache.tajo.exception.UnsupportedException;
+import org.apache.tajo.storage.Tuple;
+
+import java.util.BitSet;
+import java.util.Iterator;
+import java.util.List;
+
+/**
+ * Extract raw level values (primitive or String/byte[]) from each of key columns before sorting
+ * Uses indirection for efficient swapping
+ */
+public class VectorizedSorter implements IndexedSortable, TupleSorter {
+
+  private final Tuple[] tuples;         // source tuples
+  private final TupleVector[] vectors;  // values of key columns
+  private final int[] mappings;         // index indirection
+
+  public VectorizedSorter(List<Tuple> source, SortSpec[] sortKeys, int[] keyIndex) {
+    this.tuples = source.toArray(new Tuple[source.size()]);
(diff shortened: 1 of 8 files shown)


navis:
Removed JDK7 only method

jihoonson:
Looks interesting.
I'll review tonight.

tajo-core/src/main/java/org/apache/tajo/engine/planner/physical/VectorizedSorter.java
```diff
@@ -0,0 +1,191 @@
...
+import org.apache.tajo.datum.Datum;
+import org.apache.tajo.datum.TextDatum;
+import org.apache.tajo.exception.UnsupportedException;
+import org.apache.tajo.storage.Tuple;
+
+import java.util.BitSet;
+import java.util.Iterator;
+import java.util.List;
+
+public class VectorizedSorter implements IndexedSortable, TupleSorter {
```
jihoonson: Would you mind adding some descriptions on this class?

navis:
Added comments and now it supports interval type.

jihoonson:
I'll review this weekend.

tajo-core/src/main/java/org/apache/tajo/engine/planner/physical/VectorizedSorter.java
```diff
@@ -0,0 +1,210 @@
...
+  public VectorizedSorter(List<Tuple> source, SortSpec[] sortKeys, int[] keyIndex) {
+    this.tuples = source.toArray(new Tuple[source.size()]);
+    vectors = new TupleVector[sortKeys.length];
+    mappings = new int[tuples.length];
+    for (int i = 0; i < vectors.length; i++) {
+      TajoDataTypes.Type type = sortKeys[i].getSortKey().getDataType().getType();
+      boolean nullFirst = sortKeys[i].isNullFirst();
+      boolean ascending = sortKeys[i].isAscending();
+      boolean nullInvert = nullFirst && ascending || !nullFirst && !ascending;
+      vectors[i] = new TupleVector(TupleVector.getType(type), tuples.length, nullInvert, ascending);
```
jihoonson: TajoDataTypes.Type is the enum type of PB. Each type already has a corresponding integer.
type.getNumber() returns an integer value corresponding to the given type, and Type.valueOf() returns a type corresponding to the given integer value.
How about use these functions?

tajo-core/src/test/java/org/apache/tajo/engine/planner/physical/TestTupleSorter.java
```diff
@@ -0,0 +1,111 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.tajo.engine.planner.physical;
+
+import com.google.common.collect.Iterables;
```
jihoonson: Please remove unused imports.

tajo-core/src/test/java/org/apache/tajo/engine/planner/physical/TestTupleSorter.java
```diff
@@ -0,0 +1,111 @@
...
+import com.google.common.collect.Iterators;
+import org.apache.tajo.catalog.Column;
+import org.apache.tajo.catalog.Schema;
+import org.apache.tajo.catalog.SortSpec;
+import org.apache.tajo.common.TajoDataTypes;
```
jihoonson: Please remove unused imports.

jihoonson:
@navis, this is truly impressive work.
With my laptop, the vectorized sorter is faster than Collections.sort() about 25%!
I left some trivial comments.
In addition, it will be great if you add some tests to verify sort result.

tajo-core/src/main/java/org/apache/tajo/engine/planner/physical/VectorizedSorter.java
navis: It's possible but I prefer switch-case block to be compact as possible as it can be. And most of types in PB enum cannot be supported by this.

tajo-core/src/test/java/org/apache/tajo/engine/planner/physical/TestTupleSorter.java
navis: done.

navis:
Revised test case, which can be fail sometimes

tajo-core/src/main/java/org/apache/tajo/engine/planner/physical/VectorizedSorter.java
jihoonson: Ok. It looks good.

jihoonson:
I run the test 5 times, and it passed every time.
Would you share when the test fails?

navis:
2M rows sometimes made GCs, causing inversion of elapsed time.

jihoonson:
Oh, I think it's ok. The revised test doesn't contain any comparisons of elapsed times.
In addition, MemSortExec is executed only when there remains a sufficient amount of memory.
Here is my +1, and I'll commit it shortly.
Thanks for your great work!

jihoonson:
Oops, sorry. I have to wait the CI result. 
Would you put the latest patch on Jira? 
I'd like to commit this patch as soon as possible, and Jenkins will be much faster than Travis CI.

PR #428, opened 2015-03-18
TAJO-1408 Make IntermediateEntryProto more compact

navis:
Passed all tests


jihoonson:
I'll review today.

jinossy:
@navis 
It would be great if you share the improved compaction results.

hyunsik:
Hi @navis,

The patch looks good to me. The effect is obvious because it just changes ```List<Pair<Long,Long>>``` into a list of ```long``` values. Here is my +1.

There is one trivial things. You changed host to hostAddress, still meaning only hostname or ip address. But, the field includes ```hostname:port```. If you agree, I'd like to change it to just ```address```.

PR #430, opened 2015-03-18
TAJO-1409 Clients calling remote services returning BoolProto ignores false values

navis:
Now server throws exception to client instead of returning false.

### DIFF
--- a/tajo-core/src/main/java/org/apache/tajo/querymaster/QueryMasterManagerService.java
+++ b/tajo-core/src/main/java/org/apache/tajo/querymaster/QueryMasterManagerService.java
@@ -32,6 +32,8 @@ import org.apache.tajo.ipc.QueryMasterProtocol;
 import org.apache.tajo.ipc.TajoWorkerProtocol;
 import org.apache.tajo.master.container.TajoContainerId;
 import org.apache.tajo.master.event.*;
+import org.apache.tajo.rpc.InvocationFailure;
+import org.apache.tajo.rpc.RemoteCallException;
 import org.apache.tajo.session.Session;
 import org.apache.tajo.rpc.AsyncRpcServer;
 import org.apache.tajo.rpc.protocolrecords.PrimitiveProtos;
@@ -168,7 +170,7 @@ public class QueryMasterManagerService extends CompositeService
       done.run(TajoWorker.TRUE_PROTO);
     } catch (Exception e) {
       LOG.error(e.getMessage(), e);
-      done.run(TajoWorker.FALSE_PROTO);
+      throw new InvocationFailure(e);
     }
   }
 
@@ -193,7 +195,7 @@ public class QueryMasterManagerService extends CompositeService
       done.run(TajoWorker.TRUE_PROTO);
     } catch (Exception e) {
       LOG.error(e.getMessage(), e);
-      done.run(TajoWorker.FALSE_PROTO);
+      throw new InvocationFailure(e);
     }
   }
 
@@ -209,7 +211,7 @@ public class QueryMasterManagerService extends CompositeService
       done.run(TajoWorker.TRUE_PROTO);
     } catch (Exception e) {
       LOG.error(e.getMessage(), e);
-      done.run(TajoWorker.FALSE_PROTO);
+      throw new InvocationFailure(e);
     }
   }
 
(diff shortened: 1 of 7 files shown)


navis:
rebased to master

navis:
Rebased to master

tajo-core/src/test/java/org/apache/tajo/querymaster/TestKillQuery.java
```diff
@@ -63,7 +63,7 @@
   private static TajoTestingCluster cluster;
   private static TajoConf conf;
   private static TajoClient client;
-  private static String queryStr = "select t1.l_orderkey, t1.l_partkey, t2.c_custkey " +
+  private static String queryStr = "select t1.l_orderkey, t1.l_partkey, t2.c_custkey, sleep(1) " +
```
jihoonson: The problem of a test failure was fixed in https://issues.apache.org/jira/browse/TAJO-1138 and https://issues.apache.org/jira/browse/TAJO-1440.

tajo-core/src/main/java/org/apache/tajo/querymaster/QueryMasterManagerService.java
```diff
@@ -168,7 +170,7 @@ public void statusUpdate(RpcController controller, TajoWorkerProtocol.TaskStatus
       done.run(TajoWorker.TRUE_PROTO);
```
jihoonson: This line is also meaningless.

jihoonson:
+1
Thanks @navis. Your patch looks good to me.
I left trivial comments. If you don't mind, I'll reflect them before commit.

tajo-core/src/test/java/org/apache/tajo/querymaster/TestKillQuery.java
navis: good. I'll remove this line

tajo-core/src/main/java/org/apache/tajo/querymaster/QueryMasterManagerService.java
navis: I cannot understand this. Is it?

jihoonson: I mean, we can assume that the remote call is successfully finished if the InvocationFailure does not occur.

jinossy:
@jihoonson 
done.run must be called once.  If throwing exception is allowed, I think that we need to handle all case error instead of InvocationFailure

https://issues.apache.org/jira/browse/TAJO-292?focusedCommentId=13838973&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-13838973

@navis 
Thank you for your contribution. the patch is very useful but It seems to need failure test
Could you add unit test ?

navis:
@jinossy you are right. Because clients are calling this APIs with null controller, exception made in server side is not rethrown to client. Seemed need some more works.

navis:
@jinossy I have a question. 
````
    NettyClientBase client = getQueryMasterConnection();
    try {
      QueryMasterProtocol.QueryMasterProtocolService.Interface stub = client.getStub();
      stub.doneExecutionBlock(null, reporter, NullCallback.get());
    } finally {
      connPool.releaseConnection(client);
    }
````
thread calling stub.doneExecutionBlock() is exiting when it ends flushing packet to wire. Is it possible connPool.releaseConnection(client) to close the connection prematurely? Should we release connection when callback is called (with whatever from server)?

**Comment by jinossy** [comment]
@navis 
We don’t need release. Actually, releaseConnection is exception cleanup for reconnection.
some kind of codes is my fault. I've remove the legacy code in TAJO-1497

**Comment by jihoonson** [comment]
Thanks @jinossy for addressing missing points.

**Comment by jinossy** [comment]
@navis 
I will fix network hangs by TAJO-1563

**Comment by jinossy** [comment]
If you set an error message in controller, Async {{CallFuture.get()}} can throw exception.

like this example :
```
@Override
  public void done(RpcController controller, TajoWorkerProtocol.TaskCompletionReport report,
                   RpcCallback<PrimitiveProtos.BoolProto> done) {
    try {
      QueryMasterTask queryMasterTask = queryMaster.getQueryMasterTask(
          new QueryId(report.getId().getTaskId().getExecutionBlockId().getQueryId()));
      if (queryMasterTask != null) {
        queryMasterTask.getEventHandler().handle(new TaskCompletionEvent(report));
      }
      done.run(TajoWorker.TRUE_PROTO);
    } catch (Exception e) {
      LOG.error(e.getMessage(), e);
      controller.setFailed(e.getMessage());
      done.run(TajoWorker.FALSE_PROTO);
    }
  }
```

jinossy:
Solved by TAJO-1563, TAJO-1584
please close this PR.

PR #432, opened 2015-03-18
TAJO-1412 Add polling API in TajoMasterClientProtocol to wait result effectively

navis:


tajo-core/src/main/java/org/apache/tajo/master/QueryInfo.java
```diff
@@ -119,14 +119,37 @@ public int getQueryMasterClientPort() {
     return queryMasterClientPort;
   }
 
-  public TajoProtos.QueryState getQueryState() {
+  public synchronized TajoProtos.QueryState getQueryState() {
```
jihoonson: Declaring ```synchronized``` is redundant. QueryState is changed by only the QueryInProgress.

jihoonson:
Thanks for your contribution.
The patch looks good to me. I left one comment.
If you agree, I'll commit.

tajo-core/src/main/java/org/apache/tajo/master/QueryInfo.java
jihoonson: It seems that all ```synchronized``` declaration is redundant.

navis: queryState is volatile, so it does not need a lock. But for waitState method, I need a lock to synchronize. I just choosen "this" not to add any other object.

jihoonson: Oh, I got your point.

tajo-core/src/main/java/org/apache/tajo/master/QueryInfo.java
```diff
@@ -119,14 +119,37 @@ public int getQueryMasterClientPort() {
...
+  public synchronized void setQueryState(TajoProtos.QueryState queryState) {
     this.queryState = queryState;
+    notifyAll();
   }
 
+  public synchronized boolean waitState(TajoProtos.QueryState expect, long timeout)
+      throws InterruptedException {
+    long prev = System.currentTimeMillis();
+    while (timeout > 0 && expect != queryState && !isTerminalState()) {
+      wait(timeout);
```
jihoonson: I think that this approach has a potential problem of process hang. Once wait() is called, it can wake up only when setQueryState() is called. During query processing, setQueryState() is called only when TajoMaster receives heartbeats from workers. That means, if some heartbeats are lost, this process cannot wake up even though waiting time exceeds the given timeout. 

This will be not a big problem because we usually expect to receive at least one heartbeat from a lot of workers. However, I think that it is still a problem because the process's work can be different from our intention.

What do you think?

navis: wait(timeout) waits on lock for timeout msec. Isn't it?

jihoonson: Oh, you are right. I made so much mistakes due to lack of sleep.

jihoonson:
I'll finish my review soon.

jihoonson:
Hi @navis. I've reviewed this patch carefully.
Sorry to say, but I think that this is not appropriate for our architecture.
When a client calls ```pollQueryStatus()```, the process who waits for the query state to be changed is ```TajoMaster```'s rpc server. As you know, we will support multi-tenancy soon. In multi-tenant environment, master's rpc server will be getting slower as more clients submit queries and wait for their results. IMHO, polling in only client will be better than blocking master.
What do you think?

PR #434, opened 2015-03-18
TAJO-1403: Improve 'Simple Query' with only partition columns and constant values

dongjoon-hyun:


jihoonson:
I'll review this weekend.

dongjoon-hyun:
Finally! Thank you, @jihoonson . I'm working this weekend also, Please ask anything if you need.

tajo-storage/tajo-storage-hdfs/src/main/java/org/apache/tajo/storage/FileStorageManager.java
```diff
@@ -722,8 +737,18 @@ public void purgeTable(TableDesc tableDesc) throws IOException {
 
     List<FileStatus> nonZeroLengthFiles = new ArrayList<FileStatus>();
     if (fs.exists(tablePath)) {
-      getNonZeroLengthDataFiles(fs, tablePath, nonZeroLengthFiles, currentPage, numResultFragments,
-          new AtomicInteger(0), tableDesc.hasPartition(), 0, partitionDepth);
+      if (!partitionPath.isEmpty())
+      {
```
jihoonson: Would you move this parenthesis according to our code convention?

dongjoon-hyun: Oh, Sure.

tajo-plan/src/main/java/org/apache/tajo/plan/util/PlannerUtil.java
```diff
@@ -133,11 +134,90 @@ public static boolean checkIfSimpleQuery(LogicalPlan plan) {
           }
         }
       }
+
+      if (!noWhere && scanNode.getTableDesc().isExternal() && scanNode.getTableDesc().getPartitionMethod() != null) {
```
jihoonson: A managed table can also be partitioned. 
Would you explain why you added the isExternal() condition?

dongjoon-hyun: Yes, I developed handle both, but I found that Tajo does not support partitioned insert well.
So, I created the following issue, and change to handle external table only.
https://issues.apache.org/jira/browse/TAJO-1416

dongjoon-hyun: After fixing that issue, I planned to move to managed table creation. 
Actually, I don't know Tajo internal managed table file structure. 
I'm familiar with Hive only.

jihoonson:
@dongjoon-hyun, thanks for your nice work.
In overall, the patch looks good, but I want to check my understanding. 
If the ```WHERE``` clause involves only a subset of partition columns, it seems that every partition directory which are not specified in the ```where``` clause will be read. 
Please correct me if I misunderstand something.

tajo-plan/src/main/java/org/apache/tajo/plan/util/PlannerUtil.java
jihoonson: Ok. If so, would you add a TODO comment? 
It will be nice if it contains the specific issue number and description.

dongjoon-hyun:
You're right, but when I can construct prefix from the subset of partition column, this patch handle that as a simple query. 
/table_name/year=x/day=y/hour=z  (full path is simple query)
/table_name/year=x/day=y               (prefix path is constructed, so simple query)
/table_name/          /day=y/hour=z  (year is unknown, this is not simple query. It will be handled in an old fashion)
/table_name/year=x/         /hour=z  (day is unknown, this is not simple query, too.)

tajo-plan/src/main/java/org/apache/tajo/plan/util/PlannerUtil.java
dongjoon-hyun: No problem.

jihoonson:
Thanks. It's clear to me.
I left only trivial comments. When you reflect all of them, I'll give +1.
Anyway, I'm not pushing you, but do you have any plans to support the last two cases?

dongjoon-hyun:
For the last two cases, I thought it will need to eval on each tuple, so I'm not sure it's compatible with Tajo philosophy. Do you have some idea in mind?

dongjoon-hyun:
I changed the code according your comment. (Indentation and TODO documentation with issues)

jihoonson:
I think that it is possible to prune unnecessary partitions for the last two cases without hurting our philosophy as follows.
When visiting a partition directory, 
* if the current directory is leaf, read all files.
* otherwise,
 * if some conditions are given for the current partition, visit only qualified sub-directory.
 * otherwise, visit all sub-directory.

Anyway, here is my +1. I'll commit it.
However, I missed one more comment. EvalTreeUtil is the more proper class for checkIfPartitionSelection() and getPartitionValue() functions rather than PlannerUtil because every function related to EvalNode is located in that class.
If you don't mind, I'll move those functions to EvalTreeUtil before commit.

dongjoon-hyun:
Sure, please do that. I will learn from that. And when improving this further, I will start from there.
Thank you for your kind leading, @jihoonson

PR #439, opened 2015-03-19
TAJO-1420: Support GROUPING SET syntax

dongjoon-hyun:


dongjoon-hyun:
Here is the build log with the same command with Travis CI on my laptop.

https://app.box.com/s/wcim6qvcz87k3a2una4amnvohggkev65

jihoonson:
I recommend to implement the execution logic of the grouping set first. 

I know that we support only the syntax of data cube without execution logic. However, it is the leavings of the removed data cube operator execution logic. 
Honestly, we had code lines for data cube operator execution, but have removed them due to its low performance. However, we didn't removed the syntax part because we must support it again.

dongjoon-hyun:
Thank you for telling me the behind story, @jihoonson . I didn't know the cause of inconsistency before. 

By the way, in a social coding space, I think my small codes can be stepping stones to help someone who starts faster than me. So, we should break our big problems into small problems as possible as we can. Don't you think so? :)

IMO, if you open your code in your github and everyone can see your low-performance code, Tajo will get the better performance faster. Otherwise, you will see my or others' similar low-performance code again because nobody knows the performance cut-line. It would be a little bit waste of community efforts. What about making your code open, @jihoonson ?

In any way, I respect Tajo community's policy. I will keep this pull request and move on execution logic after finishing current ORC format issue.

Warmly,
Dongjoon.

dongjoon-hyun:
Hi, @jihoonson , may I start execution logic implementation by reusing this issue number? :)

jihoonson:
Definitely, we should break a big problem into smaller ones if possible and helpful. 
IMO, this is not a case of the above one. You already break the big problem, that is supporting grouping sets, into small ones including supporting the syntax. I mean, we should do more important and core work first. I think that this tradition is commonly shared by many development projects. When we develop something, we first design the whole architecture, and then break it into several modules. After that, we implement core modules first because trivial modules are not anythings without the core ones.

On opening the legacy code, it may be helpful as you said. However, whether the legacy code exists or not,  we should adopt an efficient algorithm. To do so, we should investigate many other projects and papers, and then discuss them. This is one of the main reasons of using Jira. 
Anyway, the performance of legacy code is not acceptable for practical workloads. So, it wouldn't be helpful so much.

Finally, you can reuse the issue if you want. Please update the issue title and description to be appropriate for the changed issue.

dongjoon-hyun:
I see. I agree with you on the purpose of Jira as a architectural design discussion tools before implementation. For this trivial issue, it was not so important. But if we move on Grouping Set issue, we need more discussion. Also, now I understand your original recommendation to implement the execution logic of the grouping set first. I will study the links (which you given) and read related projects/materials and update this issue for discussion before implementation. Thank you, @jihoonson. Always. :)

jihoonson:
Thanks for understanding. I'll wait for your work!

PR #441, opened 2015-03-20
TAJO-1426 Support "explain physical" to get physical plan

navis:
Quite useful


hyunsik:
It looks really useful. I'll review it soon.

jihoonson:
@hyunsik seems to be busy.
I'll review this weekend, too.

hyunsik:
Thank you @jihoonson. Actually, I'm. For few more weeks, I can spend only a little time.

jihoonson:
@navis thanks for your work. This will be definitely useful. 
However, it seems that the title sould be changed. I think that you already know, but would like to explain for clarity. 
In Tajo, there are three planning phases, i.e., logical planning phase, global planning phase, and physical planning phase. When a query is submitted, the logical planner generates a logical plan based on the query. And then, the global planner generates a master plan that explains how the query is executed in distributed manner. Once the master plan is generated, the query master executes execution blocks of that master plan subsequently. 
An execution block is executed by multiple workers. For distributed execution, each worker receives the logical plan corresponds to the currently executing execution block. Finally, a worker generates a physical plan based on the received logical plan. It indicates which types of physical operators are used to execute the logical plan. For example, given a logical plan for join, the physical plan indicates that the hash join or sorted-merge join will be used. 
Your work seems to see the the master (global) plan. So, it would be better to change the title. 

Anyway, I'm still reviewing. I'll finish soon.

tajo-core/src/test/java/org/apache/tajo/QueryTestCaseBase.java
```diff
@@ -317,6 +333,60 @@ public ResultSet executeQuery() throws Exception {
...
+    try {
+      for (int i = 0; i < queries.length; i++) {
+        ResultSet result = client.executeQueryAndGetResult(queries[i]);
+        Path resultPath = StorageUtil.concatPath(
+            currentResultPath, methodName + "." + String.valueOf(i + 1) + ".result");
+        if (currentResultFS.exists(resultPath)) {
+          assertEquals("Result Verification",
+              FileUtil.readTextFromStream(currentResultFS.open(resultPath)), resultSetToString(result).trim());
+        } else if (!isNull(result)) {
+          FileUtil.writeTextToStream(resultSetToString(result).trim(), currentResultFS.create(resultPath));
```
jihoonson: Would you explain the purpose of the else if clause?

tajo-core/src/test/java/org/apache/tajo/engine/query/TestSelectQuery.java
```diff
@@ -105,6 +105,20 @@ public final void testExplainSelect() throws Exception {
   }
 
   @Test
+  @SimpleTest(queries = {
```
jihoonson: This still looks good, but will be great if the following two problems are fixed.
First, it seems to be difficult to execute tests individually. 
Second, when a query test fails, it will be difficult to figure out which query failed.
Do you have any good idea?

jihoonson:
@navis your work is nice in overall. I left some comments.
Honestly, I'm writing tests to verify generated plans while working on improving join ordering (https://issues.apache.org/jira/browse/TAJO-1352). However, your test approach using the explain query looks better than mine. So, I'd like to share some challenges which I met.
* Left and right children of a commutative join (inner and cross) are interchangeable with each other. So, the plan can be changed.
* The order of shuffle keys between execution blocks can be changed.

These challenges should be addressed to validate query plans.

tajo-core/src/test/java/org/apache/tajo/QueryTestCaseBase.java
navis: In the first patch, it was just else { FileUtils.write..} ,which is for creating gold file for new test. In hive there are option (-Dtest.overwrite=true) but I couldn't find something like that in Tajo. So just I've made one if there is no result file expected.

Adding "isNull" to the clause in next patch was because I though some DDL queries can be used in here, which has no output result. For example,

SimpleTest(queres= {"create table x ...", "select * from x.."}, cleanup="x")

tajo-core/src/test/java/org/apache/tajo/engine/query/TestSelectQuery.java
navis: Yes, it's for simple tests, as the name implies. Serious test would need some more respect
 (separate test method). And for the second question, would it be enough adding index for fail message something like,
assertEquals("Result Verification for: " + (i+1) + "th test",

navis:
I think you can make some revise on this because I don't know how it(commutative join,  shuffle keys, etc.) should be shown. Doing that, you can also change the terms to more tajo-like way, because I'm not yet accustomed with terms in tajo.

tajo-core/src/test/java/org/apache/tajo/QueryTestCaseBase.java
jihoonson: Thanks. Honestly, I'm not familiar with this kind of data generation to verify the result.
It seems to be quite useful and can reduce human effort.

tajo-core/src/test/java/org/apache/tajo/engine/query/TestSelectQuery.java
jihoonson: Thanks. It looks reasonable.

jihoonson:
Thanks @navis. I will happily revise your work, and share with you.

Before starting revision, I have one more thing to talk with you.
IMHO, the simple test is not appropriate for query plan verifications. As you know, different query plans can produce the same result, but their efficiency can be significantly different. So, we must add a sufficient amount of tests to check whether Tajo's query planner always generates the best plan, especially for complex queries. 
So, I wonder what kinds of queries will be used for simple tests.
Do you have any good idea?

PR #442, opened 2015-03-20
TAJO-1430: Improve SQLAnalyzer by session-based parsing-result caching

dongjoon-hyun:
Please see the issue TAJO-1430 for the effect.

### DIFF
--- a/tajo-common/src/main/java/org/apache/tajo/conf/TajoConf.java
+++ b/tajo-common/src/main/java/org/apache/tajo/conf/TajoConf.java
@@ -203,6 +203,7 @@ public class TajoConf extends Configuration {
 
     // Query Configuration
     QUERY_SESSION_TIMEOUT("tajo.query.session.timeout-sec", 60, Validators.min("0")),
+    QUERY_SESSION_QUERY_CACHE_SIZE("tajo.query.session.query-cache-size-kb", 1024, Validators.min("0")),
 
     // Shuffle Configuration --------------------------------------------------
     PULLSERVER_PORT("tajo.pullserver.port", 0, Validators.range("0", "65535")),
--- a/tajo-core/src/main/java/org/apache/tajo/master/GlobalEngine.java
+++ b/tajo-core/src/main/java/org/apache/tajo/master/GlobalEngine.java
@@ -19,6 +19,10 @@
 package org.apache.tajo.master;
 
 import com.google.common.annotations.VisibleForTesting;
+import com.google.common.cache.CacheBuilder;
+import com.google.common.cache.CacheLoader;
+import com.google.common.cache.LoadingCache;
+import com.google.common.cache.Weigher;
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
 import org.apache.hadoop.service.AbstractService;
@@ -32,7 +36,9 @@ import org.apache.tajo.catalog.CatalogService;
 import org.apache.tajo.catalog.Schema;
 import org.apache.tajo.catalog.TableDesc;
 import org.apache.tajo.catalog.proto.CatalogProtos.StoreType;
+import org.apache.tajo.conf.TajoConf;
 import org.apache.tajo.engine.parser.SQLAnalyzer;
+import org.apache.tajo.engine.parser.SQLSyntaxError;
 import org.apache.tajo.engine.query.QueryContext;
 import org.apache.tajo.ipc.ClientProtos;
 import org.apache.tajo.master.TajoMaster.MasterContext;
@@ -54,6 +60,7 @@ import org.apache.tajo.util.CommonTestingUtil;
 
 import java.io.IOException;
 import java.sql.SQLException;
+import java.util.concurrent.TimeUnit;
 
 import static org.apache.tajo.ipc.ClientProtos.SubmitQueryResponse;
 
@@ -143,6 +150,25 @@ public class GlobalEngine extends AbstractService {
       newQueryContext.putAll(CommonTestingUtil.getSessionVarsForTest());
     }
 
+    // Set queryCache in session
+    int queryCacheSize = context.getConf().getIntVar(TajoConf.ConfVars.QUERY_SESSION_QUERY_CACHE_SIZE);
+    if (queryCacheSize > 0 && session.getQueryCache() == null) {
+      Weigher<String, Expr> weighByLength = new Weigher<String, Expr>() {
+        public int weigh(String key, Expr expr) {
+          return key.length();
+        }
+      };
+      LoadingCache<String, Expr> cache = CacheBuilder.newBuilder()
+        .maximumWeight(queryCacheSize * 1024)
+        .weigher(weighByLength)
+        .expireAfterAccess(1, TimeUnit.HOURS)
+        .build(new CacheLoader<String, Expr>() {
+          public Expr load(String sql) throws SQLSyntaxError {
+            return analyzer.parse(sql);
+          }
+        });
+      session.setQueryCache(cache);
+    }
     return newQueryContext;
   }
 
@@ -155,7 +181,7 @@ public class GlobalEngine extends AbstractService {
       if (isJson) {
         planningContext = buildExpressionFromJson(query);
       } else {
-        planningContext = buildExpressionFromSql(query);
+        planningContext = buildExpressionFromSql(query, session);
       }
 
       String jsonExpr = planningContext.toJson();
@@ -184,10 +210,22 @@ public class GlobalEngine extends AbstractService {
     return JsonHelper.fromJson(json, Expr.class);
   }
 
-  public Expr buildExpressionFromSql(String sql) throws InterruptedException, IOException,
+  public Expr buildExpressionFromSql(String sql, Session session) throws InterruptedException, IOException,
       IllegalQueryStatusException {
     context.getSystemMetrics().counter("Query", "totalQuery").inc();
-    return analyzer.parse(sql);
+    try {
+      if (session.getQueryCache() == null) {
+        return analyzer.parse(sql);
+      } else {
+        return (Expr) session.getQueryCache().get(sql.trim()).clone();
+      }
+    } catch (Exception e) {
+      if (e.getCause() instanceof SQLSyntaxError) {
+        throw (SQLSyntaxError) e.getCause();
+      } else {
+        throw new SQLSyntaxError(e.getCause().getMessage());
+      }
+    }
   }
 
   public QueryId updateQuery(QueryContext queryContext, String sql, boolean isJson) throws IOException,
--- a/tajo-core/src/main/java/org/apache/tajo/session/Session.java
+++ b/tajo-core/src/main/java/org/apache/tajo/session/Session.java
@@ -18,11 +18,13 @@
 
 package org.apache.tajo.session;
 
+import com.google.common.cache.LoadingCache;
 import com.google.common.collect.ImmutableMap;
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
 import org.apache.tajo.QueryId;
 import org.apache.tajo.SessionVars;
+import org.apache.tajo.algebra.Expr;
 import org.apache.tajo.master.exec.NonForwardQueryResultScanner;
 import org.apache.tajo.util.KeyValueSet;
 import org.apache.tajo.common.ProtoObject;
@@ -40,6 +42,7 @@ public class Session implements SessionConstants, ProtoObject<SessionProto>, Clo
   private String currentDatabase;
   private final Map<String, String> sessionVariables;
   private final Map<QueryId, NonForwardQueryResultScanner> nonForwardQueryMap = new HashMap<QueryId, NonForwardQueryResultScanner>();
+  private LoadingCache<String, Expr> cache;
 
   // transient status
   private volatile long lastAccessTime;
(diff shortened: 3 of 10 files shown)


dongjoon-hyun:
The cache is maintained by approximately LRU manner with 200 size and also cached item will be expired after 1 hour since last access.
'''
session.setQueryCache(CacheBuilder.newBuilder()
          .maximumSize(200)
          .expireAfterAccess(1, TimeUnit.HOURS)
          .build(new CacheLoader<String, Expr>() {
            public Expr load(String sql) throws SQLSyntaxError {
              return analyzer.parse(sql);
            }
      })
);
'''

dongjoon-hyun:
Rebased.

dongjoon-hyun:
Rebased.

dongjoon-hyun:
Hmm. I attached the test pass result of the following command. 
```
mvn clean install -Pparallel-test,hcatalog-0.12.0 -DLOG_LEVEL=ERROR -Dmaven.fork.count=2 > TAJO-1430.travis.log.txt
```
https://app.box.com/s/rx85b6pv26sgo400e5dxf30bepyr86nl

Currently, Travis CI seems to be unstable and the builds fail on other reasons like the following.
```
ERROR: org.apache.tajo.master.rm.TajoWorkerResourceManager (run(346)) - java.lang.InterruptedException
2015-04-04 10:12:39,174 ERROR: org.apache.tajo.util.history.HistoryWriter (writeHistory(318)) - Error while saving query history: q_1428141094535_0543:Filesystem closed
```

tajo-core/src/main/java/org/apache/tajo/master/GlobalEngine.java
```diff
@@ -155,7 +174,7 @@ public SubmitQueryResponse executeQuery(Session session, String query, boolean i
       if (isJson) {
         planningContext = buildExpressionFromJson(query);
```
jihoonson: It would be better if the cache is supported for json queries.

tajo-core/src/main/java/org/apache/tajo/session/Session.java
```diff
@@ -121,6 +124,10 @@ public synchronized String getCurrentDatabase() {
     return currentDatabase;
   }
 
+  public synchronized void setQueryCache(LoadingCache<String, Expr> cache) { this.cache = cache; }
```
jihoonson: The cache instance is always updated instead of adding a cache item to a singleton cache instance. I don't think that this is a valid cache operation.

In addition, please follow our coding convention.

tajo-core/src/main/java/org/apache/tajo/master/GlobalEngine.java
```diff
@@ -143,6 +148,20 @@ private QueryContext createQueryContext(Session session) {
       newQueryContext.putAll(CommonTestingUtil.getSessionVarsForTest());
     }
 
+    // Set queryCache in session
+    synchronized (session) {
```
jihoonson: Please remove the redundant synchronized block.

jihoonson:
Thanks for the contribution.
I left some comments.

tajo-core/src/main/java/org/apache/tajo/master/GlobalEngine.java
dongjoon-hyun: Is it slow?

tajo-core/src/main/java/org/apache/tajo/master/GlobalEngine.java
dongjoon-hyun: No problem.

tajo-core/src/main/java/org/apache/tajo/session/Session.java
dongjoon-hyun: That's not true. According to @hyunsik 's advice,it's a sesstion based cache. It is set once for each sesstion.

dongjoon-hyun:
Thank you, @jihoonson . By the way, could you see the Jira, too? There was some discussion about this.

https://issues.apache.org/jira/browse/TAJO-1430

It's just I want to make sure what I understand. 
By the way, sorry for the outdated title of this pull request outdate. 
I'll update now according to the Jira title.

tajo-core/src/main/java/org/apache/tajo/master/GlobalEngine.java
jihoonson: Honestly, I didn't estimate the exact time. But, using cache will be much faster than parsing huge json queries, I think.

tajo-core/src/main/java/org/apache/tajo/session/Session.java
jihoonson: I've already read Jira. I concerned about whether the cache is maintained within a session. But it was my misunderstanding. I'm not familiar with LoadingCache. Sorry. 

Anyway, it would be better if this cache initialization is moved to Session.

dongjoon-hyun: Currently, in this patch, I try to minimize dependency change in order to keep Tajo's good modularity. So, I put only Expr dependency into Session class. (It's inevitable.) If we put the cache initialization there, we need "import org.apache.tajo.engine.parser.SQLAnalyzer" there. I am not sure that Tajo allows this kind of dependency. If you think this is okay, I prefer to move that into Session. I need your guide. :)

jihoonson: Ok. Let's keep it. We can refactor later.

dongjoon-hyun:
@jihoonson , I changed this patch according to your advice except JSON.
I'm still not sure if JSON parser has performance issue. I think it's a little bit beyond this issue's scope.

tajo-core/src/main/java/org/apache/tajo/master/GlobalEngine.java
```diff
@@ -143,6 +148,18 @@ private QueryContext createQueryContext(Session session) {
       newQueryContext.putAll(CommonTestingUtil.getSessionVarsForTest());
     }
 
+    // Set queryCache in session
+    if (session.getQueryCache() == null) {
+      session.setQueryCache(CacheBuilder.newBuilder()
+        .maximumSize(200)
```
jihoonson: Here, the size means the number of items contained in the cache. So, the actual cache size depends on the size of cached queries. This will have a problem of potentially exhausting memory. So, it would be better to set the maximum weight rather than the maximum size.

In addition, I think that the maximum weight and the expiration period should be configurable. ```ConfVar``` would be a good place because we don't need to maintain different configurations for each user.

jihoonson:
Ok. I'll investigate the json parsing performance. If you have any references for that, please share with me.

I left another comment.
Thanks.

tajo-core/src/main/java/org/apache/tajo/master/GlobalEngine.java
dongjoon-hyun: That's a good idea. No problem. By the way, the weight should be estimated by SQL query string length, right? Or, do you have some Util to measure Expr memory consumption? Up to now, I can not find that kind of code in Expr. Do you think SQL query string length is okay?

dongjoon-hyun:
In fact, I don't know any use cases of JSON in the real fields. Is it popular to use JSON query?

jihoonson:
We support json queries to address some requests to integrate other olap tools. As you know, generating json is much easier than generating sql. 

On the weight, the query length looks good.

dongjoon-hyun:
According to your advice, I updated and finished testing the code. In these days, Travis CI is prone to fail. If you need, I will upload my tested log.

dongjoon-hyun:
Travis CI fails. Here is my succeeded test log.
```
mvn clean install -Pparallel-test,hcatalog-0.12.0 -DLOG_LEVEL=WARN -Dmaven.fork.count=2 > TAJO-1430.150407.log.txt
```
https://app.box.com/s/s7ivihn6myvpr2sbr0t6cfd8fb55zawj

jihoonson:
Thanks. Would you mind triggering Jenkins by putting your patch on Jira?
Jenkins is better than Travis because it also includes the Findbugs test.

dongjoon-hyun:
Sure!

dongjoon-hyun:
Hi, @jihoonson . Jenkins fails to start the test. 

```
Compiling /home/jenkins/jenkins-slave/workspace/PreCommit-TAJO-Build/incubator-tajo
/home/jenkins/tools/maven/latest/bin/mvn clean test -DskipTests -Phcatalog-0.12.0 > /home/jenkins/jenkins-slave/workspace/PreCommit-TAJO-Build/patchprocess/trunkJavacWarnings.txt 2>&1
Trunk compilation is broken?
```

What can I do now? Please visit the following link for detail info.

https://builds.apache.org/job/PreCommit-TAJO-Build/719/console

dongjoon-hyun:
Hmm. Let me check for a second. I found the following Jenkins logs.
```
[ERROR] COMPILATION ERROR : 
[INFO] -------------------------------------------------------------
[ERROR] /home/jenkins/jenkins-slave/workspace/PreCommit-TAJO-Build/incubator-tajo/tajo-core/src/main/java/org/apache/tajo/master/GlobalEngine.java:[160,13] setQueryCache(com.google.common.cache.LoadingCache<java.lang.String,org.apache.tajo.algebra.Expr>) in org.apache.tajo.session.Session cannot be applied to (com.google.common.cache.LoadingCache<java.lang.Object,org.apache.tajo.algebra.Expr>)
[INFO] 1 error
```

dongjoon-hyun:
For JDK6, I split one function call line into two lines.

tajo-common/src/main/java/org/apache/tajo/conf/TajoConf.java
```diff
@@ -203,6 +203,7 @@ public static int setDateOrder(int dateOrder) {
 
     // Query Configuration
     QUERY_SESSION_TIMEOUT("tajo.query.session.timeout-sec", 60, Validators.min("0")),
+    QUERY_SESSION_CACHE_SIZE("tajo.query.session.cache-size", 1000000, Validators.min("1000000")),
```
jihoonson: I have a couple of comments on this line.

* The cache size seems to be specified in bytes. This will be difficult for humans to specify the exact cache size. How about using the size unit of KB? 
* The session variable name looks too general even though this cache is only for parsed queries. It would be better if users can figure out what will be configured from its name. Also, the size unit should be included in the name like ```tajo.task.size-mb```.
* The minimum cache size is 1 GB. Do you have any reasons?
* In addition to the size configuration, it would be great if users can turn off/on this cache feature. This is because the cache may be useless in some workloads such as ad-hoc analysis.

dongjoon-hyun: Hi, @jihoonson . Thank you for kind advice.
* What about *tajo.query.session.query-cache-size-mb*, then? I think it would be more proper according to your advice.
* The minimum cache size is *1MB* in terms of SQL String length. As you see in TAJO-1430 example, 100K-length query takes over 30 seconds in SQL parsing, so the cache will hold up to 10 x 100K queries accessed in last 1 hour. In real situation, 10 or more MB is needed.
* For the cache on/off feautre, is it okay by using the condition *tajo.query.session.query-cache-size-mb=0*?

jihoonson: Oh, right. I meant 1MB not 1GB. According to your comment, it takes quite long time to parse queries even though their size is a few KBs. So, I think that KB is appropriate for the unit of size. 

On minimum cache size, ```Validators.min("1000000")``` means that the cache size under 1MB is an invalid value for this configuration. I think that you didn't intend that, but if so, please tell me the reason.

On the cache on/off, your suggestion looks good. In addition to your suggestion, it will be good if we can avoid checking the cache when the configured cache size is 0. I think that this cache will not be used in many cases because the cached data can be used only when the exactly same queries are submitted repeatedly.

dongjoon-hyun: 1. *tajo.query.session.query-cache-size-kb*, then? No problem.
2. On minimum cache size, let say 1K-size cache holding 10 x 100Byte SQL queries. The effect is small. In this case, we had better off the cache. In that sense, we need to guide the effective cache size practically. 
3. If cache is off, the cache will be null and Tajo will avoid checking the cache. I agree that it's important.
4. This query cache will be important especially for *PreparedStatement* of TAJO-1435. (I'm working on this too.) PreparedStatement is popular in real enterprise environments. I did add PLACE_HOLDER in SQL syntax for '?' of PreparedStatement and am trying to replace them after cloning Expr. (Anyway, this is beyond the scope of this issue).

jihoonson: Thanks for your rapid reply. Here are my answers.
1. Thanks for understanding.
2. I mean, with ```Validators.min("1000000")```, Tajo won't accept _tajo.query.session.query-cache-size-mb=0_. Validators should be used to validate the configuration. So, IMO, ```Validators.min(0)``` is better to prohibit negative values.
3. Thanks for understanding.
4. I also think that query caching is important. I mean, the current implementation will not be popularly used because the cached data can be used only when the exactly same queries are submitted repeatedly.

dongjoon-hyun: 2. You're right. Validators.min(0) is proper now because we use 0 for cache off option, too.
4. I see. Current implementation does. 
Now, I will update the code soon. Thank you for reviewing, always!

jihoonson: Thanks for your understanding!

dongjoon-hyun:
I changed the code and finished testing on my laptop.

jihoonson:
Thanks. I think that this work is almost done with the latest patch. 
I'll commit tonight after some tests.

dongjoon-hyun:
Thank you, @jihoonson .

jihoonson:
+1 LGTM!
Thanks for your contribution.
I think that it would be great if there are any test cases, but it can be added at https://issues.apache.org/jira/browse/TAJO-1435.
I'll commit shortly.

PR #448, opened 2015-03-22
TAJO-1284: Add alter partition method to CatalogStore.

blrunner:
This patch will implement ALTER TABLE ADD PARTITION method and ALTER TABLE DROP PARTITION method to CatalogStore.


tajo-catalog/tajo-catalog-common/src/main/java/org/apache/tajo/catalog/partition/PartitionKey.java
```diff
@@ -0,0 +1,148 @@
...
+  public void setColumnName(String columnName) { this.columnName = columnName; }
+
+  public int hashCode() {
+    return Objects.hashCode(partitionValue, columnName);
+  }
+
+  public boolean equals(Object o) {
+    if (o instanceof PartitionKey) {
+      PartitionKey another = (PartitionKey) o;
+      boolean eq = ((columnName != null && another.columnName != null
```
jihoonson: How about to use TUtil.checkEquals()?

tajo-catalog/tajo-catalog-common/src/main/proto/CatalogProtos.proto
```diff
@@ -281,8 +282,7 @@ message SortSpecProto {
 
 
 message PartitionsProto {
-  required TableIdentifierProto tableIdentifier = 1;
-  repeated PartitionDescProto partition = 2;
+  repeated PartitionDescProto partition = 21;
```
jihoonson: This will not induce any problems, but it would be better to change to 2.

tajo-catalog/tajo-catalog-common/src/main/proto/CatalogProtos.proto
```diff
@@ -293,10 +293,21 @@ message PartitionMethodProto {
 }
 
 message PartitionDescProto {
-  optional string partitionName = 2;
-  required int32  ordinalPosition = 3;
-  optional string partitionValue = 4;
-  optional string path = 5;
+  required string partitionName = 1;
+  repeated PartitionKeyProto partitionKeys = 32;
```
jihoonson: This will not induce any problems, but it would be better to change to 2.

jihoonson:
I'm still reviewing. I'll finish soon.

blrunner:
Thanks @jihoonson 
I've just fixed the patch by your comments.
If you review the patch continuously, it would be pleasure to me. :)

tajo-catalog/tajo-catalog-common/src/main/java/org/apache/tajo/catalog/CatalogConstants.java
```diff
@@ -48,11 +48,16 @@
...
   public static final String TB_PARTITION_METHODS = "PARTITION_METHODS";
   public static final String TB_PARTTIONS = "PARTITIONS";
+  public static final String TB_PARTTION_KEYS = "PARTITION_KEYS";
 
   public static final String COL_TABLESPACE_PK = "SPACE_ID";
   public static final String COL_DATABASES_PK = "DB_ID";
   public static final String COL_TABLES_PK = "TID";
   public static final String COL_TABLES_NAME = "TABLE_NAME";
+
+  public static final String COL_PARTITIONS_PK = "PID";
```
jihoonson: I think that it would be better if the name is more intuitive.

tajo-catalog/tajo-catalog-common/src/main/java/org/apache/tajo/catalog/partition/PartitionDesc.java
```diff
@@ -50,15 +73,21 @@ public PartitionDesc(CatalogProtos.PartitionDescProto proto) {
...
+      this.partitionKeys.add(partitionKey);
     }
+
     if(proto.hasPath()) {
       this.path = proto.getPath();
     }
   }
 
+  public String getPartitionName() { return partitionName; }
+  public void setPartitionName(String partitionName) { this.partitionName = partitionName; }
```
jihoonson: Is this function different with the below setName() function?
In addition, please move brackets according to our coding convention.

tajo-catalog/tajo-catalog-server/src/main/java/org/apache/tajo/catalog/store/MemStore.java
```diff
@@ -498,37 +498,16 @@ public void dropPartitionMethod(String databaseName, String tableName) throws Ca
   }
 
   @Override
-  public void addPartitions(CatalogProtos.PartitionsProto partitionDescList) throws CatalogException {
+  public List<CatalogProtos.PartitionDescProto> getPartitions(String databaseName, String tableName) throws CatalogException {
     throw new RuntimeException("not supported!");
```
jihoonson: MemStore is used for tests, so we need to implement this function.

tajo-catalog/tajo-catalog-server/src/main/java/org/apache/tajo/catalog/store/MemStore.java
```diff
@@ -498,37 +498,16 @@ public void dropPartitionMethod(String databaseName, String tableName) throws Ca
...
   }
 
   @Override
-  public void addPartition(String databaseName, String tableName, CatalogProtos.PartitionDescProto
-      partitionDescProto) throws CatalogException {
+  public CatalogProtos.PartitionDescProto getPartition(String databaseName, String tableName,
+                                                       String partitionName) throws CatalogException {
     throw new RuntimeException("not supported!");
```
jihoonson: MemStore is used for tests, so we need to implement this function.

tajo-catalog/tajo-catalog-server/src/main/resources/schemas/postgresql/postgresql.xml
```diff
@@ -151,7 +151,6 @@ xsi:schemaLocation="http://tajo.apache.org/catalogstore ../DBMSSchemaDefinition.
   				PID INT NOT NULL PRIMARY KEY,
   				TID INT NOT NULL,
   				PARTITION_NAME VARCHAR(128),
-  				ORDINAL_POSITION INT NOT NULL,
```
jihoonson: This schema is not compatible with that of other databases.
It looks that some changes are missed.

jihoonson:
@blrunner thanks for your nice patch.
I left some comments.
Would you add a table for partition keys to ```information_schema```?
In addition, it would be great if you add more tests. TestAlterTable looks good for testing.

tajo-catalog/tajo-catalog-common/src/main/java/org/apache/tajo/catalog/CatalogConstants.java
blrunner: I followed existing naming rule of CatalogConstants. I think there are three options as follows.
 - COL_PARTITIONS_PK
 - COL_PARTITIONS_ID
 - PARTITIONS_ID

What's your opinion about above options? :)

jihoonson: IMO, ```PARTITION_ID``` looks good.

blrunner: Okay, I also think that it is more simple and more intuitive. :)
And how about rename existing variables as following:
- COL_TABLESPACE_PK  --> TABLESPACE_ID

blrunner: @jihoonson 

I misunderstood your first comments.
I'll rename the primary key name PARTITION_ID. :)

blrunner:
@jihoonson 

Thanks for your detailed review.
I updated the patch according to your opinion.
And I will add more unit test cases for TestAlterTable with another issue. 
Please refer the following site: 
https://issues.apache.org/jira/browse/TAJO-1345

jihoonson:
Ok. 
Here is my +1. 
Ship it!

PR #451, opened 2015-03-23
TAJO-1440: Some tests fail in parallel test environment in TestKillQuery

eminency:
Currently, state variable is exposed directly to outside.
Thus, changing state operation is out of locking scope and consistency may be broken.

Changing state should be protected by acquiring write lock.


blrunner:
Thanks for your contribution.
I'll review the patch tomorrow. :)

PR #452, opened 2015-03-23
TAJO-1419: Tsql session command doesn't work

charsyam:
This has some problems.
1] case sensitive.
 -> Tajo Session Vars distinguish session key name.
 -> so "\set timezone GMT+1" is not valid command
 -> "\set TIMEZONE GMT+1" is valid.

2] client side var
 -> currently client side vars are stored in TajoCliContext with set command
 -> but ResultSet just using client session information that are stored in Server Session.
 -> and "SET TIME ZONE 'GMT+1'" command save TIMEZONE Setting as Server Session.
   so If timezone value is stored in client side, it is strange.
 
so, I changed TIMEZONE from CLISIDE_VAR to SERVER_VAR.


jihoonson:
Hi @charsyam, sorry for late review.

In Tajo, there are three different configurations for timezone. One of them is the session configuration which you removed. We should not remove it.
For more detailed explanation, please see http://tajo.apache.org/docs/current/time_zone.html.

In this issue, the problem is that ```\set``` command does not work. We should investigate problems from there.

charsyam:
@jihoonson, Thank you for your review :)

There are some problems. 
Actually, in Tajo,

1.  Client Side Var is not working with ResultSet now.
Because, getCliSideVar() and getClientSideSessionVars() see different vairable map;

2. in "SET TIME ZONE 'GMT+1' case
It is parsed and executed in Server Side. but in tajo, There is no ability to set client side var from server.
so, currently, SET TIME ZONE commands just set TIMEZONE into Server Side Variable.

That's why I just change TIMEZONE as Server Side Var.

jihoonson:
Thank you for investigating this issue. 
IMO, the problems you found are all bugs, so, we need to fix them.
Please refer to https://issues.apache.org/jira/browse/TAJO-1234 and https://issues.apache.org/jira/browse/TAJO-1241.

charsyam:
@jihoonson I changed TIMEZONE's type to DEFAULT :)

jihoonson:
Thanks @charsyam. We had a discussion about the session variable system in Tajo, and finally found that the variable type of ```TIMEZONE``` should be ```DEFAULT``` rather than ```CLI_SIDE_VAR```. 
This patch looks good to me, but I think that it would be great if we have a test code for this change.
Would you mind adding a test?

jihoonson:
Thanks @charsyam. Newly added tests look great, but it will be perfect if you add a line to verify query results which are changed according to the timezone session value.

charsyam:
@jihoonson I added new testcases for TIMEZONE in CLI :) Thanks.

jihoonson:
+1 LGTM!
I'll commit shortly.

PR #454, opened 2015-03-23
TAJO-1415: Window frame support

sirpkt:
It supports all ROWS and RANGE window frame.

Cases when window frame is not applied
- no order by clause is used
- some built-in window functions where window frame is not supported: row_number, rank, dense_rank, percent_rank, cume_dist, tile, lag, lead 

Cases when window frame should be applied
- other built-in window functions: first_value, last_value, nth_value
- normal aggregation functions

Based on above information, this patch distinguishes window function types as following three:
 1. built-in window function without window frame support
 2. built-in window function with window frame support
 3. normal aggregation functions used as a window function. In this case, window frame should be supported

And, it further distinguishes window frame types as following four:
 1. entire partition
 2. from the start of the partition to the moving end point relative to current row
 3. from the moving start point relative to current row to the end of the partition
 4. sliding frame as the current row position varies

Case 1 is the same as previous handling of window function.
Case 2 is handled as incremental termination of aggregation function, which means for every row call merge() and terminate() of the given function
Case 3 is handled almost the same as case 2 except feeding rows to the function from the end of the partition to the start of the frame, i.e., in reverse order
Case 4 is handled by two pass approach: making small loop of feeding rows to the function for each row value computation, I think, which is inevitable since aggregation function does not support sliding window aggregation.

All above are implemented for ROWS first, 
and then expanded to support RANGE by including rows that has the same order by value with current row in computation of window function.

This patch includes following changes
- parser can handle integer offset PRECEDING and FOLLOWING
- ExprAnnotator can reflect window frame information on WindowFunctionEval including default value handling
- WindowAggExec can handles ROWS and RANGE with window frame support
- Parameter checking in parser and ExprAnnotator is included
- last_value is re-implemented as WindowAggFunc. First_value implementation becomes more simple
- Window related classes in tajo-plan has new prefix 'Logical' to distinguish themselves with the same name class in tajo-algebra
- plan.proto is modified to support data structure to distinguish function types and frame types
- add test cases for window frame


sirpkt:
I checked 'mvn clean install' passed in my laptop.

jihoonson:
Thanks. I'll review today.

jihoonson:
Sorry for late review. It seems the patch have gone stale.
Would you mind rebasing it?

sirpkt:
Thank you for the review, @jihoonson 
I just rebased and resolved some conflicts from recent patches.

jihoonson:
Thanks. I'll finish my review soon.

tajo-algebra/src/main/java/org/apache/tajo/algebra/WindowSpec.java
```diff
@@ -128,12 +128,14 @@ public boolean equals(Object obj) {
   public static enum WindowFrameStartBoundType {
     UNBOUNDED_PRECEDING,
     CURRENT_ROW,
-    PRECEDING
+    PRECEDING,
+    FOLLOWING
   }
```
jihoonson: We don't need to separately maintain the start bound type and end bound type. Please see the discussion at https://github.com/apache/tajo/pull/13#issuecomment-46403452.

tajo-algebra/src/main/java/org/apache/tajo/algebra/WindowSpec.java
```diff
@@ -128,12 +128,14 @@ public boolean equals(Object obj) {
   public static enum WindowFrameStartBoundType {
```
jihoonson: Please remove some redundant static modifiers for inner enums in this class.

tajo-core/src/main/java/org/apache/tajo/engine/function/window/CurrentValue.java
```diff
@@ -24,41 +24,34 @@
...
+    return new CurrentValueContext();
   }
 
   @Override
   public void eval(FunctionContext ctx, Tuple params) {
-    FirstValueContext firstValueCtx = (FirstValueContext)ctx;
-    if(firstValueCtx.isSet == false) {
-      firstValueCtx.isSet = true;
+    CurrentValueContext currentValueCtx = (CurrentValueContext)ctx;
       if (params.get(0).isNotNull()) {
```
jihoonson: This line seems to check whether ```params.get(0)``` is the ```NullDatum``` or not. In addition to this, we should check ```params.get(0)``` is actually ```null``` or not.

tajo-core/src/main/java/org/apache/tajo/engine/parser/SQLAnalyzer.java
```diff
@@ -491,14 +506,16 @@ public WindowSpec buildWindowSpec(SQLParser.Window_specificationContext ctx) {
     WindowFrameStartBoundType boundType = null;
     if (checkIfExist(context.UNBOUNDED())) {
       boundType = WindowFrameStartBoundType.UNBOUNDED_PRECEDING;
-    } else if (checkIfExist(context.unsigned_value_specification())) {
+    } else if (checkIfExist(context.PRECEDING())) {
       boundType = WindowFrameStartBoundType.PRECEDING;
+    } else if (checkIfExist(context.FOLLOWING())) {
+      boundType = WindowFrameStartBoundType.FOLLOWING;
     } else {
       boundType = WindowFrameStartBoundType.CURRENT_ROW;
```
jihoonson: In pgsql, the default start bound is ```CURRENT_ROW```. Please see the following quote.
```
The default framing option is RANGE UNBOUNDED PRECEDING, which is the same as RANGE BETWEEN UNBOUNDED PRECEDING AND CURRENT ROW.
```

tajo-core/src/main/java/org/apache/tajo/engine/planner/physical/WindowAggExec.java
```diff
@@ -65,8 +66,8 @@
   private boolean [] orderedFuncFlags;
   private boolean [] aggFuncFlags;
   private boolean [] windowFuncFlags;
-  private boolean [] endUnboundedFollowingFlags;
-  private boolean [] endCurrentRowFlags;
+  private int [] startOffset;
+  private int [] endOffset;
```
jihoonson: It would be nice if these names are more intuitive.

tajo-core/src/main/java/org/apache/tajo/engine/planner/physical/WindowAggExec.java
```diff
@@ -301,30 +301,327 @@ private void evaluationWindowFrame() {
...
 
-      for (int i = 0; i < accumulatedInTuples.size(); i++) {
-        Tuple inTuple = accumulatedInTuples.get(i);
-        Tuple outTuple = evaluatedTuples.get(i);
+      LogicalWindowSpec.LogicalWindowFrame.WindowFrameType windowFrameType = functions[idx].getLogicalWindowFrame().getFrameType();
+      WindowSpec.WindowFrameUnit windowFrameUnit = functions[idx].getLogicalWindowFrame().getFrameUnit();
+      int frameStart = 0, frameEnd = accumulatedInTuples.size() - 1;
+      int startOffset = functions[idx].getLogicalWindowFrame().getStartBound().getNumber();
+      int endOffset = functions[idx].getLogicalWindowFrame().getEndBound().getNumber();
+      functions[idx].bind(inSchema);
```
jihoonson: bind() is to prepare the function evaluation. Thus, it should be called once before evaluation.

tajo-core/src/main/java/org/apache/tajo/engine/planner/physical/WindowAggExec.java
```diff
@@ -301,30 +301,327 @@ private void evaluationWindowFrame() {
...
+            a) Entire window partition
+            b) From the start of window partition to the moving end point relative to current row position
+            c) From the moving start point relative to current row position to the end of window partition
+            d) Both start point and end point of window frame are moving relative to the current row position
+
+         In the case of RANGE, there can be three window frame type
+            i) From the start of window partition to the last row that has the same order by key as the current row
+            ii) From the first row that has the same order by key as the current row to the end of window partition
+            iii) For all rows that has the same order by key as the current row
+       */
```
jihoonson: Thanks for detailed and nice comments!
By the way, the below code block look quite complicated. Would you refactor it into several well-defined functions?

tajo-plan/src/main/java/org/apache/tajo/plan/ExprAnnotator.java
```diff
@@ -678,8 +679,26 @@ public EvalNode visitGeneralSetFunction(Context ctx, Stack<Expr> stack, GeneralS
...
+  public static final Set<String> NONFRAMABLE_WINDOW_FUNCTIONS =
+      Sets.newHashSet("row_number", "rank", "dense_rank", "percent_rank", "cume_dist", "ntile", "lag", "lead");
+  public static final Set<String> FRAMABLE_WINDOW_FUNCTIONS =
+      Sets.newHashSet("first_value", "last_value", "nth_value");
+
+  public static FunctionType getFunctionType(String functionName, boolean distinct) {
+//    if (NONFRAMABLE_WINDOW_FUNCTIONS.contains(functionName.toLowerCase()) || FRAMABLE_WINDOW_FUNCTIONS.contains(functionName.toLowerCase())) {
+    if (WINDOW_FUNCTIONS.contains(functionName.toLowerCase())) {
+      if (distinct && functionName.equalsIgnoreCase("row_number")) {
+        throw new NoSuchFunctionException("row_number() does not support distinct keyword.");
```
jihoonson: Is  ```row_number``` the only one that cannot be used with ```distinct```?
In addition, ```UnsupportedException``` looks more proper than ```NoSuchFunctionException```.

tajo-plan/src/main/java/org/apache/tajo/plan/ExprAnnotator.java
```diff
@@ -678,8 +679,26 @@ public EvalNode visitGeneralSetFunction(Context ctx, Stack<Expr> stack, GeneralS
     }
   }
 
+
   public static final Set<String> WINDOW_FUNCTIONS =
-      Sets.newHashSet("row_number", "rank", "dense_rank", "percent_rank", "cume_dist", "first_value", "lag");
+      Sets.newHashSet("row_number", "rank", "dense_rank", "percent_rank", "cume_dist", "ntile", "first_value", "last_value", "lag");
```
jihoonson: ```lead()``` function is missed.

jihoonson:
@sirpkt, thanks for great patch!
Thanks to your nice patch overview and plentiful comments, I was able to easily review.
In overall your patch looks good to me. I left some comments. 
Thanks!

tajo-core/src/main/java/org/apache/tajo/engine/parser/SQLAnalyzer.java
jihoonson: I found another initialization code for the implicit window frame in ```ExprAnnotator::visitWindowFunction```. I wonder how this line is related to that code.

sirpkt:
Thank you for the review, @jihoonson !
I'll reflect your comments soon.

tajo-algebra/src/main/java/org/apache/tajo/algebra/WindowSpec.java
sirpkt: Hmm... I think the discussion that you linked is not complete one
because UNBOUNDED_PRECEDING is only available at start bound type while UNBOUNDED_FOLLOWING is only available as an end bound type.
I tested in psql as following
```
mydb=# select id, max(value) over (partition by id rows between unbounded following and unbounded following) from test3;
ERROR:  frame start cannot be UNBOUNDED FOLLOWING
LINE 1: ...id, max(value) over (partition by id rows between unbounded ...
                                                             ^
```
and saw the error as I expected.

So, I think it needs separation between start bound type and end bound type.

tajo-algebra/src/main/java/org/apache/tajo/algebra/WindowSpec.java
sirpkt: Thank you!
I'll remove 'static' for WindowFrameUnit, WindowFrameStartBoundType, and WindowFrameEndBoundType.

tajo-algebra/src/main/java/org/apache/tajo/algebra/WindowSpec.java
jihoonson: Right. I hope that you read a couple of comments in the thread of the link. As I commented there, there is only one rule that the end bound cannot precede the start bound. So, naturally, the start bound cannot be unbounded following. 
I think that we can make some codes simpler with the unified bound type.

tajo-core/src/main/java/org/apache/tajo/engine/function/window/CurrentValue.java
sirpkt: You're right @jihoonson.
I'll add check code for params.

tajo-algebra/src/main/java/org/apache/tajo/algebra/WindowSpec.java
sirpkt: I don't think only one rule is enough because start bound type and end bound type can be the same. 
For example, we can use ```rows between current row and current row``` as followings
```
mydb=# select id, max(value) over (partition by id rows between current row and current row) from test3;
 id | max 
----+-----
  1 |   1
  1 |   2
  2 |   5
  2 |   6
  2 |   7
(5 rows)
```
So, actually there are two rules.
1) end bound cannot precede the start bound, but can be the same as the start bound
2) UNBOUNDED PRECEDING cannot be an end bound and UNBOUNDED FOLLOWING cannot be a start bound

However, I totally agree with you that the code becomes much simpler if we use the unified bound type. So, I'll try to unify the bound type.

tajo-core/src/main/java/org/apache/tajo/engine/parser/SQLAnalyzer.java
sirpkt: Oh, I wrongly set the default value of start bound.
I think you mean that the default value of start bound should be UNBOUNDED PRECEDING, right?
I'll change the default start boundType to UNBOUNDED PRECEDING.

sirpkt: I'm a little confused about the code because it passes several weeks after writing the code.
Sorry, forget about above comments.

Code in SQLAnalyzer.java only works when the user explicitly write the window frame specification on sql string, while ExprAnnotator::visitWindowFunction fills implicit information when user does not specify it.

sirpkt: I'll add if condition for last 'else' not to be wrongly understood as default value

tajo-core/src/main/java/org/apache/tajo/engine/planner/physical/WindowAggExec.java
sirpkt: Thank you!
I'll use windowFrameStartOffset and windowFrameEndOffset.

tajo-core/src/main/java/org/apache/tajo/engine/planner/physical/WindowAggExec.java
sirpkt: Oh, I didn't check the duplicate code in init().
I'll remove the code in line 319.

tajo-core/src/main/java/org/apache/tajo/engine/planner/physical/WindowAggExec.java
sirpkt: You are right, @jihoonson.
I'll refactor the code.

tajo-plan/src/main/java/org/apache/tajo/plan/ExprAnnotator.java
sirpkt: It is because 'lead' function is implemented not as WindowAggFunc but as AggFunction.
I think it should be re-implemented as WindowAggFunc 
but it needs more consideration about handling of arguments and default output value
so that I think it is better to handle as separate issue.

tajo-plan/src/main/java/org/apache/tajo/plan/ExprAnnotator.java
sirpkt: I just made a function wrapper for the code and did not change the code itself, 
however, I think ```distinct``` should be allowed with ```row_number()``` as we can use ```distinct row_number()``` in other DBMSs.
I'll check further about that.

tajo-algebra/src/main/java/org/apache/tajo/algebra/WindowSpec.java
jihoonson: Oh, you are right. Thanks.

tajo-core/src/main/java/org/apache/tajo/engine/parser/SQLAnalyzer.java
jihoonson: Good.

tajo-plan/src/main/java/org/apache/tajo/plan/ExprAnnotator.java
jihoonson: Thanks. I agree.

tajo-plan/src/main/java/org/apache/tajo/plan/ExprAnnotator.java
jihoonson: Right. Thanks.

sirpkt:
I updated the patch and it passed 'mvn clean install' in my laptop.

    Reflecting review comments
    
    - Refactor WindowAggExec code
    - Unify window frame start and end bound type
    - Remove unnecessary 'static'
    - Add parameter test in CurrentValue
    
    NOT CHANGED YET:
    - row_number() with DISTINCT still disabled
    - LEAD() still AggFunction not WindowAggFunc

tajo-plan/src/main/java/org/apache/tajo/plan/ExprAnnotator.java
```diff
@@ -726,47 +747,125 @@ public EvalNode visitWindowFunction(Context ctx, Stack<Expr> stack, WindowFuncti
...
-        frame = new WindowFrame(new WindowStartBound(WindowFrameStartBoundType.UNBOUNDED_PRECEDING),
-            new WindowEndBound(WindowFrameEndBoundType.UNBOUNDED_FOLLOWING));
+      frame = new LogicalWindowFrame(WindowFrameUnit.RANGE, new LogicalWindowBound(WindowFrameBoundType.UNBOUNDED_PRECEDING),
+            new LogicalWindowBound(WindowFrameBoundType.CURRENT_ROW));
+    }
+
+    // set Function Type to determine whether window frame is applied or not
+    //    NONFRAMABLE: builtin window function that works on entire partition
+    //    FRAMABLE: builtin window function that works on window frame
+    //    AGGREGATION: aggregation functions that work on window frame
```
jihoonson: How about moving these comments to the code where the enum is defined?

jihoonson:
Thanks @sirpkt. In overall, the updated patch looks good to me.
I'm still reviewing this patch. I'll finish soon.

tajo-algebra/src/main/java/org/apache/tajo/algebra/WindowSpec.java
```diff
@@ -175,14 +175,14 @@ public Object clone() throws CloneNotSupportedException {
   }
 
   public static class WindowStartBound implements Cloneable {
```
jihoonson: Do we need to keep

tajo-core/src/main/java/org/apache/tajo/engine/function/window/CurrentValue.java
```diff
@@ -24,41 +24,34 @@
...
 
   @Override
   public void eval(FunctionContext ctx, Tuple params) {
-    FirstValueContext firstValueCtx = (FirstValueContext)ctx;
-    if(firstValueCtx.isSet == false) {
-      firstValueCtx.isSet = true;
-      if (params.get(0).isNotNull()) {
-        firstValueCtx.first = params.get(0);
+    CurrentValueContext currentValueCtx = (CurrentValueContext)ctx;
+      if (params.get(0) == null || params.get(0).isNotNull()) {
```
jihoonson: Would you explain on this line?
I wonder why ```currentValueCtx .current``` becomes ```null``` instead of ```NullDatum``` when ```params.get(0)``` is null.

tajo-core/src/main/java/org/apache/tajo/engine/planner/physical/WindowAggExec.java
```diff
@@ -347,6 +350,372 @@ private void finalizeWindow() {
...
+      do {
+        numberOfTuples++;
+        startOffset--;
+      } while (startOffset >= 0 && comp.compare(accumulatedInTuples.get(startOffset), inTuple) == 0);
+    }
+
+    return numberOfTuples;
+  }
+
+  private Tuple getFunctionSpecificInput(String funcName, int dataStart, int dataEnd) {
```
jihoonson: ```WindowAggFunc``` is designed to abstract all kinds of window functions, and each function inherited ```WindowAggFunc``` does some its own work. Even though this code makes the implementation of ```FirstValue``` and ```LastValue``` simpler, I think that this is opposed to our design. In addition, I think that this kind of function-specific routine makes the implementation of ```WindowAggExec``` complex. 
Do you have any reasons?

tajo-plan/src/main/java/org/apache/tajo/plan/expr/WindowFunctionEval.java
```diff
@@ -28,18 +28,28 @@
...
 import org.apache.tajo.storage.Tuple;
 import org.apache.tajo.util.TUtil;
 
 public class WindowFunctionEval extends AggregationFunctionCallEval implements Cloneable {
   @Expose private SortSpec [] sortSpecs;
-  @Expose WindowSpec.WindowFrame windowFrame;
+  @Expose LogicalWindowSpec.LogicalWindowFrame logicalWindowFrame;
+  private Tuple params;
+
+  public static enum WindowFunctionType {
```
jihoonson: ```static``` is redundant.

tajo-plan/src/main/java/org/apache/tajo/plan/expr/WindowFunctionEval.java
```diff
@@ -28,18 +28,28 @@
...
-import org.apache.tajo.plan.logical.WindowSpec;
+import org.apache.tajo.plan.logical.LogicalWindowSpec;
 import org.apache.tajo.storage.Tuple;
 import org.apache.tajo.util.TUtil;
 
 public class WindowFunctionEval extends AggregationFunctionCallEval implements Cloneable {
   @Expose private SortSpec [] sortSpecs;
-  @Expose WindowSpec.WindowFrame windowFrame;
+  @Expose LogicalWindowSpec.LogicalWindowFrame logicalWindowFrame;
+  private Tuple params;
```
jihoonson: Seems to be unused.

tajo-core/src/test/java/org/apache/tajo/engine/query/TestWindowQuery.java
```diff
@@ -462,4 +462,148 @@ public final void testMultipleWindow() throws Exception {
       executeString("DROP TABLE multiwindow PURGE");
     }
   }
+
+  @Test
+  public final void testWindowFrame1() throws Exception {
```
jihoonson: The tests ```testWindowFrame1```, ```testWindowFrame2```, ```testWindowFrame3```, and ```testWindowFrame4``` look good, but it would be better if they are written in our test writing convention. That is separating java codes, test queries, and results.

tajo-core/src/main/java/org/apache/tajo/engine/planner/physical/WindowAggExec.java
```diff
@@ -347,6 +350,372 @@ private void finalizeWindow() {
...
+            }
+            sameEndRange --;
+
+            Datum result = functions[functionIndex].terminate(contexts[functionIndex]);
+            Tuple outTuple = evaluatedTuples.get(i);
+            outTuple.put(nonFunctionColumnNum + functionIndex, result);
+          }
+        } else {
+          if (windowFrameEndOffset > 0) {
+            int i =0; int j = 0;
```
jihoonson: Would you mind explaning on this block? 
In addition, it would be better to rename ```i```, and ```j``` to be more intuitive if possible.

tajo-core/src/main/java/org/apache/tajo/engine/planner/physical/WindowAggExec.java
```diff
@@ -347,6 +350,372 @@ private void finalizeWindow() {
...
+    } else if (funcName.equals("last_value")) {
+      // check the frame end is within the partition
+      if (dataStart <= dataEnd && dataEnd >= 0 && dataEnd < accumulatedInTuples.size()) {
+        return accumulatedInTuples.get(dataEnd);
+      }
+    }
+    return null;
+  }
+
+  private void evaluateAggregationFunction(int functionIndex, LogicalWindowSpec.LogicalWindowFrame windowFrame,
```
jihoonson: This function looks still complicated. Would you mind dividing it into several small functions?

jihoonson:
@sirpkt thanks for your nice work. Since this is a new feature which will be very useful and popularly used, I've carefully reviewed. Please think about my comments. Thanks.

tajo-core/src/main/java/org/apache/tajo/engine/planner/physical/WindowAggExec.java
jihoonson: I don't want to push you. You can reserve this refactoring work as another jira if you want.

tajo-plan/src/main/java/org/apache/tajo/plan/ExprAnnotator.java
sirpkt: It looks better!
I moved the comments about Function Type and Frame Type to the code where the enum is defined.

tajo-algebra/src/main/java/org/apache/tajo/algebra/WindowSpec.java
sirpkt: You're right, @jihoonson.
We don't need to keep them separately.
I'll merge them into ```WindowBound```.

tajo-core/src/main/java/org/apache/tajo/engine/function/window/CurrentValue.java
sirpkt: Oh, this is a definitely bug.
Thank you for finding.
I'll fix the if condition appropriately.

tajo-core/src/main/java/org/apache/tajo/engine/planner/physical/WindowAggExec.java
sirpkt: I have several reasons why I separately handle the built-in window functions with window frame support.
At first, there are only three built-in window functions that support frame, which are first_value, last_value, and nth_value. As the number of functions are fixed and small, I think code generality is less important than simple code.
Second, those functions commonly work on the position of window frame (start, end, or n-th row from start) so that above code does make much simpler to implement WindowAggExec, which now focuses on supporting window frame for aggregation functions only.

I agree that you have a point there.
If someone adds more built-in window functions with frame support, 
she or he may feel confused from different manner of implementation.
However, as I know, there are only one unimplemented function, nth_value, which can be easily added like first_value or last_value.

tajo-plan/src/main/java/org/apache/tajo/plan/expr/WindowFunctionEval.java
sirpkt: You're right.
I'll remove the ```static```.

tajo-plan/src/main/java/org/apache/tajo/plan/expr/WindowFunctionEval.java
sirpkt: Yes, you're right.
I'll remove the ```params```.

tajo-core/src/test/java/org/apache/tajo/engine/query/TestWindowQuery.java
sirpkt: You're right, @jihoonson.
I'll separate codes, queries, and results.

By the way, in other part of Tajo code, 
there exists test codes like what I did (for example, testJoinQuery, testJoinBroadcast, ...).
I think we need overall refactoring of test code to make them follow our test writing convention.

tajo-core/src/main/java/org/apache/tajo/engine/planner/physical/WindowAggExec.java
jihoonson: Ok. I got your point.
If so, would you please add some comments for the special functions (first_value, last_value, nth_value) to reduce confusion?

sirpkt:
I reflected considerate and detailed comments from @jihoonson.

tajo-catalog/tajo-catalog-drivers/pom.xml
```diff
@@ -41,6 +41,7 @@
         <configuration>
           <excludes>
             <exclude>tajo-hive/**</exclude>
+            <exclude>tajo-hcatalog/**</exclude>
```
jihoonson: This looks to be already removed in https://issues.apache.org/jira/browse/TAJO-1442.

jihoonson:
@sirpkt thanks for updating your patch. 
I've tested the following query and found that the result of Tajo is different from that of pgsql.

**Tajo**
```
default> select l_partkey, first_value(l_linestatus) over (partition by l_partkey order by l_tax rows between 5 preceding and 8 following) from lineitem;
...
l_partkey,  ?windowfunction
1,  
1,  
1,  
1,  
1,  
1,  F
1,  O
1,  O
1,  F
1,  F
1,  O
1,  F
1,  F
1,  F
1,  F
1,  O
1,  O
1,  F
1,  F
1,  F
1,  O
1,  O
1,  F
1,  F
1,  O
1,  F
1,  F
1,  F
1,  O
1,  F
1,  O
2,  
2,  
2,  
2,  
2,  
2,  O
...
```
**PostgreSQL**
```
jihoonson=# select l_partkey, first_value(l_linestatus) over (partition by l_partkey order by l_tax rows between 5 preceding and 8 following) from lineitem;
 l_partkey | first_value 
-----------+-------------
         1 | O
         1 | O
         1 | O
         1 | O
         1 | O
         1 | O
         1 | F
         1 | O
         1 | F
         1 | F
         1 | F
         1 | F
         1 | O
         1 | F
         1 | F
         1 | O
         1 | O
         1 | F
         1 | F
         1 | F
         1 | F
         1 | F
         1 | O
         1 | O
         1 | O
         1 | F
         1 | O
         1 | F
         1 | F
         1 | F
         1 | O
         2 | O
...
```
As you can see, some values are null in Tajo.

sirpkt:
Thank you for the finding, @jihoonson.
Difference between result of the patch and the result of postgresql is from the bug.
I'll fix the bug and add more test codes about window frame.

tajo-catalog/tajo-catalog-drivers/pom.xml
sirpkt: You're right, @jihoonson.
I'll remove the line.

sirpkt:
I rebased and updated the patch.

    - bug fix for the handling of built-in window functions with frame support
    - add more test codes for the preceding and following with constant cases

jihoonson:
Thanks @sirpkt, but query results are still different.
Would you check it again?

**Tajo**
```
default> select l_partkey, l_tax, first_value(l_linestatus) over (partition by l_partkey order by l_tax rows between 5 preceding and 8 following) from lineitem;
l_partkey,  l_tax,  ?windowfunction
1,  0.0,  F
1,  0.0,  F
1,  0.01,  F
1,  0.01,  F
1,  0.01,  F
1,  0.02,  F
1,  0.02,  O
1,  0.02,  F
1,  0.03,  O
1,  0.03,  F
1,  0.04,  F
1,  0.04,  O
1,  0.05,  F
1,  0.05,  F
1,  0.06,  F
1,  0.06,  O
1,  0.06,  O
1,  0.06,  F
1,  0.06,  F
1,  0.07,  F
1,  0.07,  O
1,  0.07,  F
1,  0.07,  F
1,  0.07,  O
1,  0.07,  O
1,  0.08,  F
1,  0.08,  O
1,  0.08,  F
1,  0.08,  F
1,  0.08,  F
1,  0.08,  F
2,  0.0,  O
2,  0.0,  O
2,  0.0,  O
2,  0.01,  O
2,  0.01,  O
...
```
**PostgreSQL**
```
jihoonson=# select l_partkey, l_tax, first_value(l_linestatus) over (partition by l_partkey order by l_tax rows between 5 preceding and 8 following) from lineitem;
 l_partkey | l_tax | first_value 
-----------+-------+-------------
         1 |     0 | O
         1 |     0 | O
         1 |  0.01 | O
         1 |  0.01 | O
         1 |  0.01 | O
         1 |  0.02 | O
         1 |  0.02 | F
         1 |  0.02 | O
         1 |  0.03 | F
         1 |  0.03 | F
         1 |  0.04 | F
         1 |  0.04 | F
         1 |  0.05 | O
         1 |  0.05 | F
         1 |  0.06 | F
         1 |  0.06 | O
         1 |  0.06 | O
         1 |  0.06 | F
         1 |  0.06 | F
         1 |  0.07 | F
         1 |  0.07 | F
         1 |  0.07 | F
         1 |  0.07 | O
         1 |  0.07 | O
         1 |  0.07 | O
         1 |  0.08 | F
         1 |  0.08 | O
         1 |  0.08 | F
         1 |  0.08 | F
         1 |  0.08 | F
         1 |  0.08 | O
         2 |     0 | O
         2 |     0 | O
         2 |     0 | O
         2 |  0.01 | O
         2 |  0.01 | O
         2 |  0.01 | O
         2 |  0.02 | O
         2 |  0.02 | O
         2 |  0.02 | O
...
```

PR #455, opened 2015-03-23
TAJO-1410: TPC-DS TestCase

babokim:
The TPC-DS data set which is generated scale factor 1 is almost 1.2GB. It is too big to run test case on the single machine. So TPC-DS testcase is separated from main test case. In this patch, I added 'tpcds-test' profile to run TPC-DS test case. TPC-DS test case doesn't run with default build option. 
You can choose pre-generated test data set with '-Dtpcds.gen.data=false' and '-Dtpcds.data.dir=<TPC-DS DATA DIRECTORY>' option.

Use -Dtpcds-test to run TPC-DS test case with teh following additional options:
    -Dtpcds.gen.data=<true or false> : required
    -Dtpcds.data.dir=<TPC-DS DATA DIRECTORY>: required
    -Dtpcds.generator=<TPC-DS DATA GENERATOR>: if tpcds.gen.data=true, required

I used the following command on my laptop.
mvn test -Ptpcds-test -Dtpcds.gen.data=yes -Dtpcds.data.dir=/Users/babokim/tmp/tpcds1 -Dtpcds.generator=/Users/babokim/Downloads/tpcds-kit-master/tools/dsdgen


babokim:
After discussion, this patch will be hosted in the following repository : 
https://github.com/babokim/tajo-tcpds

PR #456, opened 2015-03-23
TAJO-1436 Add Bind method to EvalNode

navis:
eval(Schema schema, Tuple tuple) --> bind(Schema schema) + eval(Tuple tuple)

### DIFF
--- a/tajo-core/src/main/java/org/apache/tajo/engine/planner/physical/HashLeftOuterJoinExec.java
+++ b/tajo-core/src/main/java/org/apache/tajo/engine/planner/physical/HashLeftOuterJoinExec.java
@@ -120,6 +120,11 @@ public class HashLeftOuterJoinExec extends BinaryPhysicalExec {
     leftKeyTuple = new VTuple(leftKeyList.length);
 
     rightNumCols = rightChild.getSchema().size();
+
+    joinQual.bind(inSchema);
+    if (joinFilter != null) {
+      joinFilter.bind(inSchema);
+    }
   }
 
   @Override
@@ -177,8 +182,8 @@ public class HashLeftOuterJoinExec extends BinaryPhysicalExec {
       frameTuple.set(leftTuple, rightTuple); // evaluate a join condition on both tuples
 
       // if there is no join filter, it is always true.
-      boolean satisfiedWithFilter = joinFilter == null ? true : joinFilter.eval(inSchema, frameTuple).isTrue();
-      boolean satisfiedWithJoinCondition = joinQual.eval(inSchema, frameTuple).isTrue();
+      boolean satisfiedWithFilter = joinFilter == null || joinFilter.eval(frameTuple).isTrue();
+      boolean satisfiedWithJoinCondition = joinQual.eval(frameTuple).isTrue();
 
       // if a composited tuple satisfies with both join filter and join condition
       if (satisfiedWithJoinCondition && satisfiedWithFilter) {
--- a/tajo-core/src/main/java/org/apache/tajo/engine/planner/physical/SeqScanExec.java
+++ b/tajo-core/src/main/java/org/apache/tajo/engine/planner/physical/SeqScanExec.java
@@ -27,6 +27,7 @@ import org.apache.tajo.catalog.partition.PartitionMethodDesc;
 import org.apache.tajo.catalog.proto.CatalogProtos;
 import org.apache.tajo.catalog.statistics.TableStats;
 import org.apache.tajo.datum.Datum;
+import org.apache.tajo.datum.NullDatum;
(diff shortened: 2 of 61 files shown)


tajo-core/src/main/java/org/apache/tajo/engine/planner/physical/HashLeftOuterJoinExec.java
```diff
@@ -120,11 +120,11 @@ public HashLeftOuterJoinExec(TaskAttemptContext context, JoinNode plan, Physical
     leftKeyTuple = new VTuple(leftKeyList.length);
 
     rightNumCols = rightChild.getSchema().size();
-  }
 
-  @Override
-  protected void compile() {
-    joinQual = context.getPrecompiledEval(inSchema, joinQual);
```
jihoonson: The compile() function should not be removed.

tajo-core/src/main/java/org/apache/tajo/engine/planner/physical/SeqScanExec.java
```diff
@@ -69,9 +70,8 @@ public SeqScanExec(TaskAttemptContext context, ScanNode plan,
     this.qual = plan.getQual();
     this.fragments = fragments;
 
-    if (fragments != null
-        && plan.getTableDesc().hasPartition()
```
jihoonson: The variable ```fragments``` can be null, and ```rewriteColumnPartitionedTableSchema()``` should not be called in that case.

tajo-core/src/main/java/org/apache/tajo/engine/planner/physical/HashLeftOuterJoinExec.java
navis: Yes, mistake. Nice catch.

tajo-core/src/main/java/org/apache/tajo/engine/planner/physical/SeqScanExec.java
```diff
@@ -197,6 +206,9 @@ public Tuple next() throws IOException {
     if (fragments == null) {
       return null;
     }
+    if (scanner == null) {
+      System.out.println("[SeqScanExec/next] ");
```
jihoonson: Please remove debug codes.

tajo-core/src/main/java/org/apache/tajo/engine/planner/physical/SeqScanExec.java
navis: Even fragments is null, partition columns in expr should be replaced with null. Without this, some tests fails with exception("No Such Column Reference").

tajo-plan/src/main/java/org/apache/tajo/plan/expr/FunctionEval.java
```diff
@@ -40,13 +40,27 @@
   @Expose protected FunctionDesc funcDesc;
 	@Expose protected EvalNode [] argEvals;
 
+  private transient Tuple params;
```
jihoonson: This variable is missed in ```hashCode()```, ```clone()```, and ```equals()``` functions.

navis: It's transient field as it declared, which should not be included in hash or equals calculation.

tajo-core/src/main/java/org/apache/tajo/engine/planner/physical/SeqScanExec.java
navis: done

tajo-plan/src/main/java/org/apache/tajo/plan/expr/FunctionEval.java
jihoonson: Oops. Sorry.

jihoonson:
Thanks @navis. The patch looks good.
I have some additional comments as follows.
* Changes on NettyClientBase look for https://issues.apache.org/jira/browse/TAJO-1424. If so, it would be better to resolve in that issue. Please remove them.
* The bind() function must be called before eval(). So, it would be much great if you add a routine to check the binding state.

navis:
I've also thought about that (bind check), but thinking that the intention of this issue was for removing any non-working code from eval, I decided to remove it. Opinion?

I'll fix TAJO-1424, which seemed abondoned for a long time.

jihoonson:
Ok. I created another issue for bind check. https://issues.apache.org/jira/browse/TAJO-1499

jihoonson:
Would you please rebase?

jihoonson:
Oh, you have already rebased. Sorry. I'll review ASAP.

PR #458, opened 2015-03-24
TAJO-923: Add VAR_SAMP and VAR_POP window functions

dongjoon-hyun:
I coded in the same way in https://github.com/apache/tajo/pull/427 (TAJO-921)
If needed, after merging pull-request 427, I will do refactoring.

This code is passed the following command.
```
mvn clean install -Pparallel-test,hcatalog-0.12.0 -DLOG_LEVEL=ERROR -Dmaven.fork.count=2 &> TAJO-923.travis.ERROR.txt
```
The result file is https://app.box.com/s/clnx9d2bisfhbh2upgo3q6g7bdx7aic5

### DIFF
--- a/tajo-core/src/main/java/org/apache/tajo/engine/function/builtin/VarPop.java
+++ b/tajo-core/src/main/java/org/apache/tajo/engine/function/builtin/VarPop.java
@@ -16,23 +16,27 @@
  * limitations under the License.
  */
 
-option java_package = "org.apache.tajo";
-option java_outer_classname = "InternalTypes";
-option java_generic_services = false;
-option java_generate_equals_and_hash = true;
+package org.apache.tajo.engine.function.builtin;
 
-message AvgLongProto {
-  required int64 sum = 1;
-  required int64 count = 2;
-}
+import org.apache.tajo.catalog.Column;
+import org.apache.tajo.datum.Datum;
+import org.apache.tajo.datum.DatumFactory;
+import org.apache.tajo.datum.NullDatum;
+import org.apache.tajo.plan.function.FunctionContext;
 
-message AvgDoubleProto {
-  required double sum = 1;
-  required int64 count = 2;
-}
+public abstract class VarPop extends Variance {
+  public VarPop(Column[] definedArgs) {
+    super(definedArgs);
+  }
 
-message VarianceProto {
-  required double squareSumOfDiff = 1;
-  required double avg = 2;
-  required int64 count = 3;
+  @Override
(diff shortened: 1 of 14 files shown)


dongjoon-hyun:
Rebased.

dongjoon-hyun:
Rebased.

jihoonson:
I'll review today.

jihoonson:
The only difference between this patch and TAJO-921 is the absence of square root. Why should we maintan mostly duplicated codes separately?

jihoonson:
No offense. Sorry for late review.

dongjoon-hyun:
@jihoonson , I knew that 20 days ago. :) But, there was no way to implement variance at that time, since TAJO-921 is not committed yet. As I mentioned my first comment, we need to change. 

Here, I need your guide how to reduce the redundancy. Is it proper to compute STD_DEV first and square that?

jihoonson:
Thanks for your rapid reply. 
Yes, I also think that will be a good way.

dongjoon-hyun:
Hi, @jihoonson . I removed the following redundancy. Now, VarSamp and VarPop use StdDevContext.
* ProtocolBuffer message for VarSamp/VarPop
* Var.java file

tajo-core/src/main/java/org/apache/tajo/engine/function/builtin/VarPop.java
```diff
@@ -0,0 +1,42 @@
...
+
+package org.apache.tajo.engine.function.builtin;
+
+import org.apache.tajo.catalog.Column;
+import org.apache.tajo.datum.Datum;
+import org.apache.tajo.datum.DatumFactory;
+import org.apache.tajo.datum.NullDatum;
+import org.apache.tajo.plan.function.FunctionContext;
+
+public abstract class VarPop extends StdDev {
```
jihoonson: I think that we need to change ```StdDev``` to more general name.
How about ```Variance```? Or, do you have any good idea?

dongjoon-hyun: It sounds good to me, too. I agree that `Variance` looks more general to cover all VarPop/VarSamp/StdDevPop/StdDevSamp functions. Then, `StdDevContext` should be changed into `VarianceContext` accordingly.

dongjoon-hyun: @jihoonson , I changed the code according to your comments. Thank you always.

jihoonson:
+1 thanks for your work.

PR #461, opened 2015-03-24
TAJO-1147: Simple query doesn't work in Web UI

eminency:
'resultRows' can be -1.
-1 means TajoConstants.UNKNOWN_ROW_NUMBER.


blrunner:
Thanks for your contribution.
I'll review the patch tomorrow. :)

blrunner:
+1

It looks good to me and it runs successfully as expected.

PR #464, opened 2015-03-26
TAJO-1451: implement elasticsearch storage for tajo.

HowookJeong:
I'm sorry.
I retried my pull request and then please review this request.


blrunner:
Thanks for your contribution and I suggest you to consider following opinions.

First, I found that you had wrote your name and current date. But we don't write our name and some date at all class files. Thus, you need to remove your name and date.
Second, you need to write description for new added class files.

For the reference, I just reviewed the patch simply. I'll review it continuously without above comments.

blrunner:
Could you trigger this for Travis CI build?

PR #468, opened 2015-03-27
TAJO-1457 Expose task retry count as session var

navis:
simple patch


navis:
Fixed build failure

jihoonson:
I think we don't need to restrict the maximum number of retrying. How about setting the minimum number?

PR #473, opened 2015-03-27
TAJO-1344: Python UDF support

jihoonson:
To support Python's dynamic typing, I added AnyDatum which can contain a datum of any types.


jihoonson:
I've added comments and documentations.
Please review the patch.

jinossy:
Great!  Really useful feature
I'll review this weekend.

jihoonson:
I'll investigate the test failure.

jihoonson:
The test failure looks to be related with the version of Jython.
I fixed it and verified with ```mvn clean install -Pparallel-test,hcatalog-0.12.0 -Dmaven.fork.count=2```.
In addition, I fixed findbugs warnings.

hyunsik:
The patch looks good to me.

I have one suggestion. You make new EvalNode Type ```PYTHON_FUNCTION```. But, it may be not matched to the design of EvalNode, and it will require duplicated codes for EvalNode visitors.

You seem to do because Python function involcation is different from existing Java function. I think that the best solution is to separate EvalNode and Function Invocation code. In more detail, you should implement some function invocation interface, and separate java function invocation and python function invocation.  Each function invocation implementation can be initialized at the first function call, or you can use EvalNode::Bind(), being addressed in TAJO-1436.

jihoonson:
@hyunsik, thanks for the nice suggestion. It is definitely a better solution.
However, I have a different opinion on detailed implementation. 
The first option that initialize each function invocation implementation at the first function call is deprecated, so it is not a good choice.
And the second option using EvalNode::Bind() seems to initialize schema-dependent variables. Even though we still can change the specification of TAJO-1436, the current one looks good. Since relations have different schemas, schema-dependent variables should be initialized for each schema. 

However, function invocation implementation does not depend on relations' schema. I think that it would be better to initialize when an instance of ```FunctionEval``` is created.
What do you think?

jihoonson:
I'll open another PR.

PR #474, opened 2015-03-27
TAJO-1460 Apply TAJO-1407 to ExternalSortExec

navis:
Applied.


tajo-core/src/main/java/org/apache/tajo/engine/planner/physical/ExternalSortExec.java
```diff
@@ -168,7 +168,7 @@ private Path sortAndStoreChunk(int chunkId, List<Tuple> tupleBlock)
     int rowNum = tupleBlock.size();
 
     long sortStart = System.currentTimeMillis();
-    Collections.sort(tupleBlock, getComparator());
+    Iterable<Tuple> sorted = getSorter(tupleBlock).sort();
```
babokim: appender should use this "sorted" instead of tupleBlock in the following for statement.
```
for (Tuple t : sorted) {  // tupleBlock -> sorted
      appender.addTuple(t);
}
```
I was wonder how the code passed TestExternalSortExec. I found two reasons.
First, the above logic is not covered in current test condition. "numTuple" in TestExternalSortExec code should be changed to "3000000" to cover more use case.
Second, ProjectionExec is parent node in the test program.
ProjectionExec reuses the tuple object when next() is called. So TestExternalSortExec code should be changed as the following:
```
while ((tuple = exec.next()) != null) {
  ...
  //preVal = curVal;
  preVal = new VTuple(curVal);
  cnt++;
}
```

tajo-core/src/main/java/org/apache/tajo/engine/planner/physical/ExternalSortExec.java
```diff
@@ -630,19 +636,43 @@ public TableStats getInputStats() {
     CLOSED
   }
 
+  private static class SimpleMerger extends PairWiseMerger {
```
babokim: I think SimpleMerger is not proper name for this class. First time when I saw the next code, I can understand after checking the detail code of SimpleMerger class. How about "VectorComparePairWiseMerger".
'''
if (ComparableVector.isApplicable(sortSpecs)) {
  return new SimpleMerger(inSchema, left, right, comparator);
}
return new PairWiseMerger(inSchema, left, right, comparator);
'''

babokim:
@navis thanks for your contribution. I left some comments.

tajo-core/src/main/java/org/apache/tajo/engine/planner/physical/ExternalSortExec.java
navis: Fixed that (some undoing seemed caused that) and changed the test.

tajo-core/src/main/java/org/apache/tajo/engine/planner/physical/ExternalSortExec.java
navis: I'm thinking of rewriting current up-heap to down-heap to reduce comparing further (50% for average AFAIK). So I've named it randomly. Till then let's name it VectorComparePairWiseMerger.

babokim:
+1
Thanks, test case is modified and all tests passed. I'll commit soon.

jihoonson:
Please hold a second. 
I wonder why the parameterized test in TestJoinQuery is disabled.

jihoonson:
In addition, I recommend to add an order by clause to queries if necessary rather than just changing the order of rows of query results.

tajo-core/src/test/java/org/apache/tajo/engine/query/TestJoinQuery.java
```diff
@@ -89,9 +89,9 @@ public TestJoinQuery(String joinOption) {
   @Parameters
   public static Collection<Object[]> generateParameters() {
     return Arrays.asList(new Object[][]{
-        {"Hash_NoBroadcast"},
```
babokim: These parameterized test cases should be included.
Each query with different parameter returns different order of rows. Adding additional order by column or order by clause can resolve this problem as the following example:
testComplexJoinCondition7.sql
```
select
  n1.n_nationkey,
  n1.n_name,
  n2.n_name
from nation n1 join (select * from nation union select * from nation) n2 on substr(n1.n_name, 1, 4) = substr(n2.n_name, 1, 4)
order by n1.n_nationkey;
``` 

You can add "n2.n_name" column in where clause.

```
select
  n1.n_nationkey,
  n1.n_name,
  n2.n_name
from nation n1 join (select * from nation union select * from nation) n2 on substr(n1.n_name, 1, 4) = substr(n2.n_name, 1, 4)
order by n1.n_nationkey, n2.n_name;
```

tajo-core/src/test/resources/results/TestJoinBroadcast/testCrossJoinWithAsterisk1.result
```diff
@@ -1,8 +1,8 @@
 r_regionkey,r_name,r_comment,c_custkey,c_name,c_address,c_nationkey,c_phone,c_acctbal,c_mktsegment,c_comment
```
babokim: @jihoonson recommend to add an order by clause to queries if necessary rather than just changing the order of rows of query results. I agree with this opinion. 
For example, you can add 'c_custkey' column in order by clause in "testCrossJoinWithAsterisk1.sql".

babokim:
@jihoonson thank your for your opinion and comment. 
@navis I left another comments to reflect @jihoonson comments.

jihoonson:
Thank you for understanding. 
I missed that MemSortExec is not used anymore when reviewing TAJO-1407. Thanks for letting @navis know that.

navis:
There was a test option in hive for ordering output of test result because order-by takes more time and sometimes tricky in complex queries. How about to introduce something like that?

jihoonson:
Good idea. But, it would be better to be considered as a separate issue. Before introducing that new feature, I think it is ineviable to use order-by clauses for stable tests.

navis:
Addressed comments

jihoonson:
+1 LGTM.

babokim:
Here is my +1.
I'll commit soon.

PR #475, opened 2015-03-28
TAJO-1462: Replace CSV examples into TEXT examples in docs

dongjoon-hyun:


jihoonson:
Thanks for your contribution. I have two comments. 
- If any options are given, the delimiter of text files is ','. I think that we need at least one example without the delimiter option. 
- You modified the document for the csv file. However, that document still looks to explain how to use csv files. So, it would be better to change the document title together.

dongjoon-hyun:
@jihoonson . Thank you! First, I will omit " text.delimiter=',' " and make some notes for the default value. Second, I will replace csv.rst into text.rst. Do you think that is enough?

jihoonson:
Right. It will be nice.

dongjoon-hyun:
Hi, @jihoonson . I rebased and changed. But I need your guide for one thing. In the document, I found that Tajo defines CSV files as 'character-separated values' files, not 'comma-separated values' file. In that case, I think CSV seems to be a more specific and exact term than TEXT. As a result, `csv.rst` looks good to me. How do you think about this?  (By the way, I found that the default value is explained enough in the doc.)

jihoonson:
In that case, users may be confused because we also provide the ```csv``` file format. 
I think that it will be enough if we introduce the ```text``` file as the delimiter-separated values format.

dongjoon-hyun:
I see, @jihoonson . You're right. I renamed and changed more description according to your comments.

tajo-docs/src/main/sphinx/table_management/text.rst
```diff
@@ -1,21 +1,21 @@
 *************************************
-CSV (TextFile)
+TEXT
 *************************************
 
-A character-separated values (CSV) file represents a tabular data set consisting of rows and columns.
+A character-separated values text file represents a tabular data set consisting of rows and columns.
```
jihoonson: It would be better if readers can easily understand what text means. As you know, ```text``` can represent both the TEXT file and strings.
How about emphasize the letters ```text``` when it means the TEXT file? For sphyinx syntax, please refer to http://sphinx-doc.org/rest.html.

tajo-docs/src/main/sphinx/table_management/text.rst
```diff
@@ -1,21 +1,21 @@
...
 =========================================
-How to Create a CSV Table ?
+How to Create a TEXT Table ?
 =========================================
 
 If you are not familiar with the ``CREATE TABLE`` statement, please refer to the Data Definition Language :doc:`/sql_language/ddl`.
 
 In order to specify a certain file format for your table, you need to use the ``USING`` clause in your ``CREATE TABLE``
-statement. The below is an example statement for creating a table using CSV files.
+statement. The below is an example statement for creating a table using text files.
```
jihoonson: Same comment with the above one.

tajo-docs/src/main/sphinx/table_management/text.rst
dongjoon-hyun: Sure. That would be better.

dongjoon-hyun:
Rebased.

tajo-docs/src/main/sphinx/table_management/text.rst
```diff
@@ -34,19 +34,19 @@ Physical Properties
...
+*TEXT* format provides the following physical properties.
 
 * ``text.delimiter``: delimiter character. ``|`` or ``\u0001`` is usually used, and the default field delimiter is ``|``.
-* ``text.null``: NULL character. The default NULL character is an empty string ``''``. Hive's default NULL character is ``'\\N'``.
+* ``text.null``: ``NULL`` character. The default ``NULL`` character is an empty string ``''``. Hive's default ``NULL`` character is ``'\\N'``.
 * ``compression.codec``: Compression codec. You can enable compression feature and set specified compression algorithm. The compression algorithm used to compress files. The compression codec name should be the fully qualified class name inherited from `org.apache.hadoop.io.compress.CompressionCodec <https://hadoop.apache.org/docs/current/api/org/apache/hadoop/io/compress/CompressionCodec.html>`_. By default, compression is disabled.
-* ``csvfile.serde`` (deprecated): custom (De)serializer class. ``org.apache.tajo.storage.TextSerializerDeserializer`` is the default (De)serializer class.
+* ``text.serde``: custom (De)serializer class. ``org.apache.tajo.storage.TextSerializerDeserializer`` is the default (De)serializer class.
```
jihoonson: The default SerDe class is ```org.apache.tajo.storage.text.CSVLineSerDe```.

jihoonson:
+1. The patch LGTM in overall.
I left one comment on the wrong name of the default SerDe class.
I'll fix it before commit.

dongjoon-hyun:
Thank you, @jihoonson !

tajo-docs/src/main/sphinx/table_management/text.rst
dongjoon-hyun: Of course! Thank you again, @jihoonson !

PR #476, opened 2015-03-29
TAJO-1463: Add ORCFile store type to create ORCFile table

dongjoon-hyun:


dongjoon-hyun:
Rebased and added error handling cases according to the discussion https://issues.apache.org/jira/browse/TAJO-29.
When selecting non-empty ORCFile table, return IOException instead of NPE. I think this is more general error handling way rather than throwing NotImplementedException for a new specific storage type.
By the way, I welcome any feedback for the way of error handling.

tajo-storage/tajo-storage-common/src/main/java/org/apache/tajo/storage/StorageManager.java
```diff
@@ -538,7 +538,8 @@ public Appender getAppender(OverridableConf queryContext,
     if (scannerClass == null) {
       scannerClass = conf.getClass(
           String.format("tajo.storage.scanner-handler.%s.class", handlerName), null, Scanner.class);
-      SCANNER_HANDLER_CACHE.put(handlerName, scannerClass);
+      if (scannerClass != null)
```
blrunner: Could you explain why you add IF statement?

dongjoon-hyun: Hi, @blrunner .

I added that because SCANNER_HANDLER_CACHE raises NPE when `getClass` function fails. Even in case that Tajo cannot find a scanner-handler class, I thought it is safe that Tajo shows a reasonable message instead of NPE like the following. 
```
default> select * from orc1; 
ERROR: Unknown Storage Type: ORCFILE
```
This sub-task issue provides ORCFile table creation related patch. This patch supports creating and reading the created table (of course, empty). Until the very next issue supports reading non-empty ORCFile table, 'IF' will show the above message instead of NPE.

Otherwise, it is also possible to throw NonImplementedExcpetion for ORCFile table.

@blrunner , could you give me some advice for more preferred way in Tajo?

jihoonson:
IOException means that some errors occur during I/O. How about UnsupportedException?

tajo-storage/tajo-storage-common/src/main/java/org/apache/tajo/storage/StorageManager.java
blrunner: Hi @dongjoon-hyun 

Thank you for your detailed comments.
I just leaved the comment because I couldn't create new table without reference to the IF statement as following:

```
default> create table test1 (id int) using orc;
ERROR: Store format orc is not supported.
```

Anyway, I can't reproduce the NPE now because of above situation. I think that it may be related to other source codes. If possible, could you check it again?

blrunner: Sorry @dongjoon-hyun 

I wrote a typo for file format.
I'll test it again. :)

dongjoon-hyun:
@jihoonson No problem. I will change into UnsupportedException.

tajo-storage/tajo-storage-common/src/main/java/org/apache/tajo/storage/StorageManager.java
dongjoon-hyun: @blrunner Thank you for your review! I hope everything goes well. Please comment me anything to be changed.

blrunner:
It looks good Jihoon's opinion. And there are another way using PreLogicalPlanVerifier or LogicalPlanVerifier.

But I would like you to proceed the umbrella issue in following order.
1. Implement ORCAppender and ORCScanner
2. Implement ORCStorageManager
3. Added ORC type to Tajo grammer.

I suggested above idea because we need to proceed development avoiding users confusion. Also if you want the umbrella issue to merge to the apache master branch, you need to complete all sub tasks before the next release. For reference, I don't intend to push you and I just mean to help your schedule.

dongjoon-hyun:
@blrunner Thank you for your kind guidance again. I see. I will proceed the umbrella issue in that way from now. Anyway, I rebased and changed this issue according to @jihoonson 's comment before forgetting that. :)

dongjoon-hyun:
Rebased.

PR #477, opened 2015-03-30
TAJO-1466 Fix problems in table cache

navis:
Preliminary patch


jinossy:
Can you share an error log?

PR #479, opened 2015-03-30
TAJO-1467: Parenthesis at the start of SQL query is ignored

sirpkt:
Fix SimpleParser to accept a SQL query starts with '(' character.
Also fix SQLAnalyzer to handle non_join_query_primary properly as defined in Tajo grammar.


babokim:
This patch looks good to me. I had same problem to run TPC-DS query66. After applying this patch that query runs well and all test cases passed.
Here is my +1. I'll commit this patch soon.

babokim:
@sirpkt, I misunderstood a contributor worked this patch. I am sorry about that.

sirpkt:
It's ok, @babokim.
Thank you for the review!
I'll commit the patch soon.

sirpkt:
I just committed the patch.

PR #484, opened 2015-03-31
TAJO-1482: Cleanup the legacy cluster mode.

jinossy:


jihoonson:
+1
This is definitely necessary work. Thanks!

PR #485, opened 2015-03-31
TAJO-1484 Apply on ColPartitionStoreExec

navis:
Based on TAJO-1460

### DIFF
--- a/tajo-core/src/main/java/org/apache/tajo/engine/planner/physical/SortBasedColPartitionStoreExec.java
+++ b/tajo-core/src/main/java/org/apache/tajo/engine/planner/physical/SortBasedColPartitionStoreExec.java
@@ -23,9 +23,9 @@ package org.apache.tajo.engine.planner.physical;
 
 import org.apache.tajo.catalog.statistics.StatisticsUtil;
 import org.apache.tajo.datum.Datum;
+import org.apache.tajo.engine.planner.physical.ComparableVector.ComparableTuple;
 import org.apache.tajo.plan.logical.StoreTableNode;
 import org.apache.tajo.storage.Tuple;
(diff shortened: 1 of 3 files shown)


tajo-core/src/main/java/org/apache/tajo/engine/planner/physical/SortBasedColPartitionStoreExec.java
```diff
@@ -36,35 +36,24 @@
...
-  }
-
-  private void fillKeyTuple(Tuple inTuple, Tuple keyTuple) {
-    for (int i = 0; i < keyIds.length; i++) {
-      keyTuple.put(i, inTuple.get(keyIds[i]));
-    }
-  }
+  private transient StringBuilder sb = new StringBuilder();
 
   private String getSubdirectory(Tuple keyTuple) {
```
jihoonson: Input tuple seems not to be a key tuple anymore. Please change its name, too.

jihoonson:
Hi @navis. Your patch looks good to me. I left only a single trivial comment.
Additionally, I have a question about ```TupleType``` even though this is not related to this patch. Obviously, ```TupleType``` represents the data type rather than tuple type (even I have no idea what tuple type means). I wonder why you don't use just ```TajoDataTypes.Type```. 

Finally, truly sorry for late review, but would you rebase your patches? We have resolved the problem of jvm crash at https://issues.apache.org/jira/browse/TAJO-1568. If you trigger CI by rebasing your patches, it is definitely helpful for review.

hyunsik:
+1
The patch looks good to me. I rebased the patch against the latest patch. It was trivial. I manually tested this patch. All unit tests are passed.

PR #486, opened 2015-03-31
TAJO-1432: Support compressed result stream

superpiggy:


dongjoon-hyun:
As a user, I like this patch, @superpiggy . :)

PR #488, opened 2015-04-01
TAJO-1424 Investigate the problem of too many "Try to connect" messeges

navis:
Ok, let's fix this quickly. Hard to see the log.


navis:
Sorry, I've not resolved address when reconnecting.

jihoonson:
+1
I'll commit to ignore annoying messages which frequently incurs test failures. 
The fundamental problem will be solved in https://issues.apache.org/jira/browse/TAJO-1497

PR #489, opened 2015-04-01
TAJO-1494: Add SeekableScanner support to DelimitedTextFileScanner

jinossy:


tajo-storage/tajo-storage-common/src/main/java/org/apache/tajo/storage/BufferPool.java
```diff
@@ -44,7 +44,7 @@ public static long maxDirectMemory() {
   }
 
 
-  public synchronized static ByteBuf directBuffer(int size) {
+  public static ByteBuf directBuffer(int size) {
```
blrunner: Could you explain why you deleted synchronized keyword?

tajo-storage/tajo-storage-common/src/main/java/org/apache/tajo/storage/FSDataInputChannel.java
```diff
@@ -0,0 +1,67 @@
...
+ * FSDataInputChannel is a NIO channel implementation of direct read ability to read from HDFS
+ */
+public final class FSDataInputChannel extends InputChannel implements SeekableChannel {
+
+  private ReadableByteChannel channel;
+  private FSDataInputStream inputStream;
+  private boolean isDirectRead;
+
+  public FSDataInputChannel(FSDataInputStream inputStream) {
+    if (inputStream.getWrappedStream() instanceof ByteBufferReadable) {
```
blrunner: Is there no necessary for checking DFSInputStream?

tajo-storage/tajo-storage-hdfs/src/main/java/org/apache/tajo/storage/FileAppender.java
```diff
@@ -53,6 +53,10 @@ public FileAppender(Configuration conf, TaskAttemptId taskAttemptId, Schema sche
 
     try {
       if (taskAttemptId != null) {
+        if (!(conf instanceof TajoConf)) {
+          throw new IllegalArgumentException("Configuration must be a TajoConf instance");
```
blrunner: How about change the exception message as follows.
```
Configuration must be an instance of TajoConf.
(or Configuration is not an instance of TajoConf.)
```

tajo-storage/tajo-storage-common/src/main/java/org/apache/tajo/storage/BufferPool.java
jinossy: Because bufferPool does not have a critical section.

tajo-storage/tajo-storage-common/src/main/java/org/apache/tajo/storage/FSDataInputChannel.java
jinossy: Yes, you can find ByteBufferReadable inteface in getWrappedStream()

jinossy:
Thank you for the review.
I've update the exception message

jihoonson:
Accelerant work!
```DelimitedTextFileScanner``` greatly worked with the BST index.

jinossy:
In this patch, test case was lacking in content. I will add more unit test
Thanks.

jinossy:
I've add more unit test

tajo-storage/tajo-storage-common/src/main/java/org/apache/tajo/storage/BufferPool.java
blrunner: Oh! I see. :)

tajo-storage/tajo-storage-common/src/main/java/org/apache/tajo/storage/FSDataInputChannel.java
blrunner: O.K. Thanks for your feedback.

blrunner:
+1

Ship it. :)

PR #493, opened 2015-04-01
TAJO-1475: implement scanner for elasticsearch storage

HowookJeong:
I implemented scanner for elasticsearch storage which is based on tajo storage.

Thanks.


HowookJeong:
I close because elasticsearch doesn't support JDK 1.6.

PR #495, opened 2015-04-01
TAJO-1502: Implement SerDe for Kafka Storage.

hys9958:
Hi,
I have implemented for this issue and unit tests.
It implemented based on TextLineSerDe. So can be use other SerDe of based in TextLineSerDe like JsonLineSerDe.
You can specify SerDe when create table. like this,

```
CREATE TABLE <table_name> [(<column_name>
<data_type>, ... )]
using kafka with ('kafka.serde'='org.apache.tajo.storage.json.JsonLineSerDe')
```

Default SerDe is DelimitedSerDe.

Thanks.


blrunner:
Thanks for your contribution.
I'll review this within a few days.

PR #500, opened 2015-04-02
TAJO-1499: Check the bind status when EvalNode::eval() is called

jihoonson:


blrunner:
Thanks for your contribution.

I would like it more valuable if you add more unit test case for verifying the message for IllegalStateException. And could you trigger this for Travis CI build or could you add the patch to JIRA?

blrunner:
+1

Ship it. :)

PR #507, opened 2015-04-04
TAJO-1522 NPE making stage history before task scheduler is initialized

navis:
trivial


jinossy:
+1
It is possible situation when it killed or failed.

Thanks

PR #508, opened 2015-04-04
TAJO-1523 ClassSize should consider compressed oops

navis:
afaik, java object is 4xreference (which is 2 in ClassSize). There should be some debate on this.


hyunsik:
+1
This change is reasonable and looks good to me.

PR #512, opened 2015-04-06
TAJO-1529: Implement json_extract_path_text(string, string) function

jinossy:


jihoonson:
+1
Thanks for your nice work!

tajo-core/src/main/java/org/apache/tajo/engine/function/json/JsonExtractPathText.java
```diff
@@ -0,0 +1,86 @@
...
+)
+public class JsonExtractPathText extends GeneralFunction {
+  private JSONParser parser;
+
+  public JsonExtractPathText() {
+    super(new Column[]{
+        new Column("string", TajoDataTypes.Type.TEXT),
+        new Column("string", TajoDataTypes.Type.TEXT),
+    });
+    parser = new JSONParser(JSONParser.MODE_JSON_SIMPLE | JSONParser.IGNORE_CONTROL_CHAR);
```
hyunsik: It would be better if this initialization is moved into bind(Schema) method.

hyunsik: I misunderstood it. It's just a function. So, it does not have bind() method. We may need pre-initialization method like bind in Eval.

jinossy: I agree, we need pre-initialization with const params

PR #513, opened 2015-04-06
TAJO-1525: Implement INTERSECT [ALL] physical operator

sirpkt:
This patch implements sort based INTERSECT ALL physical operator.

INTERSECT ALL operator has two input physical operators that should return rows sorted by all the schema columns.
Then, it iterates results of both children operators and returns the common rows from both.

- SetTupleComparator is implemented to compare the whole columns of given two tuples.
- Test code is also included, which tests INTERSECT ALL exec by replacing inner join with it. It is somewhat tricky, however, useful way when Logical Planner and Global Planner are not ready for INTERSECT [ALL].


hyunsik:
+1
The physical operator works well, and the patch looks good to me. I'll finish the logical plan and global planner part in the following issue. Thank you for your work.

PR #516, opened 2015-04-07
TAJO-856: 'http://TajoMaster:26080/query.jsp' page dose not comfortable to viewing

seungunchoe:
Please review my patch.


jihoonson:
Would you please write the title properly?
Github bot cannot recognize that this PR is related to the appropriate issue.

PR #520, opened 2015-04-07
TAJO-1538 TajoWorkerResourceManager.allocatedResourceMap is increasing forever

navis:
seemed mistake from TAJO-602 (tajo-0.9.0)


jinossy:
+1 LGTM

PR #521, opened 2015-04-07
TAJO-1526: Investigate the frequent failures of Travis CI

jihoonson:
This PR is to find the origin of the problem.


jinossy:
+1

jihoonson:
Thanks @jinossy. 

The problem seems the memory overflow, especially related to the permanent generation space.
So, I increased the ```permSize``` from 128 MB to 152 MB in this patch.
Additionally, I reduced the number of test tuples  of TestExternalSortExec from 3,000,000 to 100,000. 
If there are any objections, I'll commit soon.

tajo-core/src/test/java/org/apache/tajo/engine/planner/physical/TestExternalSortExec.java
```diff
@@ -61,7 +61,7 @@
   private LogicalPlanner planner;
   private Path testDir;
 
-  private final int numTuple = 3000000;
```
navis: This is fixed for proper test of ExternalSortExec by request (It's not using merge sort but memory scanner). I'll fix that in a following issue.

jihoonson:
Right. So, I also decreased the sort buffer size to test external sort.

navis:
ah, good.

PR #522, opened 2015-04-07
TAJO-1534: DelimitedTextFile return null instead of a NullDatum 

jinossy:


tajo-storage/tajo-storage-hdfs/src/main/java/org/apache/tajo/storage/rcfile/RCFile.java
```diff
@@ -1255,16 +1255,21 @@ public void init() throws IOException {
...
           skippedColIDs[tid] = false;
 
-          SelectedColumn col = new SelectedColumn();
           col.colIndex = tid;
           col.runLength = 0;
           col.prvLength = -1;
           col.rowReadIndex = 0;
           selectedColumns[i] = col;
           colValLenBufferReadIn[i] = new NonSyncDataInputBuffer();
+        }  else {
```
blrunner: The newly added code is the mostly same as those of existing codes.

jinossy: Thank you for nice finding.
I've remove the duplicated codes

blrunner:
+1

It looks good to me and all unit test cases finished successfully.
And if possible, could you add 0.11.0 to the fix version of JIRA issue?

jinossy:
@blrunner 
Thank you for your review. I will commit soon

PR #523, opened 2015-04-07
TAJO-1537: Implement a virtual table for sessions

superpiggy:


hyunsik:
This work looks good to me. But, it only handles pre-defined session variables. It would be better if this session table can handle all user session variables.

You look busy. If you don't mind, I'll improve the session table to support all user variables.

hyunsik:
In addition, this patch allows each user to see all session variables of all other users. It seems to be dangerous. It would be better to allow a user to see its own session's variables. I'll change it as I mentioned.

PR #524, opened 2015-04-07
investigate jvm permSize after remove jersey warning

jinossy:
This PR just is finding the CI failure.


jinossy:
resolved by TAJO-1568

PR #525, opened 2015-04-08
TAJO-1530: Display warn message when the query kill button is clicked in Web UI

dongjoon-hyun:
You can test clicking kill button confirmation during the following command.
```
tsql -c 'create table t(a int)'
tsql -c 'insert into t select 1'
tsql -c 'select a from t union select sleep(10) from t'
```

### DIFF
--- a/tajo-core/src/main/resources/webapps/admin/query.jsp
+++ b/tajo-core/src/main/resources/webapps/admin/query.jsp
@@ -89,20 +89,22 @@
     <script type="text/javascript">
 
     function killQuery(queryId) {
-        $.ajax({
-            type: "POST",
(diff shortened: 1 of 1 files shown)


dongjoon-hyun:
Hi, @babokim . Could you review your issue? The change is just trivial Javascript code.

babokim:
This patch looks good to me. Here is my +1.
The Travis CI failed but it may be caused by TAJO-1526. This issue is trivial and not related any test case. All tests passed on my laptop. So I'll commit soon.

dongjoon-hyun:
Thank you, @babokim !  :)

PR #528, opened 2015-04-08
TAJO-1541 Connection timeout in netty client is not working

navis:
It does not abandon retrying connection to master. I like that but it's a bug.


ykrips:
+1
Though Travis CI build system has an issue when running testcases, but this build issue is far from this patch. Local build on my laptop has passed testcases, and it runs with JDK 6.0.

PR #529, opened 2015-04-09
TAJO-1542 Refactoring of HashJoinExecs

navis:


hyunsik:
Hi @navis,

The patch looks good to me. Even though this patch is conflict with the latest master branch, the conflict is trivial. So, I could merge it without further work. I'm reviewing this patch, and I'll give comments soon.

hyunsik:
Excellent work! This patch simplifies all hash-based join algorithms. I think that this patch seems to be ready to be committed. 

BTW, it would be better if we add more comments. BTW, you may have some time slots to do this work as far as I know. So, I'm going to add comments. If you have any opinion, feel free to suggest anything.

hyunsik:
I add some trivial comments, and removed PhysicalPlanner::newIterator, which is never used in this patch. For review, I'll make another pull request.

tajo-catalog/tajo-catalog-common/src/main/java/org/apache/tajo/catalog/Schema.java
```diff
@@ -383,6 +392,17 @@ public SchemaProto getProto() {
     return builder.build();
 	}
 
+  public Set<String> getAliases() {
```
hyunsik: Is it only for debugging? It's not bad.

tajo-core/src/main/java/org/apache/tajo/engine/planner/PhysicalPlannerImpl.java
```diff
@@ -464,14 +464,14 @@ private PhysicalExec createLeftOuterJoinPlan(TaskAttemptContext context, JoinNod
         case IN_MEMORY_HASH_JOIN:
           LOG.info("Left Outer Join (" + plan.getPID() +") chooses [Hash Join].");
           return new HashLeftOuterJoinExec(context, plan, leftExec, rightExec);
-        case NESTED_LOOP_JOIN:
-          //the right operand is too large, so we opt for NL implementation of left outer join
-          LOG.info("Left Outer Join (" + plan.getPID() +") chooses [Nested Loop Join].");
-          return new NLLeftOuterJoinExec(context, plan, leftExec, rightExec);
+        case MERGE_JOIN:
```
hyunsik: Good job. It should have removed earlier.

hyunsik:
Actually, our join algorithms are very old ones. We should have refactored and optimized them, but we haven't. I appreciate your work. I like this change. I'm not sure if you can do further work for this patch. So, I just created another pull request with some modifications, more comments and removal of unused methods.

https://github.com/apache/tajo/pull/567

PR #531, opened 2015-04-11
TAJO-1548 Refactoring condition code for CHAR into CatalogUtil

charsyam:
Currently Schema's addColumn is this
```java
public synchronized Schema addColumn(String name, Type type) {
if (type == Type.CHAR)
{ return addColumn(name, CatalogUtil.newDataTypeWithLen(type, 1)); }
return addColumn(name, CatalogUtil.newSimpleDataType(type));
}
```
I think it is better condition code into CatalogUtil.newSimpleDataType.
because some testcode make columns just using CatalogUtil.

### DIFF
--- a/tajo-catalog/tajo-catalog-common/src/main/java/org/apache/tajo/catalog/CatalogUtil.java
+++ b/tajo-catalog/tajo-catalog-common/src/main/java/org/apache/tajo/catalog/CatalogUtil.java
@@ -24,7 +24,6 @@ import org.apache.tajo.DataTypeUtil;
 import org.apache.tajo.TajoConstants;
 import org.apache.tajo.catalog.partition.PartitionMethodDesc;
 import org.apache.tajo.catalog.proto.CatalogProtos;
-import org.apache.tajo.catalog.proto.CatalogProtos.ColumnProto;
 import org.apache.tajo.catalog.proto.CatalogProtos.SchemaProto;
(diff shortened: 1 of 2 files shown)


blrunner:
Hi @charsyam 

Could you explain this PR in detail? I don't quite understand this.

charsyam:
Hi, @blrunner 
CHAR type can have length. :)
But if we make CHAR Column using this command

in tajo-storage/tajo-storage-hdfs/src/test/java/org/apache/tajo/storage/parquet/TestReadWrite.java
```java
columns.add(new Column("mychar", Type.CHAR)); 
``` 

but it's length is 0, because it is just using CatalogUtil.newSimpleDataType

and Scheme class already use condition code in addColumn method.
But I think it can cause some mistakes easily.

So I think it is better to move this condition code into CatalogUtil.newSimpleDataType
What do you think?

jihoonson:
Hi @charsyam and @blrunner, thanks for your work. 
As @charsyam said, ```CHAR``` type can have a variable length. The code in ```Schema::add()``` seems to give a default length for this type. I agree on that this patch will reduce mistakes because many other codes use ```CatalogUtil.newSimpleDataType()``` directly.

blrunner:
Hi @jihoonson 

Thanks for your review.

Hi @charsyam 

+1
Thanks for your kind comment.
I also agree with your suggestion.

I'll commit this soon. :)

PR #532, opened 2015-04-13
TAJO-1550: Implements JUL Adapter for log4j 1.2

ykrips:


ykrips:
Uploaded one missing commit for this patch, and this is a final version of this task.

PR #533, opened 2015-04-14
TAJO-1497: RPC client does not share a connection

jinossy:
I've change RpcConnectionPool to RpcConnectionManager and remove the release codes.
Because netty rpc does not need pooling.


jihoonson:
I think this patch is very important to make Tajo more stable. So, I'll review it first.

jinossy:
I've rename to RpcClientManager

tajo-client/src/main/java/org/apache/tajo/client/SessionConnection.java
```diff
@@ -286,13 +286,13 @@ public void close() {
...
     try {
-      client = connPool.getConnection(getTajoMasterAddr(), TajoMasterClientProtocol.class, false);
+      client = manager.getConnection(getTajoMasterAddr(), TajoMasterClientProtocol.class, false);
       TajoMasterClientProtocolService.BlockingInterface tajoMaster = client.getStub();
       tajoMaster.removeSession(null, sessionId);
     } catch (Throwable e) {
       // ignore
     } finally {
-      connPool.releaseConnection(client);
+      manager.cleanup(client);
```
jihoonson: This is a static function.

jinossy: Thank you for nice finding

tajo-rpc/tajo-rpc-protobuf/src/main/java/org/apache/tajo/rpc/ServerCallable.java
```diff
@@ -33,17 +33,17 @@
...
 
   public abstract T call(NettyClientBase client) throws Exception;
 
-  public ServerCallable(RpcConnectionPool connPool,  InetSocketAddress addr, Class<?> protocol, boolean asyncMode) {
+  public ServerCallable(RpcClientManager connPool,  InetSocketAddress addr, Class<?> protocol, boolean asyncMode) {
     this(connPool, addr, protocol, asyncMode, false);
   }
 
-  public ServerCallable(RpcConnectionPool connPool, InetSocketAddress addr, Class<?> protocol,
+  public ServerCallable(RpcClientManager manager, InetSocketAddress addr, Class<?> protocol,
```
jihoonson: This constructor seems to be not used anymore. How about removing it?

jinossy: I've remove constructor

jinossy:
I've enable idle timeout. If all requests is done and client is idle state, client will be closed
```
public static final int RPC_IDLE_TIMEOUT = 43200; // 12 hour
```

jihoonson:
+1
thanks for the nice patch!
I think that we can make our rpc cleaner with this patch.

PR #538, opened 2015-04-17
TAJO-1560: HashShuffle report should be ignored when a succeed tasks are not included

jinossy:
Additionally, I’ve remove unnecessary synchronized block in HistoryWriter


tajo-core/src/main/java/org/apache/tajo/querymaster/Stage.java
```diff
@@ -1300,6 +1301,52 @@ protected void stopFinalization() {
...
+        }
+      }
+    } else {
+      LOG.info(getId() + ", Received " + type + " reports " +
+          completedShuffleTasks.get() + "/" + succeededObjectCount);
+    }
+  }
+
+  /**
+   * HASH_SHUFFLE, SCATTERED_HASH_SHUFFLE should get report from worker nodes when ExecutionBlock is stopping.
```
jihoonson: It would be great if you add a comment that describes why we don't need to collect reports when the shuffle type is RANGE_SHUFFLE.

jihoonson:
+1. I just a minor comment. Please address before commit.

jinossy:
Thank you for your review.
I will commit it soon after I upload patch for branch-0.10.1 in jira

PR #540, opened 2015-04-17
TAJO-1568: Apply UnpooledByteBufAllocator when a tajo.test.enabled is set to enable.

jinossy:


jinossy:
BufferPool allocator is to be UnpooledByteBufAllocator when a tajo.test.enabled is set to enable. 
IMO, we don’t need pooling for testing.

jihoonson:
+1 thanks for your nice work!
This patch will fix the problem of jvm crash!

jinossy:
Thank you for quick review!!

PR #541, opened 2015-04-18
TAJO-1569: BlockingRpcClient can make other request fail

jinossy:


tajo-rpc/tajo-rpc-protobuf/src/main/java/org/apache/tajo/rpc/BlockingRpcClient.java
```diff
@@ -188,10 +187,12 @@ protected void channelRead0(ChannelHandlerContext ctx, RpcResponse rpcResponse)
     @Override
     public void exceptionCaught(ChannelHandlerContext ctx, Throwable cause)
         throws Exception {
+      /* Current requests will be failed */
       for(ProtoCallFuture callback: requests.values()) {
         callback.setFailed(cause.getMessage(), cause);
       }
-      
+      requests.clear();
```
hyunsik: This change will remove other request handles, prohibiting subsequent responses. It should be done only if we cannot able to recover this rpc connection. I'm not sure that ``exceptionCaught`` means a unrecoverable situation.

hyunsik:
Even though It is trivial change, it seems to be a critical issue. Other changes look good to me. I left one comment.

tajo-rpc/tajo-rpc-protobuf/src/main/java/org/apache/tajo/rpc/BlockingRpcClient.java
jinossy: This case is fatal error and  It will be internal error.
All request is set to fail and then requests should be removed.

hyunsik: I missed the above line, which marks all requests as 'failed'. This code also seem to be based on the assumption that exceptionCaught does not handle any recoverable exception. If it is right, please go ahead. Otherwise, we can fix the bug of TAJO-1569 right now, and we can handle this problem in another jira.

jinossy: You are right, exceptionCaught do not received recoverable exception(RemoteCallException) but if we do handle the recoverable exception, It would be better.
I will add more exception handling in TAJO-1563

hyunsik: agreed

hyunsik:
+1 Ship it

jinossy:
Thank you for the review.

jinossy:
branch-0.10.1 is required TAJO-1497
I will create new jira task

PR #542, opened 2015-04-18
TAJO-1570 CatalogUtil newSimpleDataTypeArray should use newSimpleDataType

charsyam:
It is better CatalogUtil's newSimpleDataTypeArray method use newSimpleDataType. Because their purpose is the same. 
and it is efficient when someone change newSimpleDataType method.
```java
public static DataType [] newSimpleDataTypeArray(Type... types) {
DataType [] dataTypes = new DataType[types.length];
for (int i = 0; i < types.length; i++)
{ dataTypes[i] = DataType.newBuilder().setType(types[i]).build(); }
return dataTypes;
}
```


jihoonson:
This change is trivial, but will reduce many mistakes.
Would you trigger Travis or Jenkins again after rebasing?

charsyam:
@jihoonson I rebased it. and it passed :)

jihoonson:
+1 thanks!

PR #545, opened 2015-04-19
TAJO-1573: Fix addColumn bug (fieldsByName should maintain a list per a simple field name)

dongjoon-hyun:


dongjoon-hyun:
`addColumn` function is fixed. `getNaturalJoinCondition`, `verifyProjectedFields`, `verifyIfTargetsCanBeEvaluated` functions are changed together.

jihoonson:
Test cases are also missed in this patch.

tajo-catalog/tajo-catalog-common/src/test/java/org/apache/tajo/catalog/TestSchema.java
```diff
@@ -147,13 +147,26 @@ public final void testGetColumnString() {
 		assertEquals(col3, schema.getColumn("addr"));
 	}
 
-	@Test
-	public final void testAddField() {
```
jihoonson: This test looks to verify Schema.addColumn() with an unqualified column name. Why don't you add a new test instead of modifying an existing test?

tajo-plan/src/main/java/org/apache/tajo/plan/LogicalPlanner.java
```diff
@@ -485,9 +485,9 @@ public static void verifyProjectedFields(QueryBlock block, Projectable projectab
 
         for (int i = 0; i < groupingKeyNum; i++) {
           Target target = groupbyNode.getTargets()[i];
-          if (groupbyNode.getTargets()[i].getEvalTree().getType() == EvalType.FIELD) {
+          if (target.getEvalTree().getType() == EvalType.FIELD) {
             FieldEval grpKeyEvalNode = target.getEvalTree();
-            if (!groupbyNode.getInSchema().contains(grpKeyEvalNode.getColumnRef())) {
+            if (!groupbyNode.getInSchema().containsByQualifiedName(grpKeyEvalNode.getColumnRef().getQualifiedName())) {
```
jihoonson: Schema.contains() internally calls the same logic with containsByQualifiedName(). Would you explain why you changed this line?

tajo-plan/src/main/java/org/apache/tajo/plan/LogicalPlanner.java
```diff
@@ -1200,8 +1200,8 @@ private static EvalNode getNaturalJoinCondition(JoinNode joinNode) {
     Column rightJoinKey;
 
     for (Column common : commons.getColumns()) {
-      leftJoinKey = leftSchema.getColumn(common.getQualifiedName());
-      rightJoinKey = rightSchema.getColumn(common.getQualifiedName());
+      leftJoinKey = leftSchema.getColumn(leftSchema.getColumnIdByName(common.getQualifiedName()));
```
jihoonson: Would you explain why you changed this line?

PR #546, opened 2015-04-20
TAJO-1556: "insert into select" with reordered column list does not work.

superpiggy:
fixed a bug in buildProjectedInsert.


jihoonson:
Thanks for your work. 
I've looked over your patch. It looks good to me, but will be great if you add a test for query execution in addition to the existing planner test.

PR #547, opened 2015-04-20
TAJO-1552: NPE occurs when GreedyHeuristicJoinOrderAlgorithm.getCost() returns infinity.

babokim:
I added TestJoinOrderAlgorithm test case for this issue and also added "ignoreOrdering" parameter in assertResultSet() method.


jihoonson:
Hi @babokim. This patch is closely related to TAJO-1352. I'll review soon.

tajo-core/src/test/java/org/apache/tajo/QueryTestCaseBase.java
```diff
@@ -439,7 +437,18 @@ public ResultSet executeJsonFile(String jsonFileName) throws Exception {
...
+  }
+
+  /**
+   * Assert the equivalence between the expected result and an actual query result.
+   * If it isn't it throws an AssertionError.
+   *
+   * @param result Query result to be compared.
+   * @param ignoreOrdering If true, don't concern ordering.
+   */
+  public final void assertResultSet(ResultSet result, boolean ignoreOrdering) throws IOException {
```
jihoonson: I think that this is not a good convention. We need to make tajo to be more determinant including tests. 
How about adding order-by clauses to test queries?

jihoonson:
Hi @babokim. The patch looks good in overall. I left one comment.
In addition, I think that we need to discuss how writing query plan tests. I opened a jira issue for it. Please see https://issues.apache.org/jira/browse/TAJO-1577.

jihoonson:
Thanks @babokim.
This patch looks good to me, but requires to be rebased against master.
I opened #655 for it.

PR #548, opened 2015-04-20
TAJO-1574: Fix NPE on natural join

dongjoon-hyun:
Please see the description in https://issues.apache.org/jira/browse/TAJO-1574 .


jihoonson:
Great. Would you please add test cases?

dongjoon-hyun:
Sure. The changed codes are in `tajo-algebra` and `tajo-plan`. I will add a testcase to cover both cases in somewhere.

jihoonson:
Well, it's good, but we need to verify query results in most cases, especially when the bug is related to query execution.

dongjoon-hyun:
Oh, I see what you mean.

dongjoon-hyun:
Hi, @jihoonson . I finished adding a testcase.

tajo-core/src/test/java/org/apache/tajo/engine/query/TestJoinQuery.java
```diff
@@ -1191,4 +1192,16 @@ public final void testJoinWithOrPredicates() throws Exception {
     assertResultSet(res);
     cleanupQuery(res);
   }
+
+  @Test
+  public final void testNaturalJoin() throws Exception {
+    ResultSet res = null;
+    try {
+      res = executeQuery();
+    } catch (Exception e) {
```
jihoonson: This catch block is redundant because the testNaturalJoin method throws Exception. I'll remove it before commit.

jihoonson:
+1 LGTM.

PR #550, opened 2015-04-21
TAJO-1576: Sometimes DefaultTajoCliOutputFormatter.parseErrorMessage() eliminates an important kind of information

eminency:
After the patch is merged, some syntax error message is displayed like the following.

Before the patch :
```
ERROR: syntax error at or near '('
```
After the patch: 
```
ERROR: syntax error at or near '('
LINE 31:4      (select *
               ^
```


eminency:
Now there is no test for error line number.
It will be added while I work on TAJO-1580.

Please consider it.

PR #551, opened 2015-04-23
TAJO-1562: Python UDAF support

jihoonson:


jihoonson:
This patch is ready for review. Users can define their own Python UDAFs by defining a class containing some mandatory functions, i.e., reset(), eval(), merge(), get_partial_result(), and get_final_result(). Here is a typical example.

```
  class AvgPy:
    sum = 0
    cnt = 0

    def __init__(self):
        self.reset()

    def reset(self):
        self.sum = 0
        self.cnt = 0

    # eval at the first stage
    def eval(self, item):
        self.sum += item
        self.cnt += 1

    # get intermediate result
    def get_partial_result(self):
        return [self.sum, self.cnt]

    # merge intermediate results
    def merge(self, list):
        self.sum += list[0]
        self.cnt += list[1]

    # get final result
    @output_type('float8')
    def get_final_result(self):
        return self.sum / float(self.cnt)
```

The additional restriction is that the return type of ```get_partial_result()``` and the input type of ```merge()``` must be same.

I'd like to give a brief explanation of how Python UDAFs are executed. During aggregation, Tajo currently keeps intermediate aggregation state per key in each implementation of aggregation operators. This approach also should be applied for Python functions to control resource usage and avoid a huge change. To do so, intermediate aggregation state of Python UDAF should be sent to Tajo workers whenever performing aggregation for different keys. However, I don't want for users to implement serialization/deserialization of intermediate aggregation state. 

So, I simply used json for that. The snapshot of the instance of Python UDAF is sent to Tajo workers when necessary. 

Let me consider an example. When a tuple (1, 10) is read and the aggregation key is 1, Tajo creates a PythonAggFunctionContext for the aggregation key and call AvgPy.eval() with an input 10. Here, the intermediate aggregation state, i.e. AvgPy.sum = 10 and AvgPy.cnt = 1, is kept in the instance of AvgPy instead of PythonAggFunctionContext. However, when a tuple (2, 5) is read as a next input, Tajo gets intermediate aggregation state from the Python controller, and stores it in PythonAggFunctionContext. After that, the instance of AvgPy is reset to compute aggregation for another key 2.

Followings are some highlights of remaining implementation.
* I've improved Python controller to support UDAF as well as UDF.
 * For UDAF execution, the Python controller executes some pre-defined functions (reset(), eval(), merge(), get_partial_result(), and get_final_result()). 
* I've improved AggregationFunctionCallEval to support script functions via AggFunctionInvoke.
 * AggFunctionInvoke is an abstract class to represent how the function is executed. There are two subclasses, ClassBasedAggFunctionInvoke and PythonAggFunctionInvoke.

Finally, I've added several test functions, but some of them are disabled. This is due to our rigid function syntax. Currently, function names are explicitly defined in our parser, and their types are identified during parsing SQLs. This architecture should be improved to support functions defined by users.

hyunsik:
Great job! The interface looks nice to me. Even though some parts seem to need more optimization, it would take more time. I think that the current patch seems to be ready to be committed because the interface and the code skeleton are great. We can do more optimization later.

I have one comment. Could you rename isUdf to others like isScalarFunction or isAggregationFunction? It would be more intuitive, I think.

jihoonson:
Thanks @hyunsik. I've changed isUdf to isScalarFunction.

jinossy:
+1 Great work!

jihoonson:
@jinossy @hyunsik, thanks for your reviews!

PR #552, opened 2015-04-24
TAJO-1580: Error line number is incorrect

eminency:
Modified to avoid adding useless new line characters


blrunner:
+1

Thanks for your contribution.
I found that this had printed right line at TajoMaster log file.

PR #553, opened 2015-04-24
TAJO-1581: Does not update last state of query stage in non-hash shuffle

jinossy:


tajo-core/src/main/java/org/apache/tajo/querymaster/Stage.java
```diff
@@ -1405,14 +1405,14 @@ public void run() {
               stage.timeoutChecker.start();
             }
           } else {
-            stage.handle(new StageEvent(stage.getId(), StageEventType.SQ_STAGE_COMPLETED));
+            stage.getEventHandler().handle(new StageEvent(stage.getId(), StageEventType.SQ_STAGE_COMPLETED));
```
jihoonson: nice finding.

jihoonson:
+1 ship it!

PR #554, opened 2015-04-24
TAJO-1563: Improve RPC error handling

jinossy:


jinossy:
I’ve test of two failure case. It can resolve by TAJO-1215 and task re-assign
* Node shutdown during query running
* Node shutdown before query start

blrunner:
Thanks for your contribution.
If you fix bugs according to FindBugs report, it would more nice patch. :)

jinossy:
I've remove findbug warnings
Thanks

tajo-client/src/main/java/org/apache/tajo/client/SessionConnection.java
```diff
@@ -87,6 +87,10 @@ public SessionConnection(ServiceTracker tracker, @Nullable String baseDatabase,
     this.properties = properties;
 
     this.manager = RpcClientManager.getInstance();
+    this.manager.setRetries(properties.getInt(RpcConstants.RPC_CLIENT_RETRY_MAX, RpcConstants.DEFAULT_RPC_RETRIES));
+    this.manager.setTimeoutSeconds(
+        properties.getInt(RpcConstants.RPC_CLIENT_TIMEOUT_SECS, 0)); // disable rpc timeout
```
blrunner: Could you explain why you disabled rpc timeout?

tajo-core/src/main/java/org/apache/tajo/master/rm/TajoResourceTracker.java
```diff
@@ -98,17 +98,17 @@ public void serviceInit(Configuration conf) {
     systemConf.setVar(TajoConf.ConfVars.RESOURCE_TRACKER_RPC_ADDRESS, NetUtils.normalizeInetSocketAddress(bindAddress));
 
     LOG.info("TajoResourceTracker starts up (" + this.bindAddress + ")");
-    super.start();
+    super.serviceInit(conf);
```
blrunner: This is a good point. :)

tajo-core/src/main/java/org/apache/tajo/querymaster/QueryMaster.java
```diff
@@ -341,6 +339,7 @@ public void stopQuery(QueryId queryId) {
 
         QueryCoordinatorProtocolService masterClientService = tmClient.getStub();
         masterClientService.heartbeat(future.getController(), queryHeartbeat, future);
+        future.get(10, TimeUnit.SECONDS);
```
blrunner: Could you explain why you get 10 seconds?

jinossy: It can handle the RPC error. Have you better idea?

tajo-client/src/main/java/org/apache/tajo/client/SessionConnection.java
jinossy: IMO, It should be handled in session timeout. 
If timeout is set to properties, it will work

blrunner:
+1

Thanks for your comments.
I also agree with your opinions.
Ship it. :)

tajo-core/src/main/java/org/apache/tajo/querymaster/QueryMaster.java
blrunner: If you separate 10 seconds to RpcConstants, it would be better nice.

jinossy: Sure!

jinossy:
Thank you for your review!
I'll commit soon

PR #558, opened 2015-04-29
TAJO-1577 Add test cases to verify join plans

navis:
@jihoonson Could you review this? If it's done in right way, I'll apply this to all tests in Test*JoinQuery


jihoonson:
Thanks @navis. I'll roughly review right now.

jihoonson:
Looks great! The test failure is due to the too much memory usage in python udfs. It's my fault. Sorry.
This problem will be solved if you rebase.

I have one suggestion. Even though the annotation name is ```SimpleTest```, it looks very proper for our all query tests. 
I think that we should test more complex and long queries. For such queries, it would be difficult to write sqls in the ```SimpleTest``` annotation. How about extending it to support sql files like the current query tests?

PR #559, opened 2015-04-29
TAJO-1584: Remove QueryMaster client sharing in TajoMaster and TajoWorke...

jinossy:
...r


jinossy:
I’ve add option to reuse configuration.
QUERY_MASTER_RPC_CLIENT_REUSE("tajo.querymaster.rpc.client.reuse", false)

tajo-core/src/main/java/org/apache/tajo/master/QueryInProgress.java
```diff
@@ -156,8 +161,15 @@ public boolean startQueryMaster() {
   private void connectQueryMaster() throws Exception {
     InetSocketAddress addr = NetUtils.createSocketAddr(queryInfo.getQueryMasterHost(), queryInfo.getQueryMasterPort());
     LOG.info("Connect to QueryMaster:" + addr);
-    queryMasterRpc =
-        RpcClientManager.getInstance().getClient(addr, QueryMasterProtocol.class, true);
+
+    /* A large scale cluster can be connection refused, if it reuse  */
```
hyunsik: It would be better if it should be changed to 'Connection can be refused in large scale clusters if 'reuse' option is TRUE'.

In addition, this description should be also described in ```QUERY_MASTER_RPC_CLIENT_REUSE``` in TajoConf and user guide.

hyunsik:
In this patch, the connection reuse is changed to be optional. In my opinion, it would be better if we completely remove the reuse feature. Actually, I'm not sure if anyone want 'reuse feature'.

Others look good to me.

jinossy:
I agree with you. I will remove the 'reuse feature'

jinossy:
@hyunsik 
Could you review again?

hyunsik:
+1 ship it!!

PR #560, opened 2015-04-30
TAJO-1586: TajoMaster HA startup failure on Yarn.

blrunner:
I tested this successfully on production cluster and I fixed various bugs related TajoMaster HA. For the reference, I replaced master IP address to master host name on Tajo web interface.


tajo-core/src/main/java/org/apache/tajo/ha/HdfsServiceTracker.java
```diff
@@ -70,12 +70,45 @@
...
     this.conf = conf;
+  }
+
+  @Override
+  public void register() throws IOException {
     initSystemDirectory();
 
     InetSocketAddress socketAddress = conf.getSocketAddrVar(ConfVars.TAJO_MASTER_UMBILICAL_RPC_ADDRESS);
-    this.masterName = socketAddress.getAddress().getHostAddress() + ":" + socketAddress.getPort();
+    this.masterName = socketAddress.getAddress().getHostName()+ ":" + socketAddress.getPort();
```
jinossy: If a hostname is not FQDN, it may be throw the unknown host exception.
In my opinion, ip address is more safety. How do you think?

blrunner: Thanks for your review.
I also think that ip address would be more safety compare than host name.

blrunner:
Hi @jinossy 

I updated the patch. Could you check it again? :)

tajo-common/src/main/java/org/apache/tajo/conf/TajoConf.java
```diff
@@ -142,6 +142,8 @@ public static int setDateOrder(int dateOrder) {
     // High availability configurations
     TAJO_MASTER_HA_ENABLE("tajo.master.ha.enable", false, Validators.bool()),
     TAJO_MASTER_HA_MONITOR_INTERVAL("tajo.master.ha.monitor.interval", 5 * 1000), // 5 sec
+    TAJO_MASTER_HA_CLIENT_RETRY_MAX_NUM("tajo.master.ha.client.read.retry.max-num", 20),
+    TAJO_MASTER_HA_CLIENT_RETRY_WAIT_TIME("tajo.master.ha.client.read.wait-time", 5 * 1000), // 5 sec
```
jinossy: IMO, 5 sec is too long. What do you think to change (500 ms * 120 retry) ?

tajo-core/src/main/java/org/apache/tajo/ha/HdfsServiceTracker.java
```diff
@@ -184,15 +181,31 @@ private void createMasterFile(boolean isActive) throws IOException {
...
+    // Set active status and delete existing file.
     if (isActive) {
       isActiveStatus = true;
+      Path backupFile = new Path(backupPath, fileName);
+      if (fs.exists(backupFile)) {
+        fs.delete(backupFile, true);
+      }
     } else {
       isActiveStatus = false;
+      Path activeFile = new Path(activePath, fileName);
```
jinossy: Can you refactor the duplicate codes?

PR #561, opened 2015-04-30
TAJO-1576: Sometimes DefaultTajoCliOutputFormatter.parseErrorMessage() eliminates an important kind of information

eminency:
After the patch is merged, some syntax error message is displayed like the following.

Before the patch :
```
ERROR: syntax error at or near '('
```
After the patch: 
```
ERROR: syntax error at or near '('
LINE 31:4      (select *
               ^
```


eminency:
I closed previous PR by my mistake.
It couldn't be re-opened, thus re-created again.

PR #572, opened 2015-05-16
TAJO-1607 Tajo Rest Cache-Id should be bigger than zero

charsyam:
currently, tajo rest api can return cache-id that is smaller than 0.
but it is not beautiful, so I think it is better cache id is begger than 0
```
{'link': 'http://127.0.0.1:26880/rest/databases/default/queries/q_1431795664397_0008/result/-2197002215675989303', 'id': -2197002215675989303}
```


hyunsik:
+1

the change looks straightforward.

hyunsik:
@charsyam I've just committed it to master branch. I missed to add an github command to close this issue. Could you please close this ticket?

PR #574, opened 2015-05-17
TAJO-1609 query rest api uri should change from /databases/{database_name}/queies to /queries/{database_name}

charsyam:
currently, "databases" Rest Api uri has too many meaning.
1. for query
2. to manage databases info

and databases/{database_name}/queries is not proper for queries 

so I suggest change from "/databases/{database_name}/queries" to
"/queries/{database_name}", it is more clear to understand.


hyunsik:
I agree with your suggestion. ``/queries`` looks intuitive. Actually, I need to read carefully REST API draft specification. After than, I'll finish the review. 

Thanks!

charsyam:
@hyunsik I'm sorry. I will close this issue. and I will make new PR.

PR #577, opened 2015-05-18
TAJO-1599: Implement NodeResourceManager and Status updater

jinossy:


jihoonson:
Hi @jinossy, it's really nice work. 
I've almost read your patch. I'll finish my review soon.

jihoonson:
+1 LGTM!

jinossy:
Thank you for the review!
I'll commit soon

hyunsik:
This patch requires updating documentation because it affects the resource configuration.

jinossy:
I was add the new key of disk configuration because if I change the configuration, it affects the current resource. 
When I remove the legacy configuration, I will update the document.

PR #579, opened 2015-05-21
TAJO-1464: Add ORCFileScanner to read ORCFile table

eminency:
This patch is to support to read ORC file in TAJO.

The code under 'thirdparty' is from Presto, thus you don't have to look into it deeply.

Adding more data type to support is leaved as future work.


blrunner:
Hi @eminency 

Thanks for your contribution.
If you add ORC type to HiveCatalogStore, the patch would be more better.

tajo-storage/tajo-storage-hdfs/src/main/java/org/apache/tajo/storage/orc/OrcScanner.java
```diff
@@ -0,0 +1,257 @@
...
+
+import java.io.IOException;
+import java.util.HashSet;
+import java.util.Map;
+import java.util.Set;
+
+/**
+ * OrcScanner for reading ORC files
+ */
+public class OrcScanner extends FileScanner {
```
blrunner: How about renaming it ORCScanner? Currently, the scanner for RCFile is named RCScanner.

blrunner:
1. Could you explain me why you use presto-orc? 
2. You need to add unit test cases to TestStorages in tajo-storage-hdfs module.

eminency:
Hi, @blrunner 

1. Presto-orc is tested and stable. And it is a bit more faster than Hive. Secondly, Hive processes ORC with row-oriented, but Presto does with column-oriented. Because ORC is columnar format, Presto's way may be more suitable. Though Tajo processes it with tuple-oriented, it can be helpful when Tajo will change to use offheap in the future.

2. Sure.
Anyway, **please hold the review**. About new storage manager code, I'm modifying.

tajo-storage/tajo-storage-hdfs/src/main/java/org/apache/tajo/storage/orc/OrcScanner.java
eminency: Ok, I will do it.

eminency:
Hi, @blrunner.

About TestStorage, it requires Appender class, too. So it may be added in ORCAppender patch later.

Please review at current state.

blrunner:
Hi @eminency 

Thank you for your response.
Could I start to review current patch?

eminency:
Yes, please.

tajo-storage/tajo-storage-hdfs/src/main/java/org/apache/tajo/storage/orc/ORCScanner.java
```diff
@@ -0,0 +1,328 @@
...
+  private int batchSize = 0;
+
+  public ORCScanner(Configuration conf, final Schema schema, final TableMeta meta, final Fragment fragment) {
+    super(conf, schema, meta, fragment);
+  }
+
+  private Vector createOrcVector(TajoDataTypes.DataType type) {
+    switch (type.getType()) {
+      case INT1: case INT2: case INT4: case INT8:
+      case UINT1: case UINT2: case UINT4: case UINT8:
```
hyunsik: Unsigned types are not used in the current implementation. So, you can omit them.

tajo-storage/tajo-storage-hdfs/src/main/java/org/apache/tajo/storage/orc/ORCScanner.java
```diff
@@ -0,0 +1,328 @@
...
+      case INET4:
+      case TIMESTAMP:
```
hyunsik: Does ORC use long type for timestamp, date, and INET4? 

As far as I know, Tajo represents INET4 or date values as a integer value and timestamp as a long value. So, if you use LongVector for those types, it won't be compatible.

tajo-storage/tajo-storage-hdfs/src/main/java/org/apache/tajo/storage/orc/ORCScanner.java
```diff
@@ -0,0 +1,328 @@
...
+    currentPosInBatch++;
+
+    return tuple;
+  }
+
+  // TODO: support more types
+  private Datum createValueDatum(Vector vector, TajoDataTypes.DataType type) {
+    switch (type.getType()) {
+      case INT1:
+      case UINT1:
```
hyunsik: As I mentioned above, unsigned types can be omitted because they are not supported.

tajo-storage/tajo-storage-hdfs/src/main/java/org/apache/tajo/storage/orc/ORCScanner.java
```diff
@@ -0,0 +1,328 @@
...
+        return DatumFactory.createText(((SliceVector) vector).vector[currentPosInBatch].getBytes());
+
+      case BLOB:
+        if (((SliceVector) vector).vector[currentPosInBatch] == null)
+          return NullDatum.get();
+
+        return DatumFactory.createBlob(((SliceVector) vector).vector[currentPosInBatch].getBytes());
+
+      case TIMESTAMP:
+        if (((LongVector) vector).isNull[currentPosInBatch])
```
hyunsik: You need to check how ORC stores timestamp values.

tajo-storage/tajo-storage-hdfs/src/main/java/org/apache/tajo/storage/orc/ORCScanner.java
```diff
@@ -0,0 +1,328 @@
...
+
+      case DATE:
+        if (((LongVector) vector).isNull[currentPosInBatch])
+          return NullDatum.get();
+
+        return DatumFactory.createDate(
+          (int) ((LongVector) vector).vector[currentPosInBatch] + DateTimeUtil.DAYS_FROM_JULIAN_TO_EPOCH);
+
+      case INET4:
+        if (((LongVector) vector).isNull[currentPosInBatch])
```
hyunsik: ORC may be not support INET4 type.

hyunsik:
I leave some trivial comments.

tajo-storage/tajo-storage-hdfs/src/main/java/org/apache/tajo/storage/orc/ORCScanner.java
eminency: It is Tajo type. It is stored as integer type in ORC file.
The code is just for converting.

tajo-storage/tajo-storage-hdfs/src/main/java/org/apache/tajo/storage/orc/ORCScanner.java
eminency: I see, I will.

tajo-storage/tajo-storage-hdfs/src/main/java/org/apache/tajo/storage/orc/ORCScanner.java
```diff
@@ -0,0 +1,323 @@
...
+        return new BooleanVector();
+
+      case BLOB:
+      case TEXT:
+      case CHAR:
+      case PROTOBUF:
+        return new SliceVector();
+
+      default:
+        throw new UnsupportedException("This data type is not supported currently: "+type.toString());
```
jihoonson: According to the recent changes in https://issues.apache.org/jira/browse/TAJO-1625, ```UmimplementedException``` looks more approprite.

tajo-storage/tajo-storage-hdfs/src/main/java/org/apache/tajo/storage/orc/ORCScanner.java
```diff
@@ -0,0 +1,323 @@
...
+    if (currentPosInBatch == batchSize) {
+      getNextBatch();
+
+      // EOF
+      if (batchSize == -1) {
+        return null;
+      }
+    }
+
+    Tuple tuple = new VTuple(targets.length);
```
jihoonson: After https://issues.apache.org/jira/browse/TAJO-1343, ```tuple``` should be singleton.

jihoonson:
@eminency, thanks for your work. Your patch looks good. I left some trivial comments.

I've also tested your patch on a real cluster. Here is an interesting result.

```
default> \d lineitem_orc_snappy

table name: default.lineitem_orc_snappy
store type: ORC
number of rows: unknown
volume: 23.6 GB
Options: 

schema: 
l_orderkey	INT8
l_partkey	INT8
l_suppkey	INT8
l_linenumber	INT8
l_quantity	FLOAT8
l_extendedprice	FLOAT8
l_discount	FLOAT8
l_tax	FLOAT8
l_returnflag	TEXT
l_linestatus	TEXT
l_shipdate	DATE
l_commitdate	DATE
l_receiptdate	DATE
l_shipinstruct	TEXT
l_shipmode	TEXT
l_comment	TEXT

default> select count(*) from lineitem_orc_snappy;
Progress: 100%, response time: 1.601 sec
?count
600037902
(1 rows, 1.601 sec, 10 B selected) 
```

jihoonson:
I missed +1.

jihoonson:
Great! I'll commit now.

eminency:
@jihoonson Thanks a lot :)

PR #580, opened 2015-05-21
TAJO-1558: HBASE_LIB/hbase-server-*.jar should be included in the CLASSPATH

eminency:
bin/tajo script is modified


blrunner:
+1

Thanks for your contribution.
It looks good to me.

PR #581, opened 2015-05-21
TAJO-1618 [Rest API] queries/{queryId} should set default print type

charsyam:
currently REST API /queries/
{queryId} cause error without print parameter.
and submit query just return url of /queries/{queryId}
without print type.
so I think it is better to set default print type when users don't pass print querystring.

### DIFF
--- a/tajo-core/src/main/java/org/apache/tajo/ws/rs/resources/QueryResource.java
+++ b/tajo-core/src/main/java/org/apache/tajo/ws/rs/resources/QueryResource.java
@@ -18,28 +18,6 @@
 
 package org.apache.tajo.ws.rs.resources;
 
-import java.net.URI;
-import java.util.ArrayList;
-import java.util.HashMap;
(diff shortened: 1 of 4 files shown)


jihoonson:
It seems that there were some problems to download some maven jars.
Would you trigger the test again?
It would be nice if you trigger the jenkins test by putting your patch on jira.

charsyam:
@jihoonson I attached patch on jira.
and it passed. :)

jihoonson:
+1. Thanks @charsyam. It works well. I'll change indent before commit.
Anyway, this patch requires to update our document for REST API (https://cwiki.apache.org/confluence/display/TAJO/TAJO+REST+API). Would you update it?

charsyam:
@jihoonson I will update it soon.

charsyam:
@jihoonson and I'm sorry for indent. Actually I fixed it , but I forgot push. sorry.

PR #582, opened 2015-05-23
TAJO-1620: random() in an SQL should generate RANDOM numbers

eminency:
One more exception case is added in ConstantFolding rule.


jihoonson:
+1. This patch does not contain any tests, but it is difficult to test random() function, I think.

PR #586, opened 2015-05-26
TAJO-1621: Compilation error with hadoop 2.7.0

jinossy:


jihoonson:
+1. Even though the latest test is failed, I've verified this patch with my laptop. In addition, every test failure is not related to the changes in this patch.

PR #590, opened 2015-05-28
TAJO-1627 Rest API should support this type of query "select 1"

charsyam:
Currently, Tajo Rest API supports well "select * from table" or "select count from table" <-- normal type
but, Tajo Rest API couldn't support this type of query like "select 1" <-- direct return type
to solve this, queries api should support resultset also.
so I suggest this.
queries return 201 Created when it is normal type of query
and query is direct query type and return 200 with result
like this (direct return type)
{"resultCode":"OK","schema":{"fields":[{"name":"?number","typeDesc":{"dataType":
{"type":"INT4"}
}}],"fieldsByQualifiedName":
{"?number":0}
,"fieldsByName":{"?number":[0]}},"serializedTupes":["AAAAAAE\u003d"]}

### DIFF
--- a/tajo-core/src/main/java/org/apache/tajo/ws/rs/resources/QueryResource.java
+++ b/tajo-core/src/main/java/org/apache/tajo/ws/rs/resources/QueryResource.java
@@ -18,10 +18,14 @@
 
 package org.apache.tajo.ws.rs.resources;
 
+import com.google.common.collect.Lists;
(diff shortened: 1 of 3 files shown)


hyunsik:
Hi @charsyam,

The proposal looks good to me. I have one suggestion. The JSON schema for Tajo schema looks different from those of other APIs. It would be great if they are consistent.

PR #591, opened 2015-05-28
TAJO-1629: Sleep() function does not work.

jihoonson:
[Review on Reviewable](https://reviewable.io/reviews/apache/tajo/591)


blrunner:
+1

It looks good to me.

jihoonson:
@blrunner thanks for your review, but I'll close this PR.
I think that we should find a better solution rather than this way because this patch will create a new thread per row and per sleep() function. 
Honestly, I forgot to withdraw this PR. Very sorry.

blrunner:
Okay, no problem. :)

jihoonson:
Thanks.

PR #592, opened 2015-05-28
TAJO-1630: Test failure after TAJO-1130

jihoonson:
I've changed only two lines as follows. Everything else is related to unit tests.
```
diff --git a/tajo-core/src/main/java/org/apache/tajo/engine/planner/global/MasterPlan.java b/tajo-core/src/main/java/org/apache/tajo/engine/planner/global/MasterPlan.java
index 22c3751..f8cd1e9 100644
--- a/tajo-core/src/main/java/org/apache/tajo/engine/planner/global/MasterPlan.java
+++ b/tajo-core/src/main/java/org/apache/tajo/engine/planner/global/MasterPlan.java
@@ -237,13 +237,13 @@ public class MasterPlan {
     sb.append("Order of Execution\n");
     sb.append("-------------------------------------------------------------------------------");
     int order = 1;
-    for (ExecutionBlock currentEB : cursor) {
+    for (ExecutionBlock currentEB : executionOrderCursor) {
       sb.append("\n").append(order).append(": ").append(currentEB.getId());
       order++;
     }
     sb.append("\n-------------------------------------------------------------------------------\n");
 
-    for (ExecutionBlock block : cursor) {
+    for (ExecutionBlock block : executionOrderCursor) {
 
       boolean terminal = false;
```

[Review on Reviewable](https://reviewable.io/reviews/apache/tajo/592)

### DIFF
--- a/tajo-core/src/main/java/org/apache/tajo/engine/planner/global/MasterPlan.java
+++ b/tajo-core/src/main/java/org/apache/tajo/engine/planner/global/MasterPlan.java
@@ -237,13 +237,13 @@ public class MasterPlan {
     sb.append("Order of Execution\n");
     sb.append("-------------------------------------------------------------------------------");
     int order = 1;
-    for (ExecutionBlock currentEB : cursor) {
+    for (ExecutionBlock currentEB : executionOrderCursor) {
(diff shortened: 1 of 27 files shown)


jinossy:
+1 ship it!!

PR #595, opened 2015-06-01
TAJO-1615: Implement TaskManager

jinossy:
TaskManager launches task to the TaskExecutor and It uses multiple threads equal to the number of cpu-cores. When a task is completed, TaskManager releases the allocated resource and NodeStatusUpdater sends node status with latest resource status to the ResourceTracker in TajoMaster


jinossy:
I'll update the wiki continuously.
https://cwiki.apache.org/confluence/display/TAJO/Tajo+Internal

jihoonson:
+1 looks good to me!
Before commit, it is better if you add a TODO comment for TaskId.getTaskId() to reduce confusion.

PR #465, opened 2015-03-26
TAJO-1450 Encapsulate Datum in Tuple

navis:
Not completed, but still worthy, imho.

PR #596, opened 2015-06-01
TAJO-1450: TAJO-1450 Encapsulate Datum in Tuple

jihoonson:
Just rebased patch of @navis' work in #465.
His patch looks good to me. If any other objections, I'll commit.

[Review on Reviewable](https://reviewable.io/reviews/apache/tajo/596)


tajo-common/src/main/java/org/apache/tajo/storage/NullTuple.java
```diff
@@ -117,6 +124,11 @@ public char getChar(int fieldId) {
   }
 
   @Override
+  public byte[] getTextBytes(int fieldId) {
+    return new byte[0];
```
jinossy: Can you change to static variable ?

tajo-common/src/main/java/org/apache/tajo/storage/VTuple.java
```diff
@@ -132,6 +140,11 @@ public char getChar(int fieldId) {
 	}
 
   @Override
+  public byte[] getTextBytes(int fieldId) {
+    return values[fieldId].asTextBytes();
+  }
+
+  @Override
 	public short getInt2(int fieldId) {
```
jinossy: please adjusting the indent

jinossy:
@jihoonson 
Is the changes not related the following error ?
```
Failed tests: 
  TestValidators.testMinValidator:193 
Expected: is <true>
     but: was <false>
```

tajo-common/src/main/java/org/apache/tajo/storage/NullTuple.java
jihoonson: Thanks. I've changed.

tajo-common/src/main/java/org/apache/tajo/storage/VTuple.java
jihoonson: Thanks. I've adjusted indent.

jihoonson:
@jinossy, thanks for your review. That error seems not related to this patch. It would be better to investigate in another issue, I think.

jinossy:
@jihoonson
I open the issue to TAJO-1635 and the patch looks good to me.
Here is my +1 
Thanks

jihoonson:
Thanks! I'll commit shortly.

PR #597, opened 2015-06-01
TAJO-1634 REST API: fix error when offset is zero

charsyam:
when offset is 0, /databases/{database-name}/queries/q000/results/0000
cause 500 error.


hyunsik:
+1
The patch looks good to me.

PR #599, opened 2015-06-05
TAJO-1638 Remove offset parameter from rest api result/{cacheId}

charsyam:
in result/{cacheId}
rest api.
offset parameter can make user confused.
because offset can't go back.
It's only set forward(not backward)
and add 3 new header for result response.
X-Tajo-Offset : from start offset
X-Tajo-Count: fetch count
X-Tajo-EOS: if there is end stream. it sets EOS true


blrunner:
@charsyam 

Thanks for your contribution.
But I failed to merge this patch as following

```
Auto-merging tajo-core/src/test/java/org/apache/tajo/ws/rs/resources/TestQueryResultResource.java
Auto-merging tajo-core/src/main/java/org/apache/tajo/ws/rs/resources/QueryResultResource.java
CONFLICT (content): Merge conflict in tajo-core/src/main/java/org/apache/tajo/ws/rs/resources/QueryResultResource.java
Automatic merge failed; fix conflicts and then commit the result.
```

Could you check it ?

charsyam:
@blrunner 
HI, Thanks. I rebased it :)

blrunner:
+1

PR #600, opened 2015-06-09
TAJO-1642: CatalogServer need to check meta table first.

blrunner:
I tested this patch on my laptop. When I tried to start up tajo-0.11.0-SNAPSHOT with the catalog-site.xml file for tajo-0.10.1-SNAPSHOT, TajoMaster printed following messages.

```
15/06/09 11:38:11 INFO store.MySQLStore: The meta table of CatalogServer already is created.
15/06/09 11:38:11 ERROR store.MySQLStore: Catalog version (2) and current driver version (4) are mismatch to each other
15/06/09 11:38:11 ERROR store.MySQLStore: =========================================================================
15/06/09 11:38:11 ERROR store.MySQLStore: | Catalog Store Migration Is Needed |
15/06/09 11:38:11 ERROR store.MySQLStore: =========================================================================
15/06/09 11:38:11 ERROR store.MySQLStore: | You might downgrade or upgrade Apache Tajo. Downgrading or upgrading |
15/06/09 11:38:11 ERROR store.MySQLStore: | Tajo without migration process is only available in some versions. |
15/06/09 11:38:11 ERROR store.MySQLStore: | In order to learn how to migration Apache Tajo instance, |
15/06/09 11:38:11 ERROR store.MySQLStore: | please refer http://s.apache.org/0_8_migration. |
15/06/09 11:38:11 ERROR store.MySQLStore: =========================================================================
```

And I found that existing column informations exists normally. For the reference, I succeed to test with only tajo-0.11.0-SNAPSHOT.

### DIFF
--- a/tajo-catalog/tajo-catalog-server/src/main/java/org/apache/tajo/catalog/store/XMLCatalogSchemaManager.java
+++ b/tajo-catalog/tajo-catalog-server/src/main/java/org/apache/tajo/catalog/store/XMLCatalogSchemaManager.java
@@ -55,6 +55,7 @@ import org.apache.tajo.catalog.CatalogConstants;
 import org.apache.tajo.catalog.CatalogUtil;
 import org.apache.tajo.catalog.exception.CatalogException;
 import org.apache.tajo.catalog.store.object.*;
+import org.apache.tajo.util.TUtil;
 
 public class XMLCatalogSchemaManager {
   protected final Log LOG = LogFactory.getLog(getClass());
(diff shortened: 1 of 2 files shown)


tajo-catalog/tajo-catalog-server/src/main/java/org/apache/tajo/catalog/store/XMLCatalogSchemaManager.java
```diff
@@ -344,6 +345,16 @@ public void upgradeBaseSchema(Connection conn, int currentVersion) {
     CatalogUtil.closeQuietly(stmt);
   }
 
+  public boolean catalogAlreadyExists(Connection conn) throws CatalogException {
+    boolean result = false;
+    try {
+      result = checkExistence(conn, DatabaseObjectType.TABLE, CatalogConstants.TB_META);
```
jihoonson: I think that every database object should be checked.

jihoonson:
@blrunner thanks for your patch.
I left one comment.

blrunner:
Hi @jihoonson 

Thanks for your review.

jihoonson:
+1 LGTM!

PR #601, opened 2015-06-15
TAJO-1644: When inserting empty data into a partitioned table, existing data would be removed.

blrunner:
When inserting empty data into a partitioned table, existing data would be removed. Tajo provides column value partition which is hive-style partition. In hive, when inserting empty data into a partition, there are two cases. If you use dynamic partitions, existing data never would be removed. But if you don't use dynamic partitions, existing data would be removed. When inserting a data to partition, tajo user don't specify each column and each column value. So, it is similar to dynamic partition of hive. It seems to update deletion logic of partitioned table.


tajo-core/src/test/java/org/apache/tajo/engine/query/TestTablePartitions.java
```diff
@@ -548,54 +526,61 @@ public final void testInsertIntoColumnPartitionedTableByThreeColumns() throws Ex
...
+    verifyMaintainExistingData(res, tableName);
 
-    ContentSummary summary = fs.getContentSummary(new Path(desc.getPath()));
+    executeString("DROP TABLE " + tableName + " PURGE").close();
+  }
 
-    assertEquals(summary.getDirectoryCount(), 1L);
-    assertEquals(summary.getFileCount(), 0L);
-    assertEquals(summary.getLength(), 0L);
+  private final void verifyMaintainExistingData(ResultSet res, String tableName) throws Exception {
```
hyunsik: It would be better if you rename verifyMaintainExistingData to verifyKeptExistingData.

tajo-storage/tajo-storage-hdfs/src/main/java/org/apache/tajo/storage/FileTablespace.java
```diff
@@ -892,9 +892,8 @@ protected Path commitOutputData(OverridableConf queryContext, boolean changeFile
           boolean movedToOldTable = false;
           boolean committed = false;
           Path oldTableDir = new Path(stagingDir, TajoConstants.INSERT_OVERWIRTE_OLD_TABLE_NAME);
-          ContentSummary summary = fs.getContentSummary(stagingResultDir);
 
-          if (!queryContext.get(QueryVars.OUTPUT_PARTITIONS, "").isEmpty() && summary.getFileCount() > 0L) {
+          if (!queryContext.get(QueryVars.OUTPUT_PARTITIONS, "").isEmpty()) {
```
hyunsik: It would be better if it works according to session variable to allow users to decide whether original data set is removed or not when inserting rows is empty.

blrunner:
Thanks @hyunsik 

I've just updated the patch using your comments. :)

hyunsik:
Hi @blrunner,

Thank you for your work. How about the insert overwrite behavior for not partitioned table? Does that the same semantic?

blrunner:
Hi @hyunsik 

Thank you for your review. I also agree with you. But when inserting empty data into a non partitioned table in hive, existing data would be removed always. I'm a bit anxious about user's confusion between hive and tajo.

tajo-common/src/main/java/org/apache/tajo/SessionVars.java
```diff
@@ -126,6 +126,9 @@
   NULL_CHAR(ConfVars.$TEXT_NULL, "null char of text file output", DEFAULT),
   CODEGEN(ConfVars.$CODEGEN, "Runtime code generation enabled (experiment)", DEFAULT),
 
+  TABLE_PARTITION_KEPT_EXISTING_DATA_ENABLED(ConfVars.$TABLE_PARTITION_KEPT_EXISTING_DATA_ENABLED,
```
hyunsik: How about {{TABLE_PARTITION_NO_RESULT_OVERWRITE_ENABLED}} for the config name? It seems to be more intuitive for me.

tajo-common/src/main/java/org/apache/tajo/SessionVars.java
```diff
@@ -126,6 +126,9 @@
...
+    "When inserting empty data into a partitioned table, set to keep existing data.", DEFAULT),
```
hyunsik: It would be better if the description is 'If True, a partitioned table is overwritten even if a sub query leads to no result. Otherwise, the table data will be kept if there is no result.'.

tajo-common/src/main/java/org/apache/tajo/conf/TajoConf.java
```diff
@@ -376,7 +375,10 @@ public static int setDateOrder(int dateOrder) {
     // Behavior Control ---------------------------------------------------------
     $BEHAVIOR_ARITHMETIC_ABORT("tajo.behavior.arithmetic-abort", false),
 
-    // ResultSet ---------------------------------------------------------
+    // When inserting empty data into a partitioned table, set to keep existing data.
```
hyunsik: This comment needs to be updated by my above comment.

tajo-common/src/main/java/org/apache/tajo/conf/TajoConf.java
```diff
@@ -376,7 +375,10 @@ public static int setDateOrder(int dateOrder) {
...
+    $TABLE_PARTITION_KEPT_EXISTING_DATA_ENABLED("tajo.table-partition.kept-existing-data.enabled", true),
```
hyunsik: This also should be updated by my above comment. Also, it would be nice if the config key is changed to 'tajo.partition.overwrite.even-if-no-result', and its default config should be ``false``.

tajo-storage/tajo-storage-hdfs/src/main/java/org/apache/tajo/storage/FileTablespace.java
```diff
@@ -894,7 +894,12 @@ protected Path commitOutputData(OverridableConf queryContext, boolean changeFile
           Path oldTableDir = new Path(stagingDir, TajoConstants.INSERT_OVERWIRTE_OLD_TABLE_NAME);
           ContentSummary summary = fs.getContentSummary(stagingResultDir);
 
-          if (!queryContext.get(QueryVars.OUTPUT_PARTITIONS, "").isEmpty() && summary.getFileCount() > 0L) {
+          // When inserting empty data into a partitioned table, check if keep existing data need to be remove or not.
+          boolean keptEnabled = queryContext.getBool(SessionVars.TABLE_PARTITION_KEPT_EXISTING_DATA_ENABLED);
+
+          // If existing data doesn't need to keep, check if there are some files.
+          if ( (!queryContext.get(QueryVars.OUTPUT_PARTITIONS, "").isEmpty())
```
hyunsik: This condition should be changed depending on the changed default value.

blrunner:
@hyunsik 

tajo-common/src/main/java/org/apache/tajo/conf/TajoConf.java
```diff
@@ -376,7 +375,11 @@ public static int setDateOrder(int dateOrder) {
     // Behavior Control ---------------------------------------------------------
     $BEHAVIOR_ARITHMETIC_ABORT("tajo.behavior.arithmetic-abort", false),
 
-    // ResultSet ---------------------------------------------------------
+    // If True, a partitioned table is overwritten even if a sub query leads to no result.
+    // Otherwise, the table data will be kept if there is no result
+    $TABLE_PARTITION_NO_RESULT_OVERWRITE_ENABLED("tajo.partition.overwrite.even-if-no-result", false),
```
hyunsik: I have one more suggestion. It would be great if you use PARTITION instead of TABLE_PARTITION. It is still clear because we only use the term 'partition' to indicate table partition.

tajo-common/src/main/java/org/apache/tajo/SessionVars.java
```diff
@@ -126,6 +126,10 @@
   NULL_CHAR(ConfVars.$TEXT_NULL, "null char of text file output", DEFAULT),
   CODEGEN(ConfVars.$CODEGEN, "Runtime code generation enabled (experiment)", DEFAULT),
 
+  TABLE_PARTITION_NO_RESULT_OVERWRITE_ENABLED(ConfVars.$TABLE_PARTITION_NO_RESULT_OVERWRITE_ENABLED,
```
hyunsik: I have one more suggestion. It would be great if you use PARTITION instead of TABLE_PARTITION. It is still clear because we only use the term 'partition' to indicate table partition.

hyunsik:
+1
Looks good to me. I leaved one trivial comment. If you agree, you can commit the patch after you reflect my comment.

PR #602, opened 2015-06-16
TAJO-1616: Implement TablespaceManager to load Tablespaces.

hyunsik:
[Review on Reviewable](https://reviewable.io/reviews/apache/tajo/602)


jihoonson:
@hyunsik thanks for the nice patch.
I'm reviewing, but it may take some time. I'll finish as soon as possible.
Anyway, the test failure is not related this patch. This is very annoying, so I'll investigate this problem at https://issues.apache.org/jira/browse/TAJO-1596. Would you please trigger the test again?

tajo-storage/tajo-storage-common/src/main/java/org/apache/tajo/storage/Tablespace.java
```diff
@@ -69,24 +77,47 @@ public Tablespace(String storeType) {
...
-   * which is the option to delete all the data.
+   * Get Table URI
    *
-   * @param tableDesc
-   * @throws java.io.IOException
+   * @param tableName
+   * @return
    */
-  public abstract void purgeTable(TableDesc tableDesc) throws IOException;
+  public abstract URI getTableUri(String databaseName, String tableName);
```
jihoonson: This API seems to be duplicated with TableSpaceManager.getTableUri(), even though this has smaller functionality. Would you please explain any differences and proper usages?

tajo-storage/tajo-storage-common/src/main/java/org/apache/tajo/storage/TableSpaceManager.java
```diff
@@ -19,236 +19,372 @@
...
-   * Clear all class cache
+   * Singleton instance
    */
-  @VisibleForTesting
-  protected synchronized static void clearCache() {
-    CONSTRUCTOR_CACHE.clear();
-    SCANNER_HANDLER_CACHE.clear();
-    APPENDER_HANDLER_CACHE.clear();
-    storageManagers.clear();
+  private static final TableSpaceManager instance;
```
jihoonson: TablespaceManager looks to be more appropriate.

tajo-docs/src/main/sphinx/storage_plugin.rst
```diff
@@ -0,0 +1,47 @@
...
+
+Overview
+========
+
+Tajo supports various storage systems, such as HDFS, Amazon S3, Openstack Swift, and HBase. Also, we have a plan to support RDBMS storages like Oracle, MySQL, PostgreSQL. Tajo already embeds HDFS, S3, Openstack, and HBase, and also Tajo allows users to register custom storages and data formats to Tajo cluster instances. This section describes how you register custom storages and data types.
+
+Register custom storage
+=======================
+
+First of all, your storage implementation should be packed as a jar file. Then, please copy the jar file into ``tajo/lib`` directory. Next, you should copy ``conf/storage-site.json.template`` into ``conf/storage-site.json`` and modify the file like the below.
```
jihoonson: Maybe ```tajo/extlib``` rather than ```tajo/lib```?

tajo-storage/tajo-storage-common/src/main/java/org/apache/tajo/storage/TableSpaceManager.java
```diff
@@ -19,236 +19,372 @@
...
+import java.util.UUID;
+
+import static org.apache.tajo.storage.StorageConstants.LOCAL_FS_URI;
 
 /**
  * It handles available table spaces and cache TableSpace instances.
+ *
+ * Default tablespace must be a filesystem-based one.
+ * HDFS and S3 can be a default tablespace if a Tajo cluster is in fully distributed mode.
+ * Local file system can be a default tablespace if a Tajo cluster runs on a single machine.
```
jihoonson: This explanation still looks good, but it will be nice if you list properties that the storage of default tablespace must have.

tajo-storage/tajo-storage-common/src/main/java/org/apache/tajo/storage/Tablespace.java
```diff
@@ -49,18 +51,24 @@
...
 
-  public static final PathFilter hiddenFileFilter = new PathFilter() {
-    public boolean accept(Path p) {
-      String name = p.getName();
-      return !name.startsWith("_") && !name.startsWith(".");
-    }
-  };
+  protected final String name;
+  protected final URI uri;
+  /** this space is visible or not. */
```
jihoonson: Would you explain what visible means?

tajo-storage/tajo-storage-common/src/test/java/org/apache/tajo/storage/TestTableSpaceManager.java
```diff
@@ -0,0 +1,35 @@
...
+import org.junit.Test;
+
+import java.io.IOException;
+
+import static org.junit.Assert.*;
+
+public class TestTableSpaceManager {
+  @Test
+  public void testConfigLoad() throws IOException {
+    //String json = FileUtil.readTextFileFromResource("storage-test1.json");
```
jihoonson: This test looks incomplete.

jihoonson:
@hyunsik thanks for nice work. I left some trivial comments. Please consider them.

tajo-storage/tajo-storage-common/src/main/java/org/apache/tajo/storage/Tablespace.java
hyunsik: Later, TablespaceManager will have additional APIs to wrap Tablespace APIs. This is because getting a tablespace and calling an API of tablespace are a frequently-used call routine. I'll add more utility static methods like them to TablespaceManager.

tajo-storage/tajo-storage-common/src/main/java/org/apache/tajo/storage/TableSpaceManager.java
hyunsik: I haven't implemented format handlers (i.e., scanner and appender) in TablespaceManager yet. This patch still keeps SCANNER_HANDLER and APPENDER_HANDLER in OldStorageManager, which is old one. In the following issues, I'll implement format handlers and then I'll remove OldStorageManager.

tajo-docs/src/main/sphinx/storage_plugin.rst
hyunsik: I've just committed ``extlib``. I'll change it. Thanks!

tajo-storage/tajo-storage-common/src/main/java/org/apache/tajo/storage/Tablespace.java
hyunsik: TablespaceManager automatically adds temporary tablespaces like ``file:///``. These temporary tables are only used internally in Tajo, and they must not not visible. The visible member variable is for that.

tajo-storage/tajo-storage-common/src/test/java/org/apache/tajo/storage/TestTableSpaceManager.java
hyunsik: I'll remove the test. In the following issues, I'll add some tests.

tajo-storage/tajo-storage-common/src/main/java/org/apache/tajo/storage/TableSpaceManager.java
hyunsik: Currently, some part still is hard coded for default tablespace. I'm not sure what happen if we use other storage as a default storage. I think that we need more tests to allow other storages to be used for default storage.

hyunsik:
Thank you for your review. I reflected the comments.

tajo-storage/tajo-storage-common/src/main/java/org/apache/tajo/storage/Tablespace.java
jihoonson: Got it. Thanks.

tajo-storage/tajo-storage-common/src/main/java/org/apache/tajo/storage/TableSpaceManager.java
jihoonson: I agree.

tajo-storage/tajo-storage-common/src/main/java/org/apache/tajo/storage/Tablespace.java
jihoonson: Got it. Thanks.

jihoonson:
Thanks @hyunsik, but it seems that there were some problems during test on Travis.
Please trigger again.

hyunsik:
I've updated the patch, and I also improved the feature of insert row values.

hyunsik:
I found some problem from unit tests through Travis.

``
java.net.ConnectException: Call From testing-worker-linux-f25aadab-1-20030-linux-7-67640030/127.0.0.1 to localhost:50374 failed on connection exception: 
``

This error message means one of three cases:
 * localhost is not 127.0.0.1
 * port 50374 is denied by iptables
 * server is already closed.

I'll dig it more and I'll upload the new patch soon.

hyunsik:
I found the problem. TablespaceManager is designed as a singleton class in a cluster, but some unit tests starts additional TajoTestingCluster. It overwrites the tablespace configurations of TablespaceManager. It results in test failures. I'll try to change them to share the one cluster.

jihoonson:
Looks nice. Please ping me when you are ready.

hyunsik:
I changed travis script to perform unit tests on jdk 7 as well as jdk 8. I also increased ulimit for thread num.

hyunsik:
@jihoonson  It seems to be ready.

hyunsik:
rebased.

tajo-core/src/main/java/org/apache/tajo/engine/planner/physical/InsertRowExec.java
```diff
@@ -0,0 +1,107 @@
...
+import org.apache.tajo.storage.TablespaceManager;
+import org.apache.tajo.storage.Tuple;
+import org.apache.tajo.worker.TaskAttemptContext;
+
+import java.io.IOException;
+
+/**
+ * This is a physical executor to store a row immediately
+ */
+public class InsertRowExec extends UnaryPhysicalExec {
```
jihoonson: Would you please explain any differences from StoreTableExec except that StoreTableExec supports the multi-file write?

tajo-core/src/main/java/org/apache/tajo/master/exec/QueryExecutor.java
```diff
@@ -436,14 +476,17 @@ public void executeDistributedQuery(QueryContext queryContext, Session session,
...
-      StorageProperty storageProperty = sm.getStorageProperty();
-      if (!storageProperty.isSupportsInsertInto()) {
+    TableDesc tableDesc = PlannerUtil.getTableDesc(catalog, plan.getRootBlock().getRoot());
+    if (tableDesc != null) {
+
+      Tablespace space = TablespaceManager.get(tableDesc.getUri()).get();
+      FormatProperty formatProperty = space.getFormatProperty(tableDesc.getMeta());
+
+      if (!formatProperty.isInsertable()) {
         throw new VerifyException("Inserting into non-file storage is not supported.");
```
jihoonson: It would be better if the storage name is printed with this exception.

tajo-core/src/main/java/org/apache/tajo/querymaster/QueryMasterTask.java
```diff
@@ -307,43 +291,32 @@ private static boolean isTerminatedState(QueryState state) {
         state == QueryState.QUERY_ERROR;
   }
 
+  private LogicalPlan plan;
+
   public synchronized void startQuery() {
-    Tablespace sm = null;
-    LogicalPlan plan = null;
+    //LogicalPlan plan = null;
```
jihoonson: Please remove this line.

tajo-core/src/main/java/org/apache/tajo/querymaster/Repartitioner.java
```diff
@@ -105,25 +99,29 @@ public static void scheduleFragmentsForJoinQuery(TaskSchedulerContext schedulerC
           ExecutionBlockId scanEBId = TajoIdUtils.createExecutionBlockId(scans[i].getTableName());
           stats[i] = masterContext.getStage(scanEBId).getResultStats().getNumBytes();
         }
-        fragments[i] = new FileFragment(scans[i].getCanonicalName(), tablePath, 0, 0, new String[]{UNKNOWN_HOST});
+
+        // TODO - We should remove dummy flagment usages
```
jihoonson: Please fix the typo. 
flagment -> fragment

tajo-core/src/main/java/org/apache/tajo/querymaster/Repartitioner.java
```diff
@@ -648,10 +644,15 @@ public static void scheduleRangeShuffledFetches(TaskSchedulerContext schedulerCo
         throw new IOException("Can't get table meta data from catalog: " +
             PlannerUtil.getStoreTableName(masterPlan.getLogicalPlan()));
       }
-      ranges = TableSpaceManager.getStorageManager(stage.getContext().getConf(), storeType)
-          .getInsertSortRanges(stage.getContext().getQueryContext(), tableDesc,
-              sortNode.getInSchema(), sortSpecs,
-              mergedRange);
+
+      Tablespace space = TablespaceManager.getAnyByScheme(storeType).get();
+      ranges = space.getInsertSortRanges(
```
jihoonson: ```space``` may be checked whether it is absent or not.

jihoonson:
I left some more trivial comments and questions.

tajo-core/src/main/java/org/apache/tajo/engine/planner/physical/InsertRowExec.java
hyunsik: Currently, StoreTableExec cannot handle OLTP-like storages. We should modify StoreTableExec::openNewFile for them. The main objective of InsertRowExec is to avoid StoreTableExec to be multiple purposed. If we use StoreTableExec here, StoreTableExec would be more messy to having more exceptional and intuitive conditions in StoreTableExec::openNewFile().

hyunsik: In additional, InsertRowExec will handle multiple rows because INSERT INTO VALUES statement allows multiple rows. So, I'll change the name for the purpose.

tajo-core/src/main/java/org/apache/tajo/master/exec/QueryExecutor.java
hyunsik: Fixed.

tajo-core/src/main/java/org/apache/tajo/querymaster/Repartitioner.java
hyunsik: In original implementation, it guarantees that TablespaceManager.getAnyByScheme(storeType) will always return the proper tablespace handler if ``sortNode.getSortPurpose()`` is ``STORAGE_SPECIFIED``.

But, its implementation is not intuitive. We should improve them. But, it requires lots of extra changes. I'll continue to improve storage layer APIs. It will be improved later too.

hyunsik:
I reflected the comments and updated the patch. Thank you for your review.

jihoonson:
Thanks for the nice patch and fast update. 
+1 Ship it!

PR #604, opened 2015-06-16
TAJO-1647: Change Create Database Tajo Rest API address from POST /databases to POST /databases/{database_name}

charsyam:
in Rest API
POST /databases/{database_name} is more understandable
comparing to POST /databases with database_name parameter


jihoonson:
Hi @charsyam, thanks for your patch.
Since I'm not familiar with RESTful API, I would like to ask a question.
Given a create database statement, the ```database_name``` is a variable defined by users.
For such variables, IMHO, passing ```database_name``` as a parameter looks more intuitive. 
Are there any conventions in designing RESTful API?

charsyam:
@jihoonson Hi, jihoon. Thanks for your review :)
You're right. It is not real REST style.
but in my case, POST /databases has only one parameter as its name.
so I think POST /databases/{database-name} is more simple and understable. :)

But, It is not REST API.
so. I don't matter if you close this issue :)

jihoonson:
Thanks for your comment.
In terms of the consistency of APIs, I think that ```create database``` API should have the equivalent form of ```create table``` API.
So, passing ```database_name``` as a parameter looks good to me.

PR #459, opened 2015-03-24
TAJO-751 JDBC driver should support cancel() method

navis:
Not yet added test cases

PR #605, opened 2015-06-16
TAJO-751: JDBC driver should support cancel() method.

jihoonson:
This is the rebased patch of navis' work (#459).
While merging the patch into master, I've found some conflict codes and thus removed them.
This patch looks good to me. Here is my +1, and I'll commit if there aren't any objections.

[Review on Reviewable](https://reviewable.io/reviews/apache/tajo/605)


jihoonson:
Anyway, I think that it would be great if this patch is included in 0.10.2.

blrunner:
+1

Thanks @jihoonson  and @navis 
This patch looks good overall.

tajo-jdbc/src/main/java/org/apache/tajo/jdbc/TajoStatement.java
```diff
@@ -319,7 +374,9 @@ public void setCursorName(String name) throws SQLException {
    * Not necessary.
    */
   @Override
-  public void setEscapeProcessing(boolean enable) throws SQLException {}
+  public void setEscapeProcessing(boolean enable) throws SQLException {
```
hyunsik: Throwing SQLFeatureNotSupportedException may cause JDBC applications to not work. It would be better to print out a warning message.

jihoonson: Done.

hyunsik:
+1

PR #610, opened 2015-06-23
TAJO-1657: Tajo Rest API /database/{database-name]/tables should return table names only without invalid external table info

charsyam:
Tajo Rest API GET /databases/{database-name}/tables should return table name only without invalid external table info.
current tables API return table name and external table info but, it always return false.


blrunner:
Could you trigger this for the travis CI build?

PR #612, opened 2015-06-24
TAJO-1659: Simplify scan iteration in SeqScan.

hyunsik:
[Review on Reviewable](https://reviewable.io/reviews/apache/tajo/612)

### DIFF
--- a/NOTICE
+++ b/NOTICE
@@ -1,5 +1,5 @@
 Apache Tajo
-Copyright 2014 The Apache Software Foundation
+Copyright 2013-2015 The Apache Software Foundation
 
 This product includes software developed at
 The Apache Software Foundation (http://www.apache.org/).


tajo-core/src/main/java/org/apache/tajo/engine/planner/physical/SeqScanExec.java
```diff
@@ -228,26 +237,15 @@ public Tuple next() throws IOException {
...
-    } else {
-      while ((tuple = scanner.next()) != null) {
-        if (qual.eval(tuple).isTrue()) {
-          projector.eval(tuple, outTuple);
-          return outTuple;
-        }
-      }
-      return null;
+    while(scanIt.hasNext()) {
+      Tuple outTuple = new VTuple(outColumnNum);
```
jihoonson: outTuple don't have to be created for every input row.

tajo-storage/tajo-storage-hdfs/src/main/java/org/apache/tajo/storage/HashShuffleAppenderManager.java
```diff
@@ -30,9 +30,11 @@
 import org.apache.tajo.catalog.TableMeta;
 import org.apache.tajo.conf.TajoConf;
 import org.apache.tajo.conf.TajoConf.ConfVars;
+import org.apache.tajo.util.Pair;
```
jihoonson: Please remove unused imports.

jihoonson:
+1. I left trivial comments. Please consider them before commit.

tajo-core/src/main/java/org/apache/tajo/engine/planner/physical/SeqScanExec.java
hyunsik: It is actually for every row. While condition is just for check if there is next row, and every while loop will be just returned for each tuple.

hyunsik:
After merging another patch, it is causing unit test failures. I'll check them and then I'll commit.

PR #613, opened 2015-06-25
TAJO-1658: Filter push down to underlying storages

hyunsik:
This patch includes TAJO-1659.

[Review on Reviewable](https://reviewable.io/reviews/apache/tajo/613)


hyunsik:
This patch also includes some cleanup for table properties.

tajo-core/src/main/java/org/apache/tajo/engine/planner/physical/EmptyScanIterator.java
```diff
@@ -0,0 +1,40 @@
...
+public class EmptyScanIterator implements ScanIterator {
+
+  @Override
+  public boolean hasNext() throws IOException {
+    return false;
+  }
+
+  @Override
+  public Tuple next() throws IOException {
+    throw new IOException(this.getClass().getSimpleName() + "::next() is invoked.");
```
jihoonson: I'm wondering that throwing an exception is better than returning a null.

tajo-core/src/main/java/org/apache/tajo/engine/planner/physical/SeqScanExec.java
```diff
@@ -167,23 +162,55 @@ public void init() throws IOException {
...
-      if (scanner.isProjectable()) {
-        qual.bind(context.getEvalContext(), projected);
-      } else {
-        qual.bind(context.getEvalContext(), inSchema);
-      }
+    return projected;
+  }
 
+  private void initScanIterator() {
+    // We should use FilterScanIterator only if underlying storage does not support filter push down.
```
jihoonson: We currently have two kinds of scan operator, sequential scan (```SeqScanExec``` and ```PartitionMergeScanExec```) and index scan (```BSTIndexScanExec```). 
In the case of when an index exists for a HDFS table, index scan will do the same operation of ```FullScanIterator```. So, there seems an unified interface for index scan and ```FullScanIterator```.

Maybe, we design a new scanner for HDFS like ```IndexedScanner```. If so, we can also handle index scan in the below code block.
What do you think? Do you have any plans for this?

tajo-storage/tajo-storage-common/src/main/java/org/apache/tajo/storage/AbstractScanner.java
```diff
@@ -55,7 +56,7 @@ public boolean isSelectable() {
   }
 
   @Override
-  public void setSearchCondition(Object expr) {
+  public void setFilter(EvalNode filter) {
```
jihoonson: It looks that an exception should be thrown.

tajo-storage/tajo-storage-common/src/main/java/org/apache/tajo/storage/MergeScanner.java
```diff
@@ -165,7 +167,7 @@ public boolean isSelectable() {
   }
 
   @Override
-  public void setSearchCondition(Object expr) {
+  public void setFilter(EvalNode filter) {
```
jihoonson: It looks that an exception should be thrown.

jihoonson:
@hyunsik thanks for your patch. It looks good to me.
I left some comments. Please consider them.

tajo-core/src/main/java/org/apache/tajo/engine/planner/physical/EmptyScanIterator.java
hyunsik: I think that invoking next() is illegal state because it means that checking hasNext() is omitted.

tajo-storage/tajo-storage-common/src/main/java/org/apache/tajo/storage/MergeScanner.java
hyunsik: I'll fix it for current implementation. However, we need to know its ambiguity, and the solution should be depend on the final purpose of MergeScanner.

According to its implementation, MergeScanner is chosen if the number of ``fragments`` is more than one.

If MergeScanner is only for intermediate shuffled data, this method should throw an exception. Otherwise, we should implement to push down a filter to its scannner elements.

tajo-storage/tajo-storage-common/src/main/java/org/apache/tajo/storage/AbstractScanner.java
hyunsik: Fixed

tajo-core/src/main/java/org/apache/tajo/engine/planner/physical/SeqScanExec.java
hyunsik: You seem to want to use an unified interface for both index scanner and this work. I'm going to add the storage-related method to determine sub filter expressions which can be push down into a specific storage. Index scanner will be affected by this method in the same manner.

hyunsik:
@jihoonson Thank you for your comment. I've updated the patch.

jihoonson:
Thanks for rapid reply.
+1!

PR #615, opened 2015-06-25
TAJO-1486: Tajo should be able to skip header and footer rows when creating external table

eminency:
Two table meta options are added .

> text.skip.headerlines
> text.skip.footerlines

This feature is only for delimited text files.

[Review on Reviewable](https://reviewable.io/reviews/apache/tajo/615)


blrunner:
@eminency 

Thanks for your contribution.
If there is a unit test case for a plain text file, it would be more better.

eminency:
@blrunner 
I see, I will try to add it.

tajo-storage/tajo-storage-hdfs/src/main/java/org/apache/tajo/storage/text/ByteBufLineReader.java
```diff
@@ -197,6 +197,6 @@ public ByteBuf readLineBuf(AtomicInteger reads) throws IOException {
       }
     }
     reads.set(readBytes);
-    return buffer.slice(startIndex, readBytes - newlineLength);
+    return buffer.slice(startIndex, readBytes - newlineLength).retain();
```
jinossy: This buffer is shared until closing the ByteBufLineReader. if you want to   keep the sliced buffer, you must copy to new buffer

jinossy:
Guys,
In my opinion, skipping the footer would be good after block interation is added

jinossy:
+1 Looks great to me
I will move the headerLineNum to local variable before commit it

eminency:
@jinossy It looks good

PR #621, opened 2015-07-02
TAJO-1670: Refactor client errors and exceptions. 

hyunsik:
[Review on Reviewable](https://reviewable.io/reviews/apache/tajo/621)


hyunsik:
This is still ongoing work. I'll do more tests and make some documentation about exception policy.

blrunner:
Thanks you for your great work.
Could you rebase this patch against the master branch?
I failed to merge it with following messages.

{code:xml}
$ git pull https://github.com/hyunsik/tajo TAJO-1670
remote: Counting objects: 1050, done.
remote: Total 1050 (delta 392), reused 392 (delta 392), pack-reused 657
Receiving objects: 100% (1050/1050), 195.43 KiB | 189.00 KiB/s, done.
Resolving deltas: 100% (490/490), completed with 153 local objects.
From https://github.com/hyunsik/tajo
 * branch            TAJO-1670  -> FETCH_HEAD
Removing tajo-plan/src/main/java/org/apache/tajo/plan/logical/NoSuchColumnException.java
Auto-merging tajo-core/src/test/java/org/apache/tajo/ws/rs/resources/TestQueryResultResource.java
Auto-merging tajo-core/src/test/java/org/apache/tajo/ws/rs/resources/TestQueryResource.java
Auto-merging tajo-core/src/main/java/org/apache/tajo/ws/rs/resources/QueryResource.java
CONFLICT (content): Merge conflict in tajo-core/src/main/java/org/apache/tajo/ws/rs/resources/QueryResource.java
Auto-merging tajo-common/src/main/java/org/apache/tajo/exception/TajoRuntimeException.java
Auto-merging tajo-common/src/main/java/org/apache/tajo/exception/TajoInternalError.java
Auto-merging tajo-common/src/main/java/org/apache/tajo/exception/TajoExceptionInterface.java
Auto-merging tajo-common/src/main/java/org/apache/tajo/exception/TajoError.java
Auto-merging tajo-common/src/main/java/org/apache/tajo/exception/InvalidAddressException.java
Auto-merging tajo-catalog/tajo-catalog-common/src/main/java/org/apache/tajo/catalog/exception/UndefinedTablespaceException.java
Auto-merging tajo-catalog/tajo-catalog-common/src/main/java/org/apache/tajo/catalog/exception/UndefinedTableException.java
Auto-merging tajo-catalog/tajo-catalog-common/src/main/java/org/apache/tajo/catalog/exception/UndefinedPartitionException.java
Auto-merging tajo-catalog/tajo-catalog-common/src/main/java/org/apache/tajo/catalog/exception/UndefinedIndexException.java
Auto-merging tajo-catalog/tajo-catalog-common/src/main/java/org/apache/tajo/catalog/exception/UndefinedFunctionException.java
Auto-merging tajo-catalog/tajo-catalog-common/src/main/java/org/apache/tajo/catalog/exception/UndefinedDatabaseException.java
Removing tajo-catalog/tajo-catalog-common/src/main/java/org/apache/tajo/catalog/exception/NoSuchPartitionException.java
Removing tajo-catalog/tajo-catalog-common/src/main/java/org/apache/tajo/catalog/exception/NoSuchColumnException.java
Auto-merging tajo-catalog/tajo-catalog-common/src/main/java/org/apache/tajo/catalog/exception/DuplicateTablespaceException.java
Auto-merging tajo-catalog/tajo-catalog-common/src/main/java/org/apache/tajo/catalog/exception/DuplicateFunctionException.java
Auto-merging tajo-catalog/tajo-catalog-common/src/main/java/org/apache/tajo/catalog/exception/DuplicateDatabaseException.java
Auto-merging tajo-catalog/tajo-catalog-common/src/main/java/org/apache/tajo/catalog/exception/DuplicateColumnException.java
Removing tajo-catalog/tajo-catalog-common/src/main/java/org/apache/tajo/catalog/exception/AlreadyExistsTableException.java
Automatic merge failed; fix conflicts and then commit the result.
{code}

hyunsik:
I've updated the patch. I've written the wiki page named 'RPC Protocol Design Guide'. 
https://cwiki.apache.org/confluence/display/TAJO/RPC+Protocol+Design+Guide

This change is based on this guide. So, it may be easy to understand the patch. Although, I'll make mode documentations about the exception hierarchy and propagation feel free to ask me if you need more documentation for this work.

Also, this work is a part of TAJO-1625. The error code and some APIs will be evolved while I'm doing other issues in TAJO-1625.

tajo-catalog/tajo-catalog-client/src/main/java/org/apache/tajo/catalog/AbstractCatalogClient.java
```diff
@@ -51,113 +55,132 @@ public AbstractCatalogClient(TajoConf conf) {
...
-      builder.setTablespaceName(tablespaceName);
-      builder.setTablespaceUri(tablespaceUri);
-      return stub.createTablespace(null, builder.build()).getValue();
-    } catch (Exception e) {
-      LOG.error(e.getMessage(), e);
-      return Boolean.FALSE;
+      return isSuccess(stub.createTablespace(null, request));
+
+    } catch (ServiceException e) {
+      throw new RuntimeException(e);
```
jinossy: What do you think to add a ClientRuntimeException or TajoRuntimeException ?

hyunsik: I used RuntimeException for unexpected error. In my design, the server API call never throws ServiceException, only communication or netty will throw ServiceException. This case will occur rarely, so I used RuntimeException.

In addition, catalog and client API will be improved further to have API specific exceptions. For example, createTablespace will throw DuplicateTablespaceException and dropTable will throw UndefinedTableException. But, this approach significantly changes lots of parts. So, I'll do it in following works.

jinossy: I understand. Thank for your description

hyunsik:
Also, I'm writing a documentation for error propagation and exception usage guide in Tajo.
https://cwiki.apache.org/confluence/display/TAJO/Exception+Propagation

I hope that it would be helpful for you guys to understand my changes.

jihoonson:
Thanks! It would be much helpful. 
I'll review today

tajo-catalog/tajo-catalog-client/src/main/java/org/apache/tajo/catalog/AbstractCatalogClient.java
hyunsik: Thank you for your understanding.

tajo-catalog/tajo-catalog-client/src/main/java/org/apache/tajo/catalog/AbstractCatalogClient.java
```diff
@@ -574,25 +652,26 @@ public final FunctionDesc getFunction(final String signature, FunctionType funcT
...
+      response = stub.getFunctionMeta(null, builder.build());
+    } catch (ServiceException se) {
+      throw new RuntimeException(se);
     }
 
-    if (descProto == null) {
-      throw new NoSuchFunctionException(signature, paramTypes);
+    if (isThisError(response.getState(), ResultCode.UNDEFINED_FUNCTION)) {
+      throw new UndefinedFunctionException(signature, paramTypes);
+    } else if (isThisError(response.getState(), ResultCode.AMBIGUOUS_FUNCTION)) {
```
jihoonson: ResultCode seems to be DUPLICATE_FUNCTION

tajo-catalog/tajo-catalog-client/src/main/proto/CatalogProtocol.proto
```diff
@@ -24,53 +24,128 @@ option java_generate_equals_and_hash = true;
 import "CatalogProtos.proto";
 import "PrimitiveProtos.proto";
 
+message GetTablespacesResponse {
```
jihoonson: I think that this name is easily confused with GetTablespaceResponse. How about renaming it to ```GetTablespaceListResponse```?

tajo-catalog/tajo-catalog-client/src/main/proto/CatalogProtocol.proto
```diff
@@ -24,53 +24,128 @@ option java_generate_equals_and_hash = true;
...
+  required ReturnState state = 1;
+  optional TablespaceProto tablespace = 2;
+}
+
+message GetDatabasesResponse {
+  required ReturnState state = 1;
+  repeated DatabaseProto database = 2;
+}
+
+message TableDescResponse {
```
jihoonson: Should have the ```Get``` prefix.

tajo-catalog/tajo-catalog-client/src/main/proto/CatalogProtocol.proto
```diff
@@ -24,53 +24,128 @@ option java_generate_equals_and_hash = true;
...
+  required ReturnState state = 1;
+  optional TableDescProto table = 2;
+}
+
+message GetTablesResponse {
+  required ReturnState state = 1;
+  repeated TableDescriptorProto table = 2;
+}
+
+message FunctionDescResponse {
```
jihoonson: Should have the ```Get``` prefix.

tajo-catalog/tajo-catalog-common/src/main/java/org/apache/tajo/catalog/exception/TMCConnectionException.java
```diff
@@ -0,0 +1,32 @@
...
+
+package org.apache.tajo.catalog.exception;
+
+import org.apache.tajo.error.Errors.ResultCode;
+import org.apache.tajo.exception.TajoError;
+
+/**
+ * Tajo Metadata Connector's connection error
+ */
+public class TMCConnectionException extends TajoError {
```
jihoonson: IMHO, it is better if the class name is intuitive even though its length is quite long. 
In addition, ```TM``` is usually used for ```TajoMaster``` in our architecture. I think that this name makes a lot of confusion.

tajo-catalog/tajo-catalog-drivers/tajo-hive/src/main/java/org/apache/tajo/catalog/store/HiveCatalogUtil.java
```diff
@@ -95,7 +99,7 @@ public static String getHiveFieldType(TajoDataTypes.DataType dataType) {
     case DATE: return serdeConstants.DATE_TYPE_NAME;
     case TIMESTAMP: return serdeConstants.TIMESTAMP_TYPE_NAME;
     default:
-      throw new CatalogException(dataType + " is not supported.");
+      throw ExceptionUtil.makeInvalidDataType(dataType);
```
jihoonson: Please shrink this to just ```makeInvalidDataType()```.

tajo-catalog/tajo-catalog-common/src/main/java/org/apache/tajo/catalog/exception/CatalogExceptionUtil.java
```diff
@@ -0,0 +1,63 @@
...
+package org.apache.tajo.catalog.exception;
+
+import org.apache.tajo.common.TajoDataTypes;
+import org.apache.tajo.error.Errors;
+import org.apache.tajo.error.Errors.ResultCode;
+import org.apache.tajo.function.FunctionUtil;
+
+import java.util.Collection;
+
+public class CatalogExceptionUtil {
```
jihoonson: I think that it would be better if all methods in this class have "Exception" as their the postfix.

tajo-client/src/main/java/org/apache/tajo/client/SessionConnection.java
```diff
@@ -98,33 +106,49 @@ public SessionConnection(@NotNull ServiceTracker tracker, @Nullable String baseD
...
+
+      } catch (Throwable t) {
+        throw SQLExceptionUtil.makeUnableToEstablishConnection(t);
       }
+
       return client;
     }
   }
 
+  protected BlockingInterface getTMStub() throws SQLException {
```
jihoonson: As I mentioned above, I think abbreviation is not intuitive because it can be understood in various ways.

tajo-common/src/main/java/org/apache/tajo/exception/TajoException.java
```diff
@@ -0,0 +1,40 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.tajo.exception;
+
+import org.apache.tajo.error.Errors.ResultCode;
+
+public class TajoException extends Exception implements TajoExceptionInterface {
```
jihoonson: Please add a brief explanation.

tajo-common/src/main/proto/errors.proto
```diff
@@ -0,0 +1,327 @@
...
+  WARNING                       = 100; // Warning
+
+  // General Errors
+  INTERNAL_ERROR                = 201; // Error caused by internal bugs (See also TajoInternalException.java)
+  NOT_IMPLEMENTED               = 202; // Planned, but not implemented yet.
+  FEATURE_NOT_SUPPORTED         = 203; // SQLState: 0A000 - Unsupported feature (usually for unreasonable feature)
+  INVALID_RPC_CALL              = 204; // When invalid RPC call is invoked (e.g., wrong message and absent fields)
+
+  // Query Management and Scheduler
+  NO_SUCH_QUERYID               = 300; // No query id in TajoMaster
```
jihoonson: For consistency, the error id of this section seems to start from 301.

tajo-common/src/main/proto/errors.proto
```diff
@@ -0,0 +1,327 @@
...
+  UNDEFINED_DATABASE                    = 501; // ?
+  UNDEFINED_SCHEMA                      = 502; // ?
+  UNDEFINED_TABLE                       = 503; // ?
+  UNDEFINED_COLUMN                      = 504; // SQLState: 42703
+  UNDEFINED_FUNCTION                    = 505; // SQLState: 42883
+  UNDEFINED_INDEX                       = 506; // ?
+  UNDEFINED_INDEX_NAME                  = 507; // ?
+  UNDEFINED_PARTITION                   = 508; // ?
+  UNDEFINED_PARTITION_METHOD            = 509; // ?
+  UNDEFINED_OPERATOR                    = 510; // SQLState: 42883 (=UNDEFINED_FUNCTION)
```
jihoonson: UNDEFINED_OPERATOR and DUPLICATE_TABLESPACE have the same error code of 510.

tajo-common/src/main/proto/errors.proto
```diff
@@ -0,0 +1,327 @@
...
+  // Data Exception (SQLState Class - 22)
+  DIVISION_BY_ZERO              = 451; // SQLState: 22012 - Division by zero
+
+
+  // Section: Class 42 - Syntax Error or Access Rule Violation
+  SYNTAX_ERROR_OR_ACCESS_RULE_VIOLATION = 601; // SQLState: 42000
+  SYNTAX_ERROR                          = 602; // SQLState: 42601
+  INSUFFICIENT_PRIVILEGE                = 603; // SQLState: 42501
+
+  UNDEFINED_TABLESPACE                  = 500; // ?
```
jihoonson: For consistency, the error id of this section seems to start from 501.

tajo-common/src/main/proto/errors.proto
```diff
@@ -0,0 +1,327 @@
...
+
+
+
+  // Expressions
+  INVALID_EXPRESSION                    = 701;
+  INVALID_CAST                          = 702;
+  INVALID_DATATYPE                      = 703;
+
+  //NUMERIC_OVERFLOW                    = 803;  // Numeric value overflow
+  //VALUE_LARGER_THAN_PRECISION         = 804;  // Value larger than column precision
```
jihoonson: What is the meaning of these commented out codes? If they are TODOs, please add some comments.

tajo-common/src/main/proto/errors.proto
```diff
@@ -0,0 +1,327 @@
...
+  EDOM                          = 10033;  // Math argument out of domain of func
+  ERANGE                        = 10034;  // Math result not representable
+;
+  EDEADLK                       = 10035;  // Resource deadlock would occur
+  ENAMETOOLONG                  = 10036;  // File name too long
+  ENOLCK                        = 10037;  // No record locks available
+  ENOSYS                        = 10038;  // Function not implemented
+  ENOTEMPTY                     = 10039;  // Directory not empty
+  ELOOP                         = 10040;  // Too many symbolic links encountered
+  // EWOULDBLOCK                = EAGAIN  // Operation would block
```
jihoonson: EWOULDBLOCK and EDEADLOCK are commented out.
What is the meaning of these commented out codes? If they are TODOs, please add some comments.

jihoonson:
@hyunsik thank you for the nice work.
In addition to the above my comments, I have more comments to discuss with you.
* Duplicate exception VS Ambiguous exception
 * It seems that every exception is the duplicate exception when there already exists the same thing. However, IMHO, it would be better if we support ambiguous exception as well depending on the situation. For example, the duplicate exception is appropriate for creating a table while the ambiguous exception is appropriate for getting or dropping a table.
* CatalogExceptionUtil and ExceptionUtil
 * These classes look good, but only some methods are used in your implementation. Do you have any reason?
* In your implementation, ```INSUFFICIENT_PRIVILEGE``` is threw if uses try to remove the default database. I think that it would be much proper to use this error code for the real privilege violation after we support the privilege feature. I've also tested that pgsql allows for users to remove the default database if they have sufficient privilege. I wonder why you chose this implementation.

Additionally, please add a description of ```TajoInternalError``` on our wiki.

tajo-catalog/tajo-catalog-client/src/main/java/org/apache/tajo/catalog/AbstractCatalogClient.java
hyunsik: There is a mismatch. DuplicateFunctionException should be AmbiguousFunctionException. It's my intension. 'Duplicate function exception' will occur only at CREATE FUNCTION.

tajo-catalog/tajo-catalog-client/src/main/proto/CatalogProtocol.proto
hyunsik: I agreed.

tajo-catalog/tajo-catalog-client/src/main/proto/CatalogProtocol.proto
hyunsik: I agreed.

tajo-catalog/tajo-catalog-client/src/main/proto/CatalogProtocol.proto
hyunsik: My intension was to merge it into FunctionResponse in ClientProtos.proto because duplicate proto type is used in multiple places. But, I finished the merge yet. I'll merge them in this patch.

tajo-catalog/tajo-catalog-common/src/main/java/org/apache/tajo/catalog/exception/TMCConnectionException.java
hyunsik: I agreed. I'll change it to MetadataConnectionException.

tajo-catalog/tajo-catalog-common/src/main/java/org/apache/tajo/catalog/exception/CatalogExceptionUtil.java
hyunsik: This is not exception class. It's an utility class for exception.

tajo-catalog/tajo-catalog-drivers/tajo-hive/src/main/java/org/apache/tajo/catalog/store/HiveCatalogUtil.java
hyunsik: Fixed

tajo-client/src/main/java/org/apache/tajo/client/SessionConnection.java
hyunsik: I won't use TM for Tajo Metadata Connection. We can use TM only for TajoMaster. How do you think?

tajo-common/src/main/java/org/apache/tajo/exception/TajoException.java
hyunsik: I agreed.

tajo-common/src/main/proto/errors.proto
hyunsik: I agreed.

tajo-common/src/main/proto/errors.proto
hyunsik: I agreed.

tajo-common/src/main/proto/errors.proto
hyunsik: I agreed.

tajo-common/src/main/proto/errors.proto
hyunsik: I'll remove them. Actually, this error table is still being evolving. I'll change many things in in later patch.

tajo-common/src/main/proto/errors.proto
hyunsik: ```EOULDBLOCK = EAGAIN``` means that they are the same value. They look enough for me.

hyunsik:
Thank you for your comments.

Duplicate and ambiguous are clearly distinguished. Duplicate error will occur in only add or creation command. For example, ALTER ADD, CREATE TABLE, CREATE DATABASE will incur duplicate name. Ambiguous occurs in getter method like ```SELECT FROM...``` and getFunction(). So, there are only three cases (table, column, and function) for ambiguous errors. Especially, JDBC specification distinguish both kinds of errors. So, we need them all too.

All methods in ExceptionUtil seems to be used. CatalogExceptionUtil includes four unused methods.
I'll remove them.

Currently, Tajo is assumed to have only one default database through planning. So, we should prohibit the removal of ``DEFAULT`` database. To handle them, in the current implementation, I regard default database and information schema database as read-only things. So, the attempt to remove ``DEFAULT`` or ``INFORMATION_SCHEMA`` cause ```INSUFFICIENT_PRIVILEGE```. In addition, It can be easily matched with JDBC errors. Otherwise, we make our own SQLState code.

tajo-client/src/main/java/org/apache/tajo/client/SessionConnection.java
jihoonson: Ok. If so, please add a brief comment.

tajo-catalog/tajo-catalog-common/src/main/java/org/apache/tajo/catalog/exception/CatalogExceptionUtil.java
jihoonson: Right. I mean, every method in this class creates an exception, but it's difficult to guess from its name. 
Its usage is primarily found in the classes of the same package. So, many parts call them via just their method name without the class name. I think it reduces the readability. 
What do you think?

tajo-common/src/main/proto/errors.proto
jihoonson: Got it. Thanks.

jihoonson:
Thank you for the detailed reply. I agree with you on all issues. 
Regarding on distinguishing duplicate and ambiguous exception, my original intention is that they should be distinguished, but not in this patch. For more specifically, I couldn't find the implementation of ambiguous (table, column, function) exception. 
Maybe you have a future plan for this because there are some error codes defined for ambiguous errors. What do you think?

hyunsik:
@jihoonson Ambiguous errors will be added in the next issue because they requires lots of changes in planning parts. I didn't want to make a big patch :)

jihoonson:
I got it. Thanks for your consideration!

jihoonson:
+1 LGTM!

PR #622, opened 2015-07-02
TAJO-1346: Run ADD PARTITION after storing column partitioned table.

blrunner:
Currently, If new partitions are added to HDFS by ColPartitionStoreExec, CatalogStore will not be aware of these partitions. So, Tajo need to run ALTER TABLE table_name ADD PARTITION on each of the newly added partitions automatically.


blrunner:
Not yet added test cases.

blrunner:
I added unit test cases. For your information, some codes would be removed after finishing TAJO-1345.

PR #623, opened 2015-07-03
TAJO-1672 Removing rest api to create table POST /databases/{database-name}/tables interface

charsyam:
Currently, POST /databases/{database-name}
/tables interface is too hard to use. Because it just serialize TableDesc Class. so I think just using query interface is better than create table interface
and for them, changed result of query interface
removing location header, and add response body has query, and ResultCode
and url.(if query is finished directlry, uri isn't)

### DIFF
--- a/tajo-core/src/main/java/org/apache/tajo/ws/rs/resources/QueryResource.java
+++ b/tajo-core/src/main/java/org/apache/tajo/ws/rs/resources/QueryResource.java
@@ -18,6 +18,7 @@
 
 package org.apache.tajo.ws.rs.resources;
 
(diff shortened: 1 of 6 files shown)


blrunner:
Hi @charsyam 

Thanks for your contribution.
Could you rebase it for the Travis CI build?

charsyam:
@blrunner I rebased it :)

blrunner:
Thanks! I'll check it. :)

blrunner:
+1 

Thanks @charsyam 
I've just committed it to the master branch.

PR #624, opened 2015-07-06
TAJO-1493: Make partition pruning based on catalog informations.

blrunner:
Not yet implemented unit test cases.


blrunner:
I implemented CatalogStore::getPartitionsByDirectSql which execute a select statement to CatalogStore for searching right partition directories. But current tajo testing cluster use MemStore for default CatalogStore and I couldn't execute the select statement because MemStore just retain all catalog informations to HashMap. If you want to verify this patch, you can test as following:

```
mvn clean install -Pparallel-test -DLOG_LEVEL=WARN -Dmaven.fork.count=2 -Dtajo.catalog.store.class=org.apache.tajo.catalog.store.DerbyStore
```

blrunner:
I found that this patch run as expected with HiveCatalogStore and MySQLStore on my testing cluster. And simple query response had been reported as following:

* Table schema:
```
create table partitioned_lineitem (L_SUPPKEY bigint, L_LINENUMBER bigint,
L_QUANTITY double, L_EXTENDEDPRICE double, L_DISCOUNT double, L_TAX double, L_RETURNFLAG text, L_LINESTATUS text,
L_SHIPDATE text, L_COMMITDATE text, L_RECEIPTDATE text, L_SHIPINSTRUCT text, L_SHIPMODE text, L_COMMENT text)
partition by column (L_ORDERKEY bigint, L_PARTKEY bigint)
```
* Partition numbers: 100,000
* Select statement: select * from partitioned_lineitem limit 10;
* Response time:
- previous rewriter: 15 ~ 16 sec
- improved rewriter: 12 ~ 13 sec

Honestly, I didn't implement unit test cases for executing queries because current almost tajo unit cases operate on MemStore. If we apply DerbyStore to some unit test cases for physical operator, we would make a lot of effort. It seems not to be the scope of this patch. So, I just added unit test cases for verifying direct sql. But if you want to test this patch with build commands, you can test with `-Dtajo.catalog.store.class` parameter as following:
```
mvn clean install -Pparallel-test -DLOG_LEVEL=WARN -Dmaven.fork.count=2 -Dtajo.catalog.store.class=org.apache.tajo.catalog.store.DerbyStore
```

tajo-catalog/tajo-catalog-server/src/main/java/org/apache/tajo/catalog/store/MemStore.java
```diff
@@ -28,7 +28,15 @@
 import org.apache.tajo.catalog.*;
 import org.apache.tajo.catalog.exception.*;
 import org.apache.tajo.catalog.proto.CatalogProtos;
-import org.apache.tajo.catalog.proto.CatalogProtos.*;
+//import org.apache.tajo.catalog.proto.CatalogProtos.ColumnProto;
```
hyunsik: You may forget the removal of commented out lines.

tajo-catalog/tajo-catalog-server/src/main/java/org/apache/tajo/catalog/CatalogServer.java
```diff
@@ -1195,6 +1249,71 @@ public GetTablePartitionsResponse getAllPartitions(RpcController controller, Nul
     }
 
     @Override
+    public GetTablePartitionsResponse getPartitionsByDirectSql(RpcController controller,
```
hyunsik: Catalog API is public APIs. In terms of this point, this API design does not make sense due to the following problems:
 * Users should know its schema and underlying persistent storage.
  * So, users should make SQL statements for all persistent storages.
 * This approach is very prone to SQL injection.

tajo-catalog/tajo-catalog-server/src/main/java/org/apache/tajo/catalog/store/AbstractDBStore.java
```diff
@@ -2114,6 +2114,127 @@ private void setPartitionKeys(int pid, PartitionDescProto.Builder partitionDesc)
...
+    } catch (SQLException se) {
+      throw new TajoInternalError(se);
+    } finally {
+      CatalogUtil.closeQuietly(pstmt, res);
+    }
+    return result;
+  }
+
+  @Override
+  public List<TablePartitionProto> getPartitionsByDirectSql(GetPartitionsWithDirectSQLRequest request)
```
hyunsik: Catalog API is public. In terms of this point, this API design does not make sense due to the following problems:
 * Users should know table schemas, their relationship, which underlying persistent storage.
  * Users should make SQL statements for all persistent storages.
 * This approach is very prone to SQL injection.

The solution is that you can use some intermediate representation, using string or java objects. They should be transformed into SQL statements depending on persistent storages in each catalog driver.

blrunner:
Hi @hyunsik 

I've implemented partition API using algebra expressions. For this feature, I implemented  some visitors as following:

* ScanQualConverter: This converts Quals of ScanNode to algebra expressions.
* PartitionFilterAlgebraVisitor: This build SQL statement for getting partition filters by visiting algebra expressions.
* PartitionFilterAlgebraVisitor: This build SQL statement for getting partition filters by visiting EvalNodes.

blrunner:
I added DateTime types and more unit test cases for various partition column types.

blrunner:
I've finished tests successfully for this PR as following:

* Data: TPC-H benchmark set (scale factor = 1)
* CatalogStore: MySQLStore, HiveCatalogStore(with apache hive 1.1.0)
* CTAS for creating partition tables (except partsupp)
```
CREATE TABLE LINEITEM_P ( 
L_ORDERKEY bigint, L_PARTKEY bigint, L_SUPPKEY bigint, L_LINENUMBER bigint, L_QUANTITY double, L_EXTENDEDPRICE double, L_DISCOUNT double, L_TAX double, L_RETURNFLAG text, L_LINESTATUS text,L_COMMITDATE date, L_RECEIPTDATE date, L_SHIPINSTRUCT text, L_SHIPMODE text, L_COMMENT text) 
USING TEXT WITH ('text.delimiter'='|')
PARTITION BY COLUMN (L_SHIPDATE date) AS 
SELECT L_ORDERKEY, L_PARTKEY, L_SUPPKEY, L_LINENUMBER, L_QUANTITY, L_EXTENDEDPRICE, L_DISCOUNT, L_TAX, L_RETURNFLAG, L_LINESTATUS, L_COMMITDATE, L_RECEIPTDATE, L_SHIPINSTRUCT, L_SHIPMODE, L_COMMENT, L_SHIPDATE FROM LINEITEM;

create table customer_p (C_CUSTKEY bigint, C_NAME text, C_ADDRESS text, C_PHONE text, C_ACCTBAL double, C_MKTSEGMENT text, C_COMMENT text) 
USING TEXT WITH ('text.delimiter'='|')
PARTITION BY COLUMN (C_NATIONKEY bigint) AS 
SELECT C_CUSTKEY, C_NAME, C_ADDRESS, C_PHONE, C_ACCTBAL, C_MKTSEGMENT, C_COMMENT, C_NATIONKEY FROM customer;

create  table supplier_p (S_SUPPKEY bigint, S_NAME text, S_ADDRESS text, S_PHONE text, S_ACCTBAL double, S_COMMENT text)
USING TEXT WITH ('text.delimiter'='|')
PARTITION BY COLUMN ( S_NATIONKEY bigint) AS 
SELECT S_SUPPKEY, S_NAME, S_ADDRESS, S_PHONE, S_ACCTBAL, S_COMMENT, S_NATIONKEY FROM supplier;

create  table part_p (P_PARTKEY bigint, P_NAME text, P_MFGR text, P_BRAND text, P_TYPE text, P_CONTAINER text, P_RETAILPRICE double, P_COMMENT text) 
USING TEXT WITH ('text.delimiter'='|')
PARTITION BY COLUMN (P_SIZE integer) AS 
SELECT P_PARTKEY, P_NAME, P_MFGR, P_BRAND, P_TYPE, P_CONTAINER, P_RETAILPRICE, P_COMMENT, P_SIZE FROM part;

create table orders_p (O_ORDERKEY bigint, O_CUSTKEY bigint, O_ORDERSTATUS text, O_TOTALPRICE double, O_ORDERPRIORITY text, O_CLERK text, O_SHIPPRIORITY int, O_COMMENT text)
USING TEXT WITH ('text.delimiter'='|')
PARTITION BY COLUMN (O_ORDERDATE date) AS 
SELECT O_ORDERKEY, O_CUSTKEY, O_ORDERSTATUS, O_TOTALPRICE, O_ORDERPRIORITY, O_CLERK, O_SHIPPRIORITY, O_COMMENT, O_ORDERDATE FROM orders;

create table nation_p (N_NATIONKEY bigint, N_NAME text, N_COMMENT text) 
USING TEXT WITH ('text.delimiter'='|')
PARTITION BY COLUMN (N_REGIONKEY bigint) AS 
SELECT N_NATIONKEY, N_NAME, N_COMMENT, N_REGIONKEY FROM nation;

create table region_p ( R_NAME text,  R_COMMENT text) 
USING TEXT WITH ('text.delimiter'='|')
PARTITION BY COLUMN (R_REGIONKEY bigint) AS 
SELECT R_NAME,  R_COMMENT, R_REGIONKEY FROM region;
```
* Test Queries : TPC-H Q1 ~ Q22

After executing CTAS statements, I fount that list of table partitions had been printed on hive shell. All test results mostly accorded with results of queries for non-partitoned tables in MySQLStore and HiveCatalogStore. But some records for double type record aggregation were slightly different in the nearest whole number, To resolve above problem, we need to provide decimal column type.

tajo-catalog/tajo-catalog-client/src/main/java/org/apache/tajo/catalog/AbstractCatalogClient.java
```diff
@@ -428,13 +429,12 @@ public final PartitionDescProto getPartition(final String databaseName, final St
   }
 
   @Override
-  public final List<PartitionDescProto> getPartitions(final String databaseName, final String tableName) {
+  public final List<PartitionDescProto> getPartitions(final String databaseName, final String tableName) throws
+    UndefinedDatabaseException, UndefinedTableException, UndefinedPartitionMethodException,
```
hyunsik: To handle such exceptions, you should put ``throwsIfThisError`` method for the exceptions.

tajo-catalog/tajo-catalog-client/src/main/java/org/apache/tajo/catalog/AbstractCatalogClient.java
```diff
@@ -446,6 +446,46 @@ public final PartitionDescProto getPartition(final String databaseName, final St
...
+      final TableIdentifierProto request = buildTableIdentifier(databaseName, tableName);
+      return isSuccess(stub.existPartitionsByTableName(null, request));
+    } catch (ServiceException e) {
+      throw new RuntimeException(e);
+    }
+  }
+
+  @Override
+  public List<PartitionDescProto> getPartitionsByAlgebra(PartitionsByAlgebraProto request) throws
+    UndefinedDatabaseException, UndefinedTableException, UndefinedPartitionMethodException,
```
hyunsik: To handle such exceptions, you should put throwsIfThisError method for the exceptions. Please refer to https://github.com/apache/tajo/pull/624/files#diff-9ecb2222e062d0cc008fb1c3da61d4b0L483.

tajo-catalog/tajo-catalog-client/src/main/java/org/apache/tajo/catalog/AbstractCatalogClient.java
```diff
@@ -446,6 +446,46 @@ public final PartitionDescProto getPartition(final String databaseName, final St
...
+    return response.getPartitionList();
+    } catch (ServiceException e) {
+      LOG.error(e.getMessage(), e);
+      return null;
+    }
+  }
+
+  @Override
+  public List<PartitionDescProto> getPartitionsByDirectSql(PartitionsByDirectSqlProto request) throws
+    UndefinedDatabaseException, UndefinedTableException, UndefinedPartitionMethodException,
```
hyunsik: To handle such exceptions, you should put throwsIfThisError method for the exceptions. Please refer to https://github.com/apache/tajo/pull/624/files#diff-9ecb2222e062d0cc008fb1c3da61d4b0L483.

tajo-catalog/tajo-catalog-common/src/main/java/org/apache/tajo/catalog/CatalogService.java
```diff
@@ -162,7 +162,19 @@ PartitionDescProto getPartition(String databaseName, String tableName, String pa
       throws UndefinedPartitionException, UndefinedPartitionMethodException, UndefinedDatabaseException,
       UndefinedTableException;
 
-  List<PartitionDescProto> getPartitions(String databaseName, String tableName);
+  List<PartitionDescProto> getPartitions(String databaseName, String tableName) throws UndefinedDatabaseException,
```
hyunsik: It would be better if ``getPartitions()`` is renamed to ``getAllPartitions()``.

tajo-catalog/tajo-catalog-client/src/main/java/org/apache/tajo/catalog/AbstractCatalogClient.java
```diff
@@ -428,13 +429,12 @@ public final PartitionDescProto getPartition(final String databaseName, final St
   }
 
   @Override
-  public final List<PartitionDescProto> getPartitions(final String databaseName, final String tableName) {
+  public final List<PartitionDescProto> getPartitions(final String databaseName, final String tableName) throws
```
hyunsik: It would be better if ``getPartitions()`` is renamed to ``getAllPartitions()``.

tajo-catalog/tajo-catalog-drivers/tajo-hive/src/main/java/org/apache/tajo/catalog/store/HiveCatalogStore.java
```diff
@@ -846,10 +855,65 @@ public boolean existPartitionMethod(String databaseName, String tableName) throw
...
+  public boolean existPartitions(String databaseName, String tableName) throws UndefinedDatabaseException,
+    UndefinedTableException, UndefinedPartitionMethodException {
+    throw new UnsupportedOperationException();
+  }
+
+  @Override
+  public List<PartitionDescProto> getPartitionsByAlgebra(PartitionsByAlgebraProto request)
+    throws UndefinedDatabaseException, UndefinedTableException, UndefinedPartitionMethodException,
+    UndefinedOperatorException {
+    throw new UndefinedOperatorException("getPartitionsByAlgebra");
```
hyunsik: UnsupportedException is proper here.

tajo-catalog/tajo-catalog-drivers/tajo-hive/src/main/java/org/apache/tajo/catalog/store/HiveCatalogStore.java
```diff
@@ -846,10 +855,65 @@ public boolean existPartitionMethod(String databaseName, String tableName) throw
...
-                                                         String tableName) {
+                                                         String tableName) throws UndefinedDatabaseException,
+    UndefinedTableException, UndefinedPartitionMethodException, UndefinedPartitionException{
     throw new UnsupportedOperationException();
   }
 
+  @Override
+  public boolean existPartitions(String databaseName, String tableName) throws UndefinedDatabaseException,
+    UndefinedTableException, UndefinedPartitionMethodException {
+    throw new UnsupportedOperationException();
```
hyunsik: UnsupportedException is proper here.

tajo-catalog/tajo-catalog-drivers/tajo-hive/src/main/java/org/apache/tajo/catalog/store/HiveCatalogStore.java
```diff
@@ -846,10 +855,65 @@ public boolean existPartitionMethod(String databaseName, String tableName) throw
 
   @Override
   public List<CatalogProtos.PartitionDescProto> getPartitions(String databaseName,
-                                                         String tableName) {
+                                                         String tableName) throws UndefinedDatabaseException,
+    UndefinedTableException, UndefinedPartitionMethodException, UndefinedPartitionException{
     throw new UnsupportedOperationException();
```
hyunsik: UnsupportedException is proper here.

tajo-catalog/tajo-catalog-server/src/main/java/org/apache/tajo/catalog/store/AbstractDBStore.java
```diff
@@ -2173,6 +2179,244 @@ private void setPartitionKeys(int pid, PartitionDescProto.Builder partitionDesc)
...
+      CatalogUtil.closeQuietly(pstmt, res);
+    }
+    return result;
+  }
+
+  @Override
+  public List<PartitionDescProto> getPartitionsByDirectSql(PartitionsByDirectSqlProto request)
+    throws UndefinedDatabaseException, UndefinedTableException, UndefinedPartitionMethodException,
+    UndefinedOperatorException {
+    throw new UndefinedOperatorException("getPartitionsByDirectSql");
```
hyunsik: It should be UnsupportedException.

tajo-catalog/tajo-catalog-drivers/tajo-hive/src/main/java/org/apache/tajo/catalog/store/HiveCatalogStore.java
```diff
@@ -846,10 +855,65 @@ public boolean existPartitionMethod(String databaseName, String tableName) throw
...
+  public List<PartitionDescProto> getPartitionsByAlgebra(PartitionsByAlgebraProto request)
+    throws UndefinedDatabaseException, UndefinedTableException, UndefinedPartitionMethodException,
+    UndefinedOperatorException {
+    throw new UndefinedOperatorException("getPartitionsByAlgebra");
+  }
+
+  @Override
+  public List<PartitionDescProto> getPartitionsByDirectSql(PartitionsByDirectSqlProto request)
+      throws UndefinedDatabaseException, UndefinedTableException, UndefinedPartitionMethodException,
+      UndefinedOperatorException {
```
hyunsik: UndefinedOperator is not proper here.

tajo-catalog/tajo-catalog-drivers/tajo-hive/src/main/java/org/apache/tajo/catalog/store/HiveCatalogStore.java
```diff
@@ -846,10 +855,65 @@ public boolean existPartitionMethod(String databaseName, String tableName) throw
...
+        builder.setPath(hivePartition.getSd().getLocation());
+
+        int startIndex = hivePartition.getSd().getLocation().indexOf(tableName) + tableName.length();
+        String partitionName = hivePartition.getSd().getLocation().substring(startIndex+1);
+        builder.setPartitionName(partitionName);
+
+        partitions.add(builder.build());
+      }
+    } catch (NoSuchObjectException e) {
+      return null;
```
hyunsik: This should throw TajoInternalError.

tajo-catalog/tajo-catalog-server/src/main/java/org/apache/tajo/catalog/CatalogServer.java
```diff
@@ -957,7 +957,61 @@ public GetPartitionDescResponse getPartitionByPartitionName(RpcController contro
...
+      if (metaDictionary.isSystemDatabase(dbName)) {
+        return errUndefinedPartitionMethod(tbName);
+      }
+
+      rlock.lock();
+      try {
+        boolean contain;
+
+        contain = store.existDatabase(dbName);
+        if (contain) {
```
hyunsik: These tests can be improved like other methods. Please refer to existsDatabase() method.

tajo-catalog/tajo-catalog-server/src/main/java/org/apache/tajo/catalog/CatalogServer.java
```diff
@@ -1027,6 +1081,136 @@ public GetTablePartitionsResponse getAllPartitions(RpcController controller, Nul
...
+      if (metaDictionary.isSystemDatabase(dbName)) {
+        return GetPartitionsResponse.newBuilder().setState(errUndefinedPartitionMethod(tbName)).build();
+      }
+
+      rlock.lock();
+      try {
+        boolean contain;
+
+        contain = store.existDatabase(dbName);
+        if (contain) {
```
hyunsik: These tests can be improved like other methods. Please refer to existsDatabase() method.

tajo-catalog/tajo-catalog-server/src/main/java/org/apache/tajo/catalog/CatalogServer.java
```diff
@@ -1027,6 +1081,136 @@ public GetTablePartitionsResponse getAllPartitions(RpcController controller, Nul
...
```
hyunsik: These tests can be improved like other methods. Please refer to existsDatabase() method.

blrunner:
@hyunsik 

Thank your for your detailed review.
I've just updated this PR using your comments.

tajo-catalog/tajo-catalog-client/src/main/java/org/apache/tajo/catalog/AbstractCatalogClient.java
```diff
@@ -446,6 +450,52 @@ public final PartitionDescProto getPartition(final String databaseName, final St
   }
 
   @Override
+  public boolean existPartitions(String databaseName, String tableName) {
+    try {
+      final BlockingInterface stub = getStub();
+      final TableIdentifierProto request = buildTableIdentifier(databaseName, tableName);
+      return isSuccess(stub.existPartitionsByTableName(null, request));
```
hyunsik: Even through this API is checking if the table exists, it potentially returns unexpected exception. If you check only its success, it may return a wrong result.

So, you should return true or false depending on only both its success and UNDEFINED_PARTITION_METHODS. You can find the similar code pattern from ``existPartitionMethod`` method in this class.

tajo-catalog/tajo-catalog-client/src/main/java/org/apache/tajo/catalog/AbstractCatalogClient.java
```diff
@@ -446,6 +450,52 @@ public final PartitionDescProto getPartition(final String databaseName, final St
...
+    UndefinedDatabaseException, UndefinedTableException, UndefinedPartitionMethodException {
+    try {
+      final BlockingInterface stub = getStub();
+      GetPartitionsResponse response = stub.getPartitionsByAlgebra(null, request);
+
+      throwsIfThisError(response.getState(), UndefinedDatabaseException.class);
+      throwsIfThisError(response.getState(), UndefinedTableException.class);
+      throwsIfThisError(response.getState(), UndefinedPartitionMethodException.class);
+      ensureOk(response.getState());
+    return response.getPartitionList();
```
hyunsik: mis indent.

tajo-catalog/tajo-catalog-client/src/main/java/org/apache/tajo/catalog/AbstractCatalogClient.java
```diff
@@ -446,6 +450,52 @@ public final PartitionDescProto getPartition(final String databaseName, final St
   }
 
   @Override
+  public boolean existPartitions(String databaseName, String tableName) {
```
hyunsik: How about renaming it into ``isPartitionedTable``? It seems to be better.

tajo-catalog/tajo-catalog-client/src/main/java/org/apache/tajo/catalog/AbstractCatalogClient.java
```diff
@@ -428,15 +429,18 @@ public final PartitionDescProto getPartition(final String databaseName, final St
   }
 
   @Override
-  public final List<PartitionDescProto> getPartitions(final String databaseName, final String tableName) {
+  public final List<PartitionDescProto> getAllPartitions(final String databaseName, final String tableName) throws
+    UndefinedDatabaseException, UndefinedTableException, UndefinedPartitionMethodException,
+    UndefinedPartitionException {
```
hyunsik: It retrieves all partitions. So it may be inproper to throw ``UndefinedPartitionException``, which should be thrown in the case where there is no partition corresponding to a specific partition key.

tajo-catalog/tajo-catalog-client/src/main/proto/CatalogProtocol.proto
```diff
@@ -121,9 +121,12 @@ service CatalogProtocolService {
   rpc existPartitionMethod(TableIdentifierProto) returns (ReturnState);
 
   rpc getPartitionByPartitionName(PartitionIdentifierProto) returns (GetPartitionDescResponse);
-  rpc getPartitionsByTableName(PartitionIdentifierProto) returns (GetPartitionsResponse);
+  rpc getPartitionsByTableName(TableIdentifierProto) returns (GetPartitionsResponse);
+  rpc existPartitionsByTableName(TableIdentifierProto) returns (ReturnState);
```
hyunsik: As I mentioned, ``isPartitionedTable`` would be better. It may not need the suffix ``byPartitionName``.

tajo-catalog/tajo-catalog-common/src/main/java/org/apache/tajo/catalog/CatalogService.java
```diff
@@ -162,7 +162,17 @@ PartitionDescProto getPartition(String databaseName, String tableName, String pa
...
+  List<PartitionDescProto> getAllPartitions(String databaseName, String tableName) throws UndefinedDatabaseException,
+    UndefinedTableException, UndefinedPartitionMethodException, UndefinedPartitionException, UndefinedOperatorException;
+
+  boolean existPartitions(String databaseName, String tableName) throws UndefinedDatabaseException,
+  UndefinedTableException, UndefinedPartitionMethodException;
+
+  List<PartitionDescProto> getPartitionsByAlgebra(PartitionsByAlgebraProto request) throws
+    UndefinedDatabaseException, UndefinedTableException, UndefinedPartitionMethodException;
+
+  List<PartitionDescProto> getPartitionsByDirectSql(PartitionsByDirectSqlProto request) throws
```
hyunsik: Its name should be changed into {{PartitionByFilter}} or {{PartitionByPredicate}}. As far as I know, the suffix ``DirectSQL`` is inspired by Hive API. But, the Hive API does predefined filter expression instead of direct sql predicate. Direct SQL is always danger, and this is not our way.

tajo-catalog/tajo-catalog-drivers/tajo-hive/src/main/java/org/apache/tajo/catalog/store/HiveCatalogStore.java
```diff
@@ -845,11 +854,72 @@ public boolean existPartitionMethod(String databaseName, String tableName) throw
   }
 
   @Override
-  public List<CatalogProtos.PartitionDescProto> getPartitions(String databaseName,
-                                                         String tableName) {
-    throw new UnsupportedOperationException();
+  public List<CatalogProtos.PartitionDescProto> getPartitions(String databaseName, String tableName) {
+    throw new TajoRuntimeException(new UnsupportedException("getPartitions"));
```
hyunsik: UnsupportedException automatically generates the error message including the method and class if it is used with a default constructor. SO, it should be ``new UnsupportedException()``.

tajo-catalog/tajo-catalog-drivers/tajo-hive/src/main/java/org/apache/tajo/catalog/store/HiveCatalogStore.java
```diff
@@ -845,11 +854,72 @@ public boolean existPartitionMethod(String databaseName, String tableName) throw
...
+  }
+
+  @Override
+  public boolean existPartitions(String databaseName, String tableName) {
+    throw new TajoRuntimeException(new UnsupportedException("existPartitions"));
+  }
+
+  @Override
+  public List<PartitionDescProto> getPartitionsByAlgebra(PartitionsByAlgebraProto request) {
+    throw new TajoRuntimeException(new UnsupportedException("getPartitionsByAlgebra"));
```
hyunsik: UnsupportedException automatically generates the error message including the method and class if it is used with a default constructor. SO, it should be new UnsupportedException().

tajo-catalog/tajo-catalog-drivers/tajo-hive/src/main/java/org/apache/tajo/catalog/store/HiveCatalogStore.java
```diff
@@ -1052,7 +1122,7 @@ public void addPartitions(String databaseName, String tableName, List<CatalogPro
       }
 
       if (addPartitions.size() > 0) {
-        client.getHiveClient().add_partitions(addPartitions, true, true);
+        List<Partition> results = client.getHiveClient().add_partitions(addPartitions, true, true);
```
hyunsik: ``List<Partition> results`` is not used.

tajo-catalog/tajo-catalog-server/src/main/java/org/apache/tajo/catalog/store/AbstractDBStore.java
```diff
@@ -2173,6 +2179,248 @@ private void setPartitionKeys(int pid, PartitionDescProto.Builder partitionDesc)
   }
 
   @Override
+  public boolean existPartitions(String databaseName, String tableName) throws UndefinedDatabaseException,
```
hyunsik: It would be better if it is renamed into ``isPartitionedTable``.

tajo-catalog/tajo-catalog-server/src/main/java/org/apache/tajo/catalog/store/AbstractDBStore.java
```diff
@@ -2132,7 +2137,8 @@ private void setPartitionKeys(int pid, PartitionDescProto.Builder partitionDesc)
 
   @Override
   public List<PartitionDescProto> getPartitions(String databaseName, String tableName)
```
hyunsik: It should be getAllPartitions for naming consistency.

tajo-catalog/tajo-catalog-server/src/main/java/org/apache/tajo/catalog/store/AbstractDBStore.java
```diff
@@ -2173,6 +2179,248 @@ private void setPartitionKeys(int pid, PartitionDescProto.Builder partitionDesc)
...
+
+    try {
+      int databaseId = getDatabaseId(request.getDatabaseName());
+      int tableId = getTableId(databaseId, request.getDatabaseName(), request.getTableName());
+      if (!existPartitionMethod(request.getDatabaseName(), request.getTableName())) {
+        throw new UndefinedPartitionMethodException(request.getTableName());
+      }
+
+      TableDescProto tableDesc = getTable(request.getDatabaseName(), request.getTableName());
+      conn = getConnection();
```
hyunsik: It may be very trivial. It would be better if this line is moved above ``pstmt = conn.prepareStatement(directSQL);``.

tajo-plan/src/main/java/org/apache/tajo/plan/rewrite/rules/PartitionedTableRewriter.java
```diff
@@ -118,32 +121,264 @@ public String toString() {
...
-    if (conjunctiveForms == null) {
-      filters = buildAllAcceptingPathFilters(partitionColumns);
-    } else {
-      filters = buildPathFiltersForAllLevels(partitionColumns, conjunctiveForms);
+    String store = queryContext.getConf().get(CatalogConstants.STORE_CLASS);
+
+    try {
+      // HiveCatalogStore provides list of table partitions with where clause because hive just provides api using
+      // the filter string. So, this rewriter need to differentiate HiveCatalogStore and other catalogs.
+      if (store.equals("org.apache.tajo.catalog.store.HiveCatalogStore")) {
```
hyunsik: A RewriteRule should not consider its catalog storage type. It does not make sense. CatalogService should just take an algebra, and HiveCatalogStore should generates a filter expression string representation from the algebra.

blrunner:
@hyunsik 

Thank you for your detailed review and I've just reflected your comments.

tajo-common/src/main/java/org/apache/tajo/exception/ReturnStateUtil.java
```diff
@@ -161,6 +161,10 @@ public static ReturnState errUndefinedPartition(String partitionName) {
     return returnError(ResultCode.UNDEFINED_PARTITION, partitionName);
   }
 
+  public static ReturnState errPartitionNotFund(String tbName) {
```
hyunsik: If it is not used, you may do not need to add this method.

tajo-catalog/tajo-catalog-server/src/main/java/org/apache/tajo/catalog/store/AbstractDBStore.java
```diff
@@ -2172,6 +2178,252 @@ private void setPartitionKeys(int pid, PartitionDescProto.Builder partitionDesc)
     return partitions;
   }
 
+  /**
+   * Check if list of partitios exist on catalog.
```
hyunsik: ``partitios`` seems to be a typo error.

tajo-catalog/tajo-catalog-server/src/main/java/org/apache/tajo/catalog/store/AbstractDBStore.java
```diff
@@ -2172,6 +2178,252 @@ private void setPartitionKeys(int pid, PartitionDescProto.Builder partitionDesc)
...
+
+    try {
+      int databaseId = getDatabaseId(request.getDatabaseName());
+      int tableId = getTableId(databaseId, request.getDatabaseName(), request.getTableName());
+      if (!existPartitionMethod(request.getDatabaseName(), request.getTableName())) {
+        throw new UndefinedPartitionMethodException(request.getTableName());
+      }
+
+      if (!existPartitionsOnCatalog(databaseId, tableId)) {
+        throw new PartitionNotFoundException(request.getTableName());
```
hyunsik: Could you elaborate the purpose of this exception? Does it means that there is no partition? If so, it does not make sense because it returns a list of partitions. So, it should return an empty list if there is no partitions.

tajo-catalog/tajo-catalog-client/src/main/java/org/apache/tajo/catalog/AbstractCatalogClient.java
```diff
@@ -25,6 +25,7 @@
 import org.apache.tajo.catalog.CatalogProtocol.CatalogProtocolService.BlockingInterface;
 import org.apache.tajo.catalog.CatalogProtocol.*;
 import org.apache.tajo.catalog.partition.PartitionMethodDesc;
+import org.apache.tajo.catalog.proto.CatalogProtos;
```
hyunsik: It seems to be not used.

tajo-plan/src/main/java/org/apache/tajo/plan/util/ScanQualConverter.java
```diff
@@ -0,0 +1,297 @@
...
+import org.apache.tajo.datum.TimestampDatum;
+import org.apache.tajo.plan.expr.*;
+
+import java.util.Stack;
+
+/**
+ * This converts Quals of ScanNode to Algebra expressions.
+ *
+ */
+public class ScanQualConverter extends SimpleEvalNodeVisitor<Object> {
```
hyunsik: I'd like to recommend renaming it to ``EvalNodeToExprConverter``. It appears to be more general in terms of converting any EvalNode tree into Expr tree.

tajo-catalog/tajo-catalog-server/src/main/java/org/apache/tajo/catalog/store/AbstractDBStore.java
blrunner: The purpose is for partition pruning when there is no partitions on catalog and partition directories exist on file system. Above case, users need to execute recover partition (https://github.com/apache/tajo/pull/626). But we need to handle their queries if  they don't recover their partitions located on file system to catalog. So, I implemented the exception and if catalog throws the exception, PartitionedTableRewriter would set a list of partitions to null. And then, the rewriter would check directories from file system.

blrunner:
Thanks @hyunsik 

I reflected your comments and left some comments. :)

tajo-core-tests/src/test/java/org/apache/tajo/engine/query/TestTablePartitions.java
```diff
@@ -1325,4 +1332,421 @@ public final void testDuplicatedPartitions() throws Exception {
...
+
+    assertEquals(expectedResult, resultSetToString(res));
+    res.close();
+
+    executeString("DROP TABLE " + tableName + " PURGE").close();
+    res.close();
+  }
+
+  @Test
+  public final void testTimeStampPartitionColumn() throws Exception {
```
hyunsik: Could you rename TimeStamp to Timestamp?

hyunsik:
Thank you for your work. The latest patch looks good to me. Here is my +1. I leaved one comment. You can commit it after reflecting my comment if you agree.

blrunner:
@hyunsik 

Thank you very much for your kind review. I've just reflected your comment. After finishing the travis CI build, I'll ship it. :)

PR #625, opened 2015-07-08
TAJO-1677: Remove unnecessary messages for the Travis CI build

blrunner:
Enable -B (batch) mode on the build. This will make the logs shorter since it avoids the dependency download progress logging.

### DIFF
--- a/tajo-core/src/main/java/org/apache/tajo/master/TajoMaster.java
+++ b/tajo-core/src/main/java/org/apache/tajo/master/TajoMaster.java
@@ -173,7 +173,7 @@ public class TajoMaster extends CompositeService {
       RpcClientManager rpcManager = RpcClientManager.getInstance();
       rpcManager.setRetries(systemConf.getInt(RpcConstants.RPC_CLIENT_RETRY_MAX, RpcConstants.DEFAULT_RPC_RETRIES));
       rpcManager.setTimeoutSeconds(
-          systemConf.getInt(RpcConstants.RPC_CLIENT_TIMEOUT_SECS, RpcConstants.DEFAULT_RPC_TIMEOUT_SECONDS));
+        systemConf.getInt(RpcConstants.RPC_CLIENT_TIMEOUT_SECS, RpcConstants.DEFAULT_RPC_TIMEOUT_SECONDS));
 
       initResourceManager();
 
@@ -236,7 +236,7 @@ public class TajoMaster extends CompositeService {
 
   private void initResourceManager() throws Exception {
     Class<WorkerResourceManager>  resourceManagerClass = (Class<WorkerResourceManager>)
-        systemConf.getClass(ConfVars.RESOURCE_MANAGER_CLASS.varname, TajoWorkerResourceManager.class);
+      systemConf.getClass(ConfVars.RESOURCE_MANAGER_CLASS.varname, TajoWorkerResourceManager.class);
     Constructor<WorkerResourceManager> constructor = resourceManagerClass.getConstructor(MasterContext.class);
     resourceManager = constructor.newInstance(context);
     addIfService(resourceManager);
@@ -246,7 +246,7 @@ public class TajoMaster extends CompositeService {
     if (!systemConf.get(CommonTestingUtil.TAJO_TEST_KEY, "FALSE").equalsIgnoreCase("TRUE")) {
       InetSocketAddress address = systemConf.getSocketAddrVar(ConfVars.TAJO_MASTER_INFO_ADDRESS);
       webServer = StaticHttpServer.getInstance(this ,"admin", address.getHostName(), address.getPort(),
-          true, null, context.getConf(), null);
+        true, null, context.getConf(), null);
       webServer.addServlet("queryServlet", "/query_exec", QueryExecutorServlet.class);
       webServer.start();
     }
@@ -328,7 +328,7 @@ public class TajoMaster extends CompositeService {
 
     // Setting the system global configs
     systemConf.setSocketAddr(ConfVars.CATALOG_ADDRESS.varname,
-        NetUtils.getConnectAddress(catalogServer.getBindAddress()));
+      NetUtils.getConnectAddress(catalogServer.getBindAddress()));
 
     try {
       writeSystemConf();
@@ -531,7 +531,7 @@ public class TajoMaster extends CompositeService {
       stream.println("Thread " + getThreadTaskName(info.getThreadId(), info.getThreadName()) + ":");
       Thread.State state = info.getThreadState();
       stream.println("  State: " + state + ", Blocked count: " + info.getBlockedCount() +
-          ", Waited count: " + info.getWaitedCount());
+        ", Waited count: " + info.getWaitedCount());
       if (contention) {
         stream.println("  Blocked time: " + info.getBlockedTime() + ", Waited time: " + info.getWaitedTime());
       }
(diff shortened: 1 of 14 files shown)


blrunner:
I've implemented this patch as follows.

- Apply -q parameter for maven :  Only show errors.
- Apply -ff parameter for maven : Stop at first failure in reactorized builds
- Apply surefire.useFile=false for maven: This causes surefire to print test failures to standard out.
- Clean up TestSystemMetrics log messages.
- Clean up TajoMaster log messages.
- Fix TestBSTIndexExec:testEqual error.
- Remove RestServer warning log messages.

tajo-core/src/main/java/org/apache/tajo/master/TajoMaster.java
```diff
@@ -255,15 +255,15 @@ private void initWebServer() throws Exception {
...
 
     // Check and Create Tajo root dir
     this.defaultFS = tajoRootPath.getFileSystem(systemConf);
     systemConf.set(CommonConfigurationKeysPublic.FS_DEFAULT_NAME_KEY, defaultFS.getUri().toString());
-    LOG.info("FileSystem (" + this.defaultFS.getUri() + ") is initialized.");
     if (!defaultFS.exists(tajoRootPath)) {
       defaultFS.mkdirs(tajoRootPath, new FsPermission(TAJO_ROOT_DIR_PERMISSION));
       LOG.info("Tajo Root Directory '" + tajoRootPath + "' is created.");
+    } else {
+      LOG.info("FileSystem (" + this.defaultFS.getUri() + ") is initialized.");
```
hyunsik: This change seems to be different to the original intension. This log appears to show which HDFS namenode is used in this TajoMaster.

tajo-core/src/main/java/org/apache/tajo/master/TajoMaster.java
```diff
@@ -280,19 +280,21 @@ private void checkAndInitializeSystemDirectories() throws IOException {
...
     // Get Warehouse dir
     this.wareHousePath = TajoConf.getWarehouseDir(systemConf);
-    LOG.info("Tajo Warehouse dir: " + wareHousePath);
 
     // Check and Create Warehouse dir
     if (!defaultFS.exists(wareHousePath)) {
       defaultFS.mkdirs(wareHousePath, new FsPermission(WAREHOUSE_DIR_PERMISSION));
       LOG.info("Warehouse dir '" + wareHousePath + "' is created");
+    } else {
+      LOG.info("Tajo Warehouse dir: " + wareHousePath);
```
hyunsik: It should be recover due to the same reason as I commented above.

tajo-core/src/main/java/org/apache/tajo/master/TajoMaster.java
```diff
@@ -280,19 +280,21 @@ private void checkAndInitializeSystemDirectories() throws IOException {
...
     }
 
     Path stagingPath = TajoConf.getDefaultRootStagingDir(systemConf);
-    LOG.info("Staging dir: " + wareHousePath);
     if (!defaultFS.exists(stagingPath)) {
       defaultFS.mkdirs(stagingPath, new FsPermission(STAGING_ROOTDIR_PERMISSION));
       LOG.info("Staging dir '" + stagingPath + "' is created");
+    } else {
+      LOG.info("Staging dir: " + wareHousePath);
```
hyunsik: It should be recover due to the same reason as I commented above.

tajo-core/src/test/java/org/apache/tajo/engine/planner/physical/TestExternalSortExec.java
```diff
@@ -98,8 +102,10 @@ public void setUp() throws Exception {
     appender.flush();
     appender.close();
 
-    System.out.println(appender.getStats().getNumRows() + " rows (" + (appender.getStats().getNumBytes() / 1048576) +
+    if (LOG.isDebugEnabled()) {
```
hyunsik: This if-condition branch seems to be not necessary.

tajo-core/src/test/java/org/apache/tajo/engine/planner/physical/TestProgressExternalSortExec.java
```diff
@@ -102,7 +106,9 @@ public void setUp() throws Exception {
     appender.flush();
     appender.close();
 
-    System.out.println(appender.getStats().getNumRows() + " rows (" + appender.getStats().getNumBytes() + " Bytes)");
+    if (LOG.isDebugEnabled()) {
```
hyunsik: This if-condition branch seems to be not necessary.

tajo-core/src/test/java/org/apache/tajo/worker/TestNodeResourceManager.java
```diff
@@ -253,7 +256,9 @@ public void run() {
                   fail(e.getMessage());
                 }
               }
-              System.out.println(Thread.currentThread().getName() + " complete requests: " + complete);
+              if (LOG.isDebugEnabled()) {
```
hyunsik: This if-condition branch seems to be not necessary.

tajo-core/src/test/java/org/apache/tajo/worker/TestNodeResourceManager.java
```diff
@@ -264,8 +269,10 @@ public void run() {
       future.get();
     }
 
-    System.out.println(parallelCount + " Thread, completed requests: " + totalComplete.get() + ", canceled requests:"
+    if (LOG.isDebugEnabled()) {
```
hyunsik: This if-condition branch seems to be not necessary.

hyunsik:
The first description mentioned that -B will be added to maven option. Was the plan changed?

hyunsik:
Please see the following volumes. quite.txt contains all messages from '-q' option. For me, `-q` seems to be not effective.
```
hyunsik@Hyunsiks-MacBook-Pro:~$ ls -l quite.txt normal.txt 
-rw-r--r--  1 hyunsik  staff  75536879 Jul 13 16:41 normal.txt
-rw-r--r--  1 hyunsik  staff  75490962 Jul 13 17:14 quite.txt
```

blrunner:
Hi @hyunsik 

I've updated the patch using your comments.
For your information, I added -q parameter instead of -B parameter. In some unexplained way, -B parameter didn't work as expect.

blrunner:
I think that the parameter on your desktop seemed like ineffective, because tajo dependencies already had been installed on your desktop. If you see build progress of this PR, you would see that there is no dependency download progress.

tajo-core/src/main/java/org/apache/tajo/master/TajoMaster.java
```diff
@@ -202,7 +202,7 @@ public void serviceInit(Configuration _conf) throws Exception {
 
       tajoMasterService = new QueryCoordinatorService(context);
       addIfService(tajoMasterService);
-      
+
```
hyunsik: You need to recover meaningless changes.

tajo-core/src/main/java/org/apache/tajo/master/TajoMaster.java
```diff
@@ -236,7 +236,7 @@ private void initSystemMetrics() {
 
   private void initResourceManager() throws Exception {
     Class<WorkerResourceManager>  resourceManagerClass = (Class<WorkerResourceManager>)
-        systemConf.getClass(ConfVars.RESOURCE_MANAGER_CLASS.varname, TajoWorkerResourceManager.class);
+      systemConf.getClass(ConfVars.RESOURCE_MANAGER_CLASS.varname, TajoWorkerResourceManager.class);
```
hyunsik: You need to recover meaningless changes.

hyunsik:
As I mentioned in some comments, most of changes are about coding format. You need to recover them. Coding format should be done in another issue.

blrunner:
Thanks @hyunsik 
I recovered codes related to coding format.

hyunsik:
+1 LGTM!

PR #626, opened 2015-07-09
TAJO-1673: Implement recover partitions

blrunner:
This is still ongoing work. I'll do more tests and update DDLExecutor after committing TAJO-1345.


hyunsik:
Could you let me know the full name of msck?

The below expression seems to be more consistent form for a full command, I think.
```
ALTER TABLE [table name] REPAIR PARTITION
```

blrunner:
Hi @hyunsik 

I'just referred hive's policy. But your suggestion looks more reasonable now. 
I'll update the patch tonight. :)

blrunner:
I've updated MSCK TABLE statement to ALTER TABLE REPAIR TABLE statement.

blrunner:
I applied CatalogStore::addPartitions to DDLExecutor and implemented some unit test cases.
When I implementing unit test cases, I found that drop partition operate on different way than expected. I think that drop partition need to delete data of table with `PURGE` option. But currently, when dropping partition on managed table without `PURGE` option, all data had been deleted. So, I've fixed this bug.

blrunner:
I added codes to check existing partitions from CatalogStore before adding missing partitions.

blrunner:
I've finished partition repair test successfully with MySQLStore and HiveCatalogStore and fixed a few bugs.

blrunner:
Rebased.

tajo-core/src/main/java/org/apache/tajo/master/exec/DDLExecutor.java
```diff
@@ -23,29 +23,36 @@
...
 import org.apache.tajo.storage.Tablespace;
 import org.apache.tajo.storage.TablespaceManager;
 import org.apache.tajo.util.Pair;
+import org.apache.tajo.util.StringUtils;
+import org.apache.tajo.util.TUtil;
 
 import java.io.IOException;
 import java.util.ArrayList;
 import java.util.List;
+import java.util.Set;
```
jihoonson: Unused import.

tajo-core/src/main/java/org/apache/tajo/master/exec/DDLExecutor.java
```diff
@@ -529,24 +536,137 @@ public void alterTable(TajoMaster.MasterContext context, final QueryContext quer
         catalog.alterTable(CatalogUtil.addOrDropPartition(qualifiedName, alterTable.getPartitionColumns(),
             alterTable.getPartitionValues(), alterTable.getLocation(), AlterTableType.DROP_PARTITION));
 
-        // When dropping partition on an managed table, the data will be delete from file system.
-        if (!desc.isExternal()) {
+        // When dropping partition on a table, the data in the table will NOT be deleted from the file system.
+        // But if PURGE is specified, the partition data will be deleted.
```
jihoonson: How about changing the comment to as follows?
```When dropping a partition on a table, its data will NOT be deleted if the 'PURGE' option is not specified.```

File: tajo-core/src/main/java/org/apache/tajo/master/exec/DDLExecutor.java [hunk 3]
```diff
@@ -529,24 +536,137 @@ public void alterTable(TajoMaster.MasterContext context, final QueryContext quer
...
+    // Find missing partitions from filesystem
+    List<PartitionDescProto> existingPartitions = catalog.getPartitions(databaseName, simpleTableName);
+    List<String> existingPartitionNames = TUtil.newList();
+    Path existingPartitionPath = null;
+
+    for(PartitionDescProto existingPartition : existingPartitions) {
+      existingPartitionPath = new Path(existingPartition.getPath());
+      existingPartitionNames.add(existingPartition.getPartitionName());
+      if (!fs.exists(existingPartitionPath)) {
+        LOG.info("Partitions missing from Filesystem:" + existingPartition.getPartitionName());
```
jihoonson: I think ```warn``` is the more appropriate logging level.

tajo-core/src/main/java/org/apache/tajo/master/exec/DDLExecutor.java
```diff
@@ -529,24 +536,137 @@ public void alterTable(TajoMaster.MasterContext context, final QueryContext quer
...
+
+    if(tableDesc.getPartitionMethod() == null) {
+      throw new UndefinedPartitionMethodException(simpleTableName);
+    }
+
+    Path tablePath = new Path(tableDesc.getUri());
+    FileSystem fs = tablePath.getFileSystem(context.getConf());
+
+    PartitionMethodDesc partitionDesc = tableDesc.getPartitionMethod();
+    Schema partitionColumns = new Schema();
```
jihoonson: Why did you create a copy of the partition schema?

tajo-core/src/main/java/org/apache/tajo/master/exec/DDLExecutor.java
```diff
@@ -529,24 +536,137 @@ public void alterTable(TajoMaster.MasterContext context, final QueryContext quer
...
+        LOG.info("Partitions missing from Filesystem:" + existingPartition.getPartitionName());
+      }
+    }
+
+    // Find missing partitions from CatalogStore
+    List<PartitionDescProto> targetPartitions = TUtil.newList();
+    for(Path filteredPath : filteredPaths) {
+      PartitionDescProto targetPartition = getPartitionDesc(simpleTableName, filteredPath);
+      if (!existingPartitionNames.contains(targetPartition.getPartitionName())) {
+        LOG.info("Partitions not in CatalogStore:" + targetPartition.getPartitionName());
```
jihoonson: This will emit too many logs.

tajo-core/src/main/java/org/apache/tajo/master/exec/DDLExecutor.java
```diff
@@ -529,24 +536,137 @@ public void alterTable(TajoMaster.MasterContext context, final QueryContext quer
...
+        targetPartitions.add(targetPartition);
+      }
+    }
+
+    catalog.addPartitions(databaseName, simpleTableName, targetPartitions, true);
+
+    for(PartitionDescProto targetPartition: targetPartitions) {
+      LOG.info("Repair: Added partition to CatalogStore " + tableName + ":" + targetPartition.getPartitionName());
```
jihoonson: This will emit too many logs.

tajo-core/src/main/java/org/apache/tajo/master/exec/DDLExecutor.java
```diff
@@ -529,24 +536,137 @@ public void alterTable(TajoMaster.MasterContext context, final QueryContext quer
...
+
+    // loop from one to the number of partition columns
+    Path [] filteredPaths = PartitionedTableRewriter.toPathArray(fs.listStatus(tablePath, filters[0]));
+
+    // Get all file status matched to a ith level path filter.
+    for (int i = 1; i < partitionColumns.size(); i++) {
+      filteredPaths = PartitionedTableRewriter.toPathArray(fs.listStatus(filteredPaths, filters[i]));
+    }
+
+    // Find missing partitions from filesystem
```
jihoonson: The below routine is too complex and includes unnecessary codes, I think.
IMHO, the algorithm will be simple as follows.
1. Getting PartitionDescProtos from the catalog
2. While iterating PartitionDescProtos, call catalog.addPartition() if the partition directory exists. Maybe catalog.addPartitions() will be called instead.

If I missed any points, please let me know.

tajo-docs/src/main/sphinx/sql_language/alter_table.rst
```diff
@@ -96,4 +96,31 @@ You can use ``ALTER TABLE ADD PARTITION`` to add partitions to a table. The loca
   ALTER TABLE table1 DROP PARTITION (col1 = '2015' , col2 = '01', col3 = '11' )
   ALTER TABLE table1 DROP PARTITION (col1 = 'TAJO' ) PURGE
 
-You can use ``ALTER TABLE DROP PARTITION`` to drop a partition for a table. This removes the data for a managed table and this doesn't remove the data for an external table. But if ``PURGE`` is specified for an external table, the partition data will be removed. The metadata is completely lost in all cases. An error is thrown if the partition for the table doesn't exists. You can use ``IF EXISTS`` to skip the error.
+You can use ``ALTER TABLE DROP PARTITION`` to drop a partition for a table. This doesn't remove the data for a table. But if ``PURGE`` is specified, the partition data will be removed. The metadata is completely lost in all cases. An error is thrown if the partition for the table doesn't exists. You can use ``IF EXISTS`` to skip the error.
```
jihoonson: In ```if the partition for the table doesn't exists```, ```exists``` should be ```exist```.

tajo-docs/src/main/sphinx/sql_language/alter_table.rst
```diff
@@ -96,4 +96,31 @@ You can use ``ALTER TABLE ADD PARTITION`` to add partitions to a table. The loca
...
+
+.. code-block:: sql
+
+  ALTER TABLE <table_name> REPAIR PARTITION
+
+*Examples*
+
+.. code-block:: sql
+
+  ALTER TABLE student REPAIR PARTITION;
```
jihoonson: I think this example is so obvious, and thus unnecessary.

tajo-docs/src/main/sphinx/sql_language/alter_table.rst
```diff
@@ -96,4 +96,31 @@ You can use ``ALTER TABLE ADD PARTITION`` to add partitions to a table. The loca
...
+
+.. note::
+
+  If you recover partitions in the situation that partitions just exists on catalog and it doesn't exist on file system, Tajo would not make directories on the file system and would print messages to TajoMaster log as following:
```
jihoonson: How about changing this paragraph to as follows?
```
Even though an information of a partition is stored in the catalog, Tajo does not recover it when its partition directory doesn't exist in the file system.
```

tajo-core/src/main/java/org/apache/tajo/master/exec/DDLExecutor.java
jihoonson: I misunderstood the below codes. Sorry.

jihoonson:
+1. Thank you for your work!

blrunner:
Thanks @jihoonson 

I'll ship it at once. :)

PR #627, opened 2015-07-13
TAJO-1679: Client APIs should validate identifiers for database object names.

hyunsik:
See https://issues.apache.org/jira/browse/TAJO-1679.

For example, the current API accepts wrong object names as follows:
```java
client.createDatabase("12345");
```

This patch does not include the unit tests because the current error propagation system just causes exceptions in TajoMaster and cannot catch this kind of errors. This problem will be solved in TAJO-1670 and failure unit tests will be added in that jira too.

[Review on Reviewable](https://reviewable.io/reviews/apache/tajo/627)


hyunsik:
this is an invalid issue. I'll close it.

PR #630, opened 2015-07-14
TAJO-1346: Create dynamic partitions to CatalogStore by running insert query or CTAS query.

blrunner:
Currently, If new partitions are added to HDFS by ColPartitionStoreExec, CatalogStore will not be aware of these partitions. So, Tajo need to run ALTER TABLE table_name ADD PARTITION on each of the newly added partitions automatically.


blrunner:
PartitionKey would be removed after committing TAJO-1670. And I found an issue when testing unit test cases for partitioned table. It is that table status didn't been updated when inserting into a partitioned table, if same partition directories already exist. It should be done in another issue.

hyunsik:
This approach seems that each individual worker directly inserts created partitions through Catalog API. I'm concerning with too many connection problems of catalog and explosive INSERT requests simultaneously across many workers. It would be better if we can change it to a bulk insertion approach in QM. How do you think of this issue?

blrunner:
Thanks @hyunsik 
I agree with you and would update codes using your advice. :-)

blrunner:
Hi @hyunsik 

I've updated the patch using your comments and unit test cases finished successfully as follows.

```
mvn clean install -Dtajo.catalog.store.class=org.apache.tajo.catalog.store.HiveCatalogStore
mvn clean install -Dtajo.catalog.store.class=org.apache.tajo.catalog.store.DerbyStore
```

blrunner:
I tested to add partitions on MySQLStore as following:

* Data Set : TPC-H 1G
* Query Running Time
  * 10,000 partitions : 6 sec
  * 100,000 partitions: 50 sec
* Query
```
create table partitioned_lineitem3 (col1 text, col2 float8) 
 partition by column(col3 int4, col4 int4) as 
select l_returnflag, l_quantity, l_orderkey, l_partkey from lineitem
where l_orderkey < 10000;

create table partitioned_lineitem4 (col1 text, col2 float8) 
 partition by column(col3 int4, col4 int4) as 
select l_returnflag, l_quantity, l_orderkey, l_partkey from lineitem
where l_orderkey < 100000;
```

tajo-catalog/tajo-catalog-common/src/main/java/org/apache/tajo/catalog/exception/AssumedPartitionDirectoryException.java
```diff
@@ -18,9 +18,13 @@
...
 package org.apache.tajo.catalog.exception;
 
-public class NoPartitionedTableException extends Exception {
 
-  public NoPartitionedTableException(String databaseName, String relName) {
-    super(String.format("ERROR: table \"%s.%s\" is not a partitioned table", databaseName, relName));
-  }
+import org.apache.tajo.error.Errors.ResultCode;
+
+public class AssumedPartitionDirectoryException extends CatalogException {
```
hyunsik: Its name seems to be not intuitive. Could you explain which kind of exception?

tajo-catalog/tajo-catalog-common/src/main/java/org/apache/tajo/catalog/statistics/StatisticsUtil.java
```diff
@@ -83,6 +83,14 @@ public static void aggregateTableStat(TableStats result, TableStats stats) {
       }
     }
 
+    // If there is partitions
```
hyunsik: TableStats may be not a good place to contain added partitions because partition is not a part of statistic information. We need to improve it here or later. How do you think of it?

tajo-catalog/tajo-catalog-common/src/main/java/org/apache/tajo/catalog/statistics/TableStats.java
```diff
@@ -36,6 +39,8 @@
 import java.util.List;
 
 public class TableStats implements ProtoObject<TableStatsProto>, Cloneable, GsonObject {
+  private static Log LOG = LogFactory.getLog(TableStats.class);
```
hyunsik: Unused logger is added.

tajo-catalog/tajo-catalog-server/src/main/java/org/apache/tajo/catalog/store/AbstractDBStore.java
```diff
@@ -1245,116 +1242,118 @@ private void addNewColumn(int tableId, CatalogProtos.ColumnProto columnProto) th
...
-      CatalogUtil.closeQuietly(pstmt, res);
     }
-    return retValue;
   }
 
-  private void addPartitionKeys(PreparedStatement pstmt, int partitionId, PartitionDescProto partition) throws
-    SQLException {
-    for (int i = 0; i < partition.getPartitionKeysCount(); i++) {
-      PartitionKeyProto partitionKey = partition.getPartitionKeys(i);
+  public void addPartitionKeys(int tableId, CatalogProtos.PartitionDescProto partition,
```
hyunsik: StringBuilder parameter seems to be used to reuse StringBuilder object. I'd like to recommend that do not try code-level optimization until the code part actually causes bottleneck. This kind of optimization is not effective whereas it is harmful for code readability.

tajo-catalog/tajo-catalog-server/src/main/resources/schemas/derby/derby.xml
```diff
@@ -24,7 +24,7 @@
       * 2 - 2014-06-09: First versioning
       * 1-  Before 2013-03-20
```
hyunsik: This histories in all schema xml files should be updated too.

tajo-core/src/main/java/org/apache/tajo/engine/planner/physical/ColPartitionStoreExec.java
```diff
@@ -36,9 +38,12 @@
 import org.apache.tajo.plan.logical.StoreTableNode;
 import org.apache.tajo.storage.*;
 import org.apache.tajo.unit.StorageUnit;
+import org.apache.tajo.util.TUtil;
```
hyunsik: TUtil, URI, List are not used here.

tajo-core/src/main/java/org/apache/tajo/querymaster/QueryMaster.java
```diff
@@ -32,6 +32,12 @@
 import org.apache.tajo.QueryId;
 
 import org.apache.tajo.TajoProtos;
+import org.apache.tajo.catalog.CatalogService;
+import org.apache.tajo.catalog.CatalogUtil;
+import org.apache.tajo.catalog.TableDesc;
+import org.apache.tajo.catalog.partition.PartitionDesc;
```
hyunsik: not used.

tajo-core/src/main/java/org/apache/tajo/querymaster/QueryMaster.java
```diff
@@ -286,6 +292,47 @@ public void stopQuery(QueryId queryId) {
 
       finishedQueryMasterTasksCache.put(queryId, queryMasterTask);
 
+      // In INSERT .. SELECT ... queries or CTAS, dynamic partitions should be added to catalog.
+      Query query = queryMasterTask.getQuery();
+      if (query.getState() == TajoProtos.QueryState.QUERY_SUCCEEDED && query.getResultDesc() != null) {
```
hyunsik: Here is not good place to do this because we cannot guarantee the query atomicity. In other words, a query is successfully completed but the partition additions can be failed. Basically, we have to guarantee both the query completion and partition addition or nothing.

QueryCompletedTransition::finalizeQuery in Query is the best place for this.

blrunner:
Hi @hyunsik 

Thanks for your detailed review.
I've just updated the patch using your comments. :)

blrunner:
I added some columns and index to ``PARTITION_KEYS`` table for getting partitions efficiently and tested on macbook successfully. The batch response time results are listed below.

* 3,000 records: 1 ~ 2 sec
* 30,000 records: 6 ~ 8 sec
* 300,000 records: 60 ~ 70 sec

Honestly, I was worried about removing foreign key of ``PARTITION_KEYS`` table because it affected to batch insert performance. But it would be good to leave it on the table because of data integrity.

tajo-catalog/tajo-catalog-server/src/main/java/org/apache/tajo/catalog/CatalogServer.java
```diff
@@ -1065,6 +1068,47 @@ public GetTablePartitionsResponse getAllPartitions(RpcController controller, Nul
     }
 
     @Override
+    public ReturnState addPartitions(RpcController controller, AddPartitionsProto request)
+      throws ServiceException {
```
hyunsik: You can remove throws ServiceException.

tajo-catalog/tajo-catalog-common/src/main/proto/CatalogProtos.proto
```diff
@@ -181,6 +181,7 @@ message TableStatsProto {
   optional int64 readBytes = 7;
   repeated ColumnStatsProto colStat = 8;
   optional int32 tid = 9;
+  repeated PartitionDescProto partitions = 10;
```
hyunsik: Could you keep PartitionDescProto in TableStatsProto? Other messages like TaskCompletionReport may be a better place.

tajo-catalog/tajo-catalog-drivers/tajo-hive/src/main/java/org/apache/tajo/catalog/store/HiveCatalogStore.java
```diff
@@ -957,6 +1026,54 @@ private boolean existColumn(final String databaseName ,final String tableName ,
...
+    List<Partition> addPartitions = TUtil.newList();
+    CatalogProtos.PartitionDescProto existingPartition = null;
+
+    try {
+      client = clientPool.getClient();
+      for (CatalogProtos.PartitionDescProto partitionDescProto : partitions) {
+        existingPartition = getPartition(databaseName, tableName, partitionDescProto.getPartitionName());
+
+        // Unfortunately, hive client add_partitions doesn't run as expected. The method never read the ifNotExists
+        // parameter. So, if Tajo add existing partition to Hive, it will threw AlreadyExistsException. To avoid
```
hyunsik: s/adds/add/

tajo-catalog/tajo-catalog-drivers/tajo-hive/src/main/java/org/apache/tajo/catalog/store/HiveCatalogStore.java
```diff
@@ -957,6 +1026,54 @@ private boolean existColumn(final String databaseName ,final String tableName ,
...
+        // above error, we need filter existing partitions before call add_partitions.
```
hyunsik: ``need`` should be ``need to``.

tajo-catalog/tajo-catalog-server/src/main/java/org/apache/tajo/catalog/CatalogServer.java
```diff
@@ -1065,6 +1068,47 @@ public GetTablePartitionsResponse getAllPartitions(RpcController controller, Nul
...
+
+      TableIdentifierProto identifier = request.getTableIdentifier();
+      String databaseName = identifier.getDatabaseName();
+      String tableName = identifier.getTableName();
+
+
+      if (metaDictionary.isSystemDatabase(databaseName)) {
+        throw new ServiceException(databaseName + " is a system database. It does not contain any partitioned tables.");
```
hyunsik: Don't throw any exception. You can return proper ReturnState.

tajo-common/src/main/java/org/apache/tajo/conf/TajoConf.java
```diff
@@ -290,6 +290,10 @@ public static int setDateOrder(int dateOrder) {
     PYTHON_CODE_DIR("tajo.function.python.code-dir", ""),
     PYTHON_CONTROLLER_LOG_DIR("tajo.function.python.controller.log-dir", ""),
 
+    // Partition
+    PARTITION_BULK_INSERT_BATCH_SIZE("tajo.partition.bulk.insert.batch.size", 1000),
```
hyunsik: I'd like to recommend four levels as follows:
```
tajo.partition.dynamic.bulk-insert.batch-size
```

tajo-core/src/main/java/org/apache/tajo/querymaster/QueryMaster.java
```diff
@@ -288,6 +292,39 @@ public void stopQuery(QueryId queryId) {
         return;
       }
 
+      // Create dynamic partitions to CatalogStore by running insert query or CTAS query.
+      Query query = queryMasterTask.getQuery();
```
hyunsik: As I mentioned in previous comment, here is not good place for it. QueryCompletedTransition in Query could be a better place.

tajo-storage/tajo-storage-common/src/main/java/org/apache/tajo/storage/TableStatistics.java
```diff
@@ -59,6 +61,8 @@ public TableStatistics(Schema schema) {
         comparable[i] = true;
       }
     }
+
+    stat = new TableStats();
```
hyunsik: It may be a mistake.

tajo-catalog/tajo-catalog-server/src/main/resources/schemas/derby/derby.xml
blrunner: Thanks, I already added the history to all schema xml files.

blrunner:
Hi @hyunsik 

Thanks for your detailed review.
I've just updated my patch using your comments. :-)

hyunsik:
+1 The latest patch looks good to me. Thank you!

blrunner:
Thanks @hyunsik ,
I've just committed it to the master branch. :-)

PR #631, opened 2015-07-16
TAJO-1685: Print enhanced message when using table data which located on local file system occasionally on fully distributed mode.

blrunner:
I added a message for avoiding users confusion as follows.

```
default> \admin -cluster
Query Master
============

Live  Dead  Tasks
----- ----- -----
3     0     0    

Live QueryMasters
=================

QueryMaster               Port  Query Heap       Status    
------------------------- ----- ----- ---------- ----------
tajo-04:28093 28092 0     16000 MB   RUNNING   
tajo-03:28093 28092 0     16000 MB   RUNNING   
tajo-05:28093 28092 0     16000 MB   RUNNING   


Worker
======

Live  Dead 
----- -----
3     0    

Live Workers
============

Worker                    Port  Tasks Mem        Disk       Heap         Status    
------------------------- ----- ----- ---------- ---------- ------------ ----------
tajo-04:28091 53960 0     0/1024     0.00/2.00  427/16000 MB RUNNING   
tajo-03:28091 34833 0     0/1024     0.00/2.00  418/16000 MB RUNNING   
tajo-05:28091 40578 0     0/1024     0.00/2.00  412/16000 MB RUNNING   


Dead Workers
============

No Dead Workers

default> \d table1;

table name: default.table1
table uri: file:/home/tajo/hadoop/data.csv
store type: text
number of rows: unknown
volume: 60 B
Options: 
	'text.delimiter'='|'

schema: 
id	INT4
name	TEXT
score	FLOAT4
type	TEXT


default> select * from table1;
id,  name,  score,  type
1,  abc,  1.1,  a
2,  def,  2.3,  b
3,  ghi,  3.4,  c
4,  jkl,  4.5,  d
5,  mno,  5.6,  e
(5 rows, 0.029 sec, 60 B selected)

default> select count(*) from table1;
ERROR: The table data should be on all hosts to run TajoWorker or be on distributed file system. : file:/home/tajo/hadoop/data.csv

```


jihoonson:
I think that replicating local data on ever worker is not a good solution. Users may suffer invalid result.

jihoonson:
Rather than doing that, I think the query should be processed by a single machine, mostly tajo master, if we should support local file system as our storage.

blrunner:
Hi @jihoonson 

Replicating local data to all worker don't look well. So I updated the message for exception.
And I also think that supporting local file system is a fundamental solution. But we need to put more efforts to implement it and this patch would be a temporary work before providing local file system.

PR #634, opened 2015-07-17
TAJO-1343: Improve the memory usage of physical executors

jihoonson:
Here are some highlights.
* Cleanup duplicated tuple creation.
* To maintain tuple creation easily, I've added three new data structures, called TupleMap, TupleList, and TupleSet. They automatically clone tuples whenever a new item is added.
* Fix a bug in WindowAggregation and two wrong tests.
* Fix a bug in timezone.

I will conduct some tests and share the results.

[Review on Reviewable](https://reviewable.io/reviews/apache/tajo/634)


jihoonson:
I've tested this patch on my laptop using YourKit.
Here are some highlights of the results.

### Query and data
I used the following query on TPC-H data set with the scale factor of 1.
```
default> select count(distinct l_partkey) from lineitem;
```

### 1. Tuple creation
The following numbers were randomly captured during executing the second phase of distinct aggregation. So, it would not be the exact comparison, but is valuable I think.

#### Master
| Class        | Objects    | Shallow Size  | Retained Size |
| ------------- |:-------------:| ------------------:| -------------------:|
|org.apache.tajo.storage.VTuple | 3,899,275 | 93,582,600 | 112,627,520 |
|org.apache.tajo.datum.Datum[] | 3,899,275 | 87,335,008 | 96,857,392 |

#### Tajo-1343
| Class        | Objects    | Shallow Size  | Retained Size |
| ------------- |:-------------:| ------------------:| -------------------:|
|org.apache.tajo.engine.planner.physical.KeyTuple | 2,573,412 | 82,349,184 | 95,913,472 |
|org.apache.tajo.storage.VTuple | 398,023 | 9,552,552 | 9,552,984 |
|org.apache.tajo.datum.Datum[] | 2,971,439 | 71,314,440 | 78,096,672 |

As can be seen in the above result, the total size of generated tuples in TAJO-1343 is less than that in master. The difference is 7,161,064.

### 2. Memory usage
The following graphs show the changes of memory usage during query execution.

##### < Memory usage of with master >
memory usage of with master
##### < Memory usage of with TAJO-1343 >
memory usage of with TAJO-1343

The gray part in each graph represents the change of memory usage during query execution.
As can be seen in the graphs, memory usage change with TAJO-1343 is more gracefully than that with master.

The following numbers were captured when the query is finished because it's the time when the numbers become the maximum. 

| branch name | Allocated all pools        | Used PS Eden Space    | Used PS Survivor Space  | Used PS Old Gen | # of GCs |
| ------------- | ------------- |:-------------:| ------------------:| -------------------:| -------------------:|
| master | 1006 | 634 | 22 | 63 | 4 |
| TAJO-1343 | 1002 | 518 | 43 | 71 | 3 |

The amount of ```allocated all pools``` with TAJO-1343 is similar to that of master. However, ```used PS eden space``` with TAJO-1343 is about 100 MB less than that with master. This means that the less objects are newly created during query execution. 

In this test, the numbers of GCs with both branches are similar, so it is of less significance. However, I think that the difference in number of GCs will be increased with more complex queries, thereby more significantly affecting to the query performance.

tajo-common/src/main/java/org/apache/tajo/storage/VTuple.java
```diff
@@ -100,7 +101,9 @@ public int size(int fieldId) {
     return values[fieldId].size();
   }
 
+  @Override
   public void put(Datum [] values) {
+    this.clear();
```
jinossy: Would you please explain why clear() need to call ?

jinossy:
@jihoonson 
This patch looks great!
If you can reduce to invoke the clear(), it would be better

jihoonson:
@jinossy thanks for your review. 
I've added ```clear()``` in ```put()``` method because I was considering unexpected errors when the datum array of different length is put.
As you think, this may accompany large overhead. I will remove it after some test.

jihoonson:
I've added a configuration ```tajo.executor.memory-tuple-slot-num``` to set the memory tuple slot.

jihoonson:
I've reverted the last changes because it makes unexpected errors.

jinossy:
+1 Looks great to me!!

tajo-common/src/main/java/org/apache/tajo/storage/VTuple.java
```diff
@@ -82,6 +83,7 @@ public void clear() {
   // Setter
   //////////////////////////////////////////////////////
   public void put(int fieldId, Datum value) {
+    this.offset = -1;
```
hyunsik: Does every put needs the offset initialization?

hyunsik:
+1
The patch looks good to me. I leaved one trivial issue. It's up to you.

In addition, you added the removal code in QueryContext.
https://github.com/apache/tajo/pull/634/files#diff-e0e47538675376e2e0b96ce307a9676bR49

I think that it would be handled in other places. But, the current workaround looks good to me. I'll solve it in another jira.

Thanks!

hyunsik:
+1 ship it!

jihoonson:
Thanks guys for your review.
I've just committed to master after removing implicit clear of offset per put() call.

PR #635, opened 2015-07-17
TAJO-1690: Create table using text file fails in HiveCatalogStore.

blrunner:
I added text file format to HiveCatalogStore.


tajo-catalog/tajo-catalog-drivers/tajo-hive/src/test/java/org/apache/tajo/catalog/store/TestHiveCatalogStore.java
```diff
@@ -59,6 +59,7 @@
 
   private static HiveCatalogStore store;
   private static Path warehousePath;
+  private final static String[] storages = {"CSV", "TEXT"};
```
jinossy: Could you change to textStoreTypes ?

tajo-catalog/tajo-catalog-drivers/tajo-hive/src/test/java/org/apache/tajo/catalog/store/TestHiveCatalogStore.java
```diff
@@ -85,34 +86,36 @@ public static void tearDown() throws IOException {
...
-    schema.addColumn("c_name", TajoDataTypes.Type.TEXT);
-    schema.addColumn("c_address", TajoDataTypes.Type.TEXT);
-    schema.addColumn("c_nationkey", TajoDataTypes.Type.INT4);
-    schema.addColumn("c_phone", TajoDataTypes.Type.TEXT);
-    schema.addColumn("c_acctbal", TajoDataTypes.Type.FLOAT8);
-    schema.addColumn("c_mktsegment", TajoDataTypes.Type.TEXT);
-    schema.addColumn("c_comment", TajoDataTypes.Type.TEXT);
-
-    TableDesc table = new TableDesc(CatalogUtil.buildFQName(DB_NAME, CUSTOMER), schema, meta,
+    for (String storage: storages) {
```
hyunsik: I don't think that we need all unit tests for each storage type. You just add OR condition for TEXT and CSV and all other codes are the same. You need to add an unit test about your change.

blrunner:
Thanks @hyunsik and @jinossy 

I've recovered existing test cases and add just a test case for this issue.

hyunsik:
The change in this patch seems to be duplicated to #638.

blrunner:
@hyunsik 

OK. I'll close this PR. :-)

PR #638, opened 2015-07-21
TAJO-1496: Remove legacy CSVFile

jinossy:
I've disable the TestTajoCli.testTimeZoneTest2. because it has a meta overwriting bug


hyunsik:
#634 adds some workaround of timezone bug. So, you may need to enable the test after you commit your patch.

jinossy:
I’m going to enable timezone unit test after TAJO-1343 is closed. All unit test is succeed and I will commit it

PR #643, opened 2015-07-23
TAJO-1597: Problem of ignoring theta join condition

jihoonson:
[Review on Reviewable](https://reviewable.io/reviews/apache/tajo/643)


hyunsik:
The filters seem to be higher than sort. It may cause sorting of unnecessary rows.

jihoonson:
Thanks! It was my mistake. I've fixed.

hyunsik:
It seems to include in and exist subquery implementation.

tajo-plan/src/main/java/org/apache/tajo/plan/algebra/BaseAlgebraVisitor.java
```diff
@@ -52,221 +52,221 @@ public RESULT visit(CONTEXT ctx, Stack<Expr> stack, Expr expr) throws TajoExcept
...
-    case TimestampLiteral:
-      current = visitTimestampLiteral(ctx, stack, (TimestampLiteral) expr);
-      break;
-    case IntervalLiteral:
-      current = visitIntervalLiteral(ctx, stack, (IntervalLiteral) expr);
-      break;
-
-    default:
-      throw new TajoInternalError("Cannot support this type algebra \"" + expr.getType() + "\"");
+      case SetSession:
```
hyunsik: The wrong coding format seems to be applied. {{switch}} and {{case}} should be the same column position.

jihoonson:
Damn. It's my mistake.
I'll make another pr.

PR #644, opened 2015-07-23
TAJO-1700: Add better exception handling in TajoMasterClientService.

hyunsik:
This change Improve TajoMasterClientService to print out stacktrace only if the exception is a real error without error code.


hyunsik:
I just fixed one wrong implementation in existSessionVariable in client side.

jihoonson:
+1. Recent your work around error propagation is really nice.
Thanks for your effort.

PR #646, opened 2015-07-24
TAJO-1597: Problem of ignoring theta join condition

jihoonson:
[Review on Reviewable](https://reviewable.io/reviews/apache/tajo/646)


tajo-plan/src/main/java/org/apache/tajo/plan/rewrite/rules/FilterPushDownRule.java
```diff
@@ -235,9 +232,50 @@ public LogicalNode visitJoin(FilterPushDownContext context, LogicalPlan plan, Lo
     }
 
     context.pushingDownFilters.removeAll(matched);
+
+    if (context.pushingDownFilters.size() > 0) {
```
hyunsik: It would be great if you add some comments to explain its logical meaning about the condition and what this code do.

hyunsik:
+1
The patch looks good to me. I leave one comment. If you agree, you can commit if you add some comments.

jihoonson:
@hyunsik thank you for the quick review.
I'll commit after leaving some explanation according to your comment!

PR #652, opened 2015-07-24
TAJO-1465: Add ORCFileAppender to write into ORCFile table

eminency:
Hi all.

This is for writer code for ORC files.

Many files are added, but most of them are from Hive. See following notes :

1. Core working is in ORCAppender.java and WriterImpl.java
2. Files in org.apache.tajo.storage.orc.objectinspector are trivial. Those are for checking data type.
3. Files in org.apache.tajo.storage.thirdparty are from Hive. WriterImpl.java is also from Hive, but it is modified to support Tajo types and Datum.

Please review and advise.
Thanks.

[Review on Reviewable](https://reviewable.io/reviews/apache/tajo/652)


hyunsik:
I'm reviewing this patch. I'll leave comments soon.

tajo-storage/tajo-storage-hdfs/src/main/java/org/apache/tajo/storage/thirdparty/orc/WriterImpl.java
```diff
@@ -0,0 +1,2251 @@
...
+      // stripe statistics
+      OrcProto.StripeStatistics.Builder stripeStatsBuilder = OrcProto.StripeStatistics.newBuilder();
+      writeStripeStatistics(stripeStatsBuilder, this);
+      stripeStatsBuilders.add(stripeStatsBuilder);
+
+      // reset the flag for next stripe
+      foundNulls = false;
+
+      builder.addColumns(getEncoding());
+      if (streamFactory.hasWriterTimeZone()) {
```
hyunsik: This timezone policy may be different from those of Tajo. Could you check the timezone? The following codes would be helpful to this work.

https://github.com/apache/tajo/blob/master/tajo-storage/tajo-storage-hdfs/src/main/java/org/apache/tajo/storage/text/TextFieldSerializerDeserializer.java#L56

hyunsik:
The patch looks good to me. I leave one comment about timezone.

Also, this patch borrows a part of orc implementation. So, we need to continue upgrading the borrowed orc implementation. Could you add some comments to point out borrowed ones or our own implementation?

eminency:
Hi @hyunsik .

I modified what you mentioned.
Please review again, thanks.

hyunsik:
It seems to need the query tests with timezone. Actually, I'm not able to ensure that this change and ORC file guarantee the correct timezone support like Text file. It can be proved from some unit tests. Could you add unit tests for ORC? You can see 5 types of tests in TestSelectQuery class.  Right now, you can add the tests into TestSelectQuery. Those timezone tests should be moved into the better place later.

eminency:
It seems timezone processing has some bug.
See https://issues.apache.org/jira/browse/TAJO-1741.
Currently writing the test is blocked.

eminency:
Hi, @hyunsik.

I modified orc timezone code and added an corresponding unit test similar to existing testTimezonedTable() tests.

Please review it and leave a comment.

tajo-common/src/main/java/org/apache/tajo/datum/DatumFactory.java
```diff
@@ -114,45 +114,42 @@ public static Datum createFromString(DataType dataType, String value) {
...
+        return createInet4(bytes);
+      case PROTOBUF:
+        try {
+          return ProtobufDatumFactory.createDatum(dataType, bytes);
+        } catch (IOException e) {
+          e.printStackTrace();
+          throw new RuntimeException(e);
+        }
+      default:
         throw new UnsupportedOperationException(dataType.toString());
```
hyunsik: It should be UndefinedOperatorException wrapped by TajoRuntimeException. You need to take a look at ErrorMessage.java in order to know how it makes the error message for UndefinedOperatorException.

I also think that UnsupportedException is also proper here.

tajo-storage/tajo-storage-hdfs/src/main/java/org/apache/tajo/storage/orc/ORCScanner.java
```diff
@@ -152,11 +155,13 @@ public void init() throws IOException {
 
     orcReader = new OrcReader(orcDataSource, new OrcMetadataReader());
 
+    TimeZone timezone = TimeZone.getTimeZone(meta.getOption(StorageConstants.TIMEZONE,
+      TajoConstants.DEFAULT_SYSTEM_TIMEZONE));
+
     // TODO: make OrcPredicate useful
-    // TODO: TimeZone should be from conf
-    // TODO: it might be splittable
+    // presto-orc uses joda timezone, so it needs to be converted.
```
hyunsik: Could you explain what is Joda timezone? It seems to be just a java timezone.

hyunsik:
It borrows many codes from ORC. It looks the best now. But, we need to make maintenance easy. How do you think separating the borrowed code into a separate maven module while keeping the same java package? It makes us to easily identify the copied and modified ones.

tajo-common/src/main/java/org/apache/tajo/datum/DatumFactory.java
eminency: I see. I'm working on it.

eminency:
I refined throwing exception code and added a list for modified files in package-info.java.
Please check them out.

hyunsik:
Thank. Could you rebase the patch against the latest revision?

tajo-cluster-tests/src/test/java/org/apache/tajo/QueryTestCaseBase.java
```diff
@@ -913,7 +913,8 @@ private Path getDataSetFile(String fileName) throws IOException {
           throw new IOException("Cannot find " + fileName + " at " + currentQueryPath + " and " + namedQueryPath);
         }
       } else {
-        throw new IOException("Cannot find " + fileName + " at " + currentQueryPath + " and " + namedQueryPath);
+        // to make empty table (used to insert test)
```
hyunsik: This code will make a directory in a resource directory. It seems to be not good. Could you remove this change?

tajo-common/src/main/java/org/apache/tajo/datum/DatumFactory.java
```diff
@@ -71,7 +71,7 @@
       case NULL_TYPE:
         return NullDatum.class;
       default:
-        throw new UnsupportedOperationException(type.name());
+        throw new TajoRuntimeException(Errors.ResultCode.UNSUPPORTED_DATATYPE, type.name());
```
hyunsik: It should be ```new TajoRuntimeException(new UnsupportedDataType(...)```.

tajo-common/src/main/java/org/apache/tajo/datum/DatumFactory.java
```diff
@@ -107,53 +107,50 @@ public static Datum createFromString(DataType dataType, String value) {
     case INET4:
       return createInet4(value);
     default:
-      throw new UnsupportedOperationException(dataType.toString());
+      throw new TajoRuntimeException(Errors.ResultCode.UNSUPPORTED_DATATYPE, dataType.toString());
```
hyunsik: It should be ```new TajoRuntimeException(new UnsupportedDataType(...)```.

tajo-common/src/main/java/org/apache/tajo/datum/DatumFactory.java
```diff
@@ -107,53 +107,50 @@ public static Datum createFromString(DataType dataType, String value) {
...
-      try {
-        builder.mergeFrom(bytes);
-        return factory.createDatum(builder.build());
-      } catch (IOException e) {
-        e.printStackTrace();
-        throw new RuntimeException(e);
-      }
-    default:
-        throw new UnsupportedOperationException(dataType.toString());
+      case BOOLEAN:
```
hyunsik: SWITCH and CASE should be in the same column position.

tajo-common/src/main/java/org/apache/tajo/datum/DatumFactory.java
```diff
@@ -107,53 +107,50 @@ public static Datum createFromString(DataType dataType, String value) {
...
+        return createInet4(bytes);
+      case PROTOBUF:
+        try {
+          return ProtobufDatumFactory.createDatum(dataType, bytes);
+        } catch (IOException e) {
+          e.printStackTrace();
+          throw new TajoRuntimeException(Errors.ResultCode.IO_ERROR, e.getMessage());
+        }
+      default:
+        throw new TajoRuntimeException(Errors.ResultCode.UNSUPPORTED_DATATYPE, dataType.toString());
```
hyunsik: It should be new TajoRuntimeException(new UnsupportedDataType(...).

tajo-common/src/main/java/org/apache/tajo/datum/DatumFactory.java
```diff
@@ -107,53 +107,50 @@ public static Datum createFromString(DataType dataType, String value) {
...
+        return createBit(bytes[0]);
+      case BLOB:
+        return createBlob(bytes);
+      case INET4:
+        return createInet4(bytes);
+      case PROTOBUF:
+        try {
+          return ProtobufDatumFactory.createDatum(dataType, bytes);
+        } catch (IOException e) {
+          e.printStackTrace();
```
hyunsik: Don't use e.printStackTrace. Exceptions will print the proper log message in some place. This exception is not normal exception. So, it should be ```throw new TajoInternalError(e)```.

tajo-common/src/main/java/org/apache/tajo/datum/DatumFactory.java
```diff
@@ -164,7 +161,7 @@ public static Datum createFromInt4(DataType type, int val) {
     case DATE:
       return new DateDatum(val);
     default:
-      throw new UnsupportedOperationException("Cannot create " + type.getType().name() + " datum from INT4");
+      throw new TajoRuntimeException(Errors.ResultCode.UNSUPPORTED_DATATYPE, type.getType().name());
```
hyunsik: It should be ```new TajoRuntimeException(new UnsupportedDataType(...)```.

tajo-common/src/main/java/org/apache/tajo/datum/DatumFactory.java
```diff
@@ -177,7 +174,7 @@ public static Datum createFromInt8(DataType type, long val) {
     case TIME:
       return createTime(val); 
     default:
-      throw new UnsupportedOperationException("Cannot create " + type.getType().name() + " datum from INT8");
+      throw new TajoRuntimeException(Errors.ResultCode.UNSUPPORTED_DATATYPE, type.getType().name());
```
hyunsik: It should be ```new TajoRuntimeException(new UnsupportedDataType(...)```.

tajo-common/src/main/java/org/apache/tajo/datum/DatumFactory.java
```diff
@@ -333,7 +330,7 @@ public static DateDatum createDate(Datum datum) {
     case DATE:
       return (DateDatum) datum;
     default:
-      throw new InvalidCastException(datum.type(), Type.DATE);
+      throw new TajoRuntimeException(Errors.ResultCode.INVALID_CAST, datum.type().toString(), Type.DATE.name());
```
hyunsik: It should be ```new TajoRuntimeException(new InvalidCastException(...)```.

tajo-common/src/main/java/org/apache/tajo/datum/DatumFactory.java
```diff
@@ -352,7 +349,7 @@ public static TimeDatum createTime(Datum datum, @Nullable TimeZone tz) {
     case TIME:
       return (TimeDatum) datum;
     default:
-      throw new InvalidCastException(datum.type(), Type.TIME);
+      throw new TajoRuntimeException(Errors.ResultCode.INVALID_CAST, datum.type().name(), Type.TIME.name());
```
hyunsik: It should be ```new TajoRuntimeException(new InvalidCastException(...)```.

tajo-common/src/main/java/org/apache/tajo/datum/DatumFactory.java
```diff
@@ -365,7 +362,7 @@ public static TimestampDatum createTimestamp(Datum datum, @Nullable TimeZone tz)
       case TIMESTAMP:
         return (TimestampDatum) datum;
       default:
-        throw new InvalidCastException(datum.type(), Type.TIMESTAMP);
+        throw new TajoRuntimeException(Errors.ResultCode.INVALID_CAST, datum.type().name(), Type.TIMESTAMP.name());
```
hyunsik: It should be ```new TajoRuntimeException(new InvalidCastException(...)```.

tajo-common/src/main/java/org/apache/tajo/datum/DatumFactory.java
```diff
@@ -462,7 +459,7 @@ public static Datum cast(Datum operandDatum, DataType target, @Nullable TimeZone
     case ANY:
       return DatumFactory.createAny(operandDatum);
     default:
-      throw new InvalidCastException(operandDatum.type(), target.getType());
+      throw new TajoRuntimeException(Errors.ResultCode.INVALID_CAST, operandDatum.type().name(), target.getType().name());
```
hyunsik: It should be ```new TajoRuntimeException(new InvalidCastException(...)```.

tajo-storage/tajo-storage-hdfs/src/test/java/org/apache/tajo/storage/orc/TestOrc.java
```diff
@@ -39,18 +47,29 @@
 
 import java.io.IOException;
 import java.net.URL;
+import java.util.List;
 
-public class TestORCScanner {
+public class TestOrc {
```
hyunsik: Does this class tests more things that TestStorages cannot cover?

hyunsik:
I leaved comments.

tajo-common/src/main/java/org/apache/tajo/datum/DatumFactory.java
eminency: @hyunsik 
InvalidCastException is not subclass of TajoException, thus it can be used.
Should it be changed?

tajo-storage/tajo-storage-hdfs/src/test/java/org/apache/tajo/storage/orc/TestOrc.java
eminency: It can be removed.
When I make the test, I didn't know about TestStorages.

hyunsik:
+1

Before commiting the patch, I slightly changed as follows
 * ``INVALID_CAST`` to ``INVALID_VALUE_FOR_CAST``
 * Add SQLState for ``INVALID_VALUE_FOR_CAST`` to ``SQLExceptionUtil``
 * Add an map entry for ``INVALID_VALUE_FOR_CAST`` to ``ExceptionUtil``

hyunsik:
Thank you for your work. I committed as c3c78fc2147768595051b8cfb1006f13ca2db1bf. But, I missed to add github command in the commit log. Could you close this ticket?

eminency:
Thank you very much for your working.
I close this issue at final.

PR #653, opened 2015-07-25
TAJO-1675: NPE when selecting data from information_schema.partition_keys

blrunner:
I added getAllPartitionKeys to CatalogStore and added some codes to NonForwardQueryResultSystemScanner. If you test it on tsql, you never reproduce NPE. But you can get any rows from partition_keys table because current tajo doesn't provide loading dynamic partitions by using insert query and CTAS query. You will get partition keys from information schema after resolving https://github.com/apache/tajo/pull/630 .


tajo-catalog/tajo-catalog-common/src/main/proto/CatalogProtos.proto
```diff
@@ -152,6 +152,12 @@ message TablePartitionProto {
   optional string path = 4;
 }
 
+message TablePartitionKeysProto {
```
jihoonson: A message represents a single partition. So, ```TablePartitionKeyProto``` looks to be more proper.

tajo-catalog/tajo-catalog-server/src/main/java/org/apache/tajo/catalog/CatalogServer.java
```diff
@@ -1230,6 +1230,29 @@ public ReturnState addPartitions(RpcController controller, AddPartitionsProto re
     }
 
     @Override
+    public GetTablePartitionKeysResponse getAllPartitionKeys(RpcController controller,
+                                                   NullProto request) throws ServiceException {
```
jihoonson: Please fix indent.

tajo-core/src/test/java/org/apache/tajo/engine/query/TestAlterTable.java
```diff
@@ -103,10 +107,79 @@ public final void testAlterTableAddPartition() throws Exception {
     assertTrue(fs.exists(partitionPath));
     assertTrue(partitionPath.toString().indexOf("col3=1/col4=2") > 0);
 
+    List<CatalogProtos.DatabaseProto> allDatabases = catalog.getAllDatabases();
```
jihoonson: Please move this test to ```TestNonForwardQueryResultSystemScanner```.

tajo-catalog/tajo-catalog-server/src/main/java/org/apache/tajo/catalog/store/AbstractDBStore.java
```diff
@@ -2122,8 +2122,10 @@ private void setPartitionKeys(int pid, PartitionDescProto.Builder partitionDesc)
     List<TablePartitionProto> partitions = new ArrayList<TablePartitionProto>();
 
     try {
-      String sql = "SELECT " + COL_PARTITIONS_PK + ", " + COL_TABLES_PK + ", PARTITION_NAME, " +
-        " PATH FROM " + TB_PARTTIONS;
+      String sql = "SELECT A." + COL_PARTITIONS_PK + ", A." + COL_TABLES_PK + ", A.PARTITION_NAME, " +
+        " A.PATH FROM " + TB_PARTTIONS + " A JOIN " + TB_TABLES + " B "
```
jihoonson: Since it is guaranteed that the tables corresponding to partitions always exist, join looks not necessary.

tajo-catalog/tajo-catalog-server/src/main/java/org/apache/tajo/catalog/store/AbstractDBStore.java
```diff
@@ -2259,6 +2261,41 @@ public void addPartitions(String databaseName, String tableName, List<CatalogPro
...
+    Connection conn = null;
+    Statement stmt = null;
+    ResultSet resultSet = null;
+
+    List<TablePartitionKeysProto> partitions = new ArrayList<TablePartitionKeysProto>();
+
+    try {
+      String sql = " SELECT C." + COL_PARTITIONS_PK + ", C.COLUMN_NAME, C.PARTITION_VALUE " +
+        " FROM " + TB_TABLES + " A " +
+        " JOIN " + TB_PARTTIONS + " B ON A." + COL_TABLES_PK + " = B." + COL_TABLES_PK
```
jihoonson: Since it is guaranteed that the tables corresponding to partitions always exist, join looks not necessary.

tajo-catalog/tajo-catalog-server/src/main/java/org/apache/tajo/catalog/store/MemStore.java
```diff
@@ -579,10 +578,19 @@ public void dropPartitionMethod(String databaseName, String tableName) throws Ca
...
+    int tableId = 0, partitionId = 0;
+    List<TableDescriptorProto> tables = getAllTables();
     List<TablePartitionProto> protos = new ArrayList<TablePartitionProto>();
-    Set<String> tables = partitions.keySet();
-    for (String table : tables) {
-      Map<String, CatalogProtos.PartitionDescProto> entryMap = partitions.get(table);
+
+    Set<String> partitionTables = partitions.keySet();
+    for (String partitionTable : partitionTables) {
+      for (TableDescriptorProto table : tables) {
```
jihoonson: Since it is guaranteed that the tables corresponding to partitions always exist, it looks not necessary to match table names.

jihoonson:
@blrunner, thanks for your work. Please consider my comments.

blrunner:
Hi @jihoonson 

I've just updated it using your comments.

tajo-core-tests/src/test/java/org/apache/tajo/engine/query/TestAlterTable.java
```diff
@@ -103,10 +107,33 @@ public final void testAlterTableAddPartition() throws Exception {
     assertTrue(fs.exists(partitionPath));
     assertTrue(partitionPath.toString().indexOf("col3=1/col4=2") > 0);
 
+    boolean existPartition = false;
+    List<CatalogProtos.TablePartitionProto> allPartitions = catalog.getAllPartitions();
```
jihoonson: As I said before, this test should be moved to TestNonForwardQueryResultSystemScanner. (Maybe this class is also renamed, for example to TestInformationSchema). Here are some reasons. 
First, this class is for testing alter table operation rather than functionality of information schema. Second, in this implementation, several tests are mixed together, so it is difficult to figure out which lines test which functionality. Unit testing is to test the functionality of individual modules.

tajo-catalog/tajo-catalog-client/src/main/proto/CatalogProtocol.proto
```diff
@@ -89,6 +89,11 @@ message GetTablePartitionsResponse {
   repeated TablePartitionProto part = 2;
 }
 
+message GetTablePartitionKeysResponse {
+  required ReturnState state = 1;
+  repeated TablePartitionKeyProto partKey = 2;
```
jihoonson: ```partitionKey``` looks more intuitive.

tajo-catalog/tajo-catalog-server/src/main/java/org/apache/tajo/catalog/CatalogServer.java
```diff
@@ -1044,6 +1044,29 @@ public ReturnState addPartitions(RpcController controller, AddPartitionsProto re
     }
 
     @Override
+    public GetTablePartitionKeysResponse getAllPartitionKeys(RpcController controller, NullProto request) throws
+      ServiceException {
```
jihoonson: Please fix the indent between ```throws``` and ```ServiceException```.

tajo-catalog/tajo-catalog-server/src/main/java/org/apache/tajo/catalog/store/AbstractDBStore.java
```diff
@@ -2138,8 +2138,10 @@ private void setPartitionKeys(int pid, PartitionDescProto.Builder partitionDesc)
     List<TablePartitionProto> partitions = new ArrayList<TablePartitionProto>();
 
     try {
-      String sql = "SELECT " + COL_PARTITIONS_PK + ", " + COL_TABLES_PK + ", PARTITION_NAME, " +
-        " PATH FROM " + TB_PARTTIONS;
+      String sql = "SELECT A." + COL_PARTITIONS_PK + ", A." + COL_TABLES_PK + ", A.PARTITION_NAME, A.PATH "
+        + " FROM " + TB_PARTTIONS + " A "
+        + " WHERE A." + COL_PARTITIONS_PK + " > 0 "
+        + " AND A." + COL_TABLES_PK + " > 0 ";
```
jihoonson: ```COL_PARTITIONS_PK``` and ```COL_TABLES_PK``` are primary keys and thus cannot have negative values.

tajo-catalog/tajo-catalog-server/src/main/java/org/apache/tajo/catalog/store/AbstractDBStore.java
```diff
@@ -2277,6 +2279,41 @@ public void addPartitions(String databaseName, String tableName, List<CatalogPro
...
+    Connection conn = null;
+    Statement stmt = null;
+    ResultSet resultSet = null;
+
+    List<TablePartitionKeyProto> partitions = new ArrayList<TablePartitionKeyProto>();
+
+    try {
+      String sql = " SELECT A." + COL_PARTITIONS_PK + ", A.COLUMN_NAME, A.PARTITION_VALUE " +
+        " FROM " + TB_PARTTION_KEYS + " A " +
+        " WHERE A." + COL_PARTITIONS_PK + " > 0 "+
```
jihoonson: What does ```COL_PARTITIONS_PK > 0``` mean?

tajo-catalog/tajo-catalog-server/src/main/java/org/apache/tajo/catalog/store/AbstractDBStore.java
```diff
@@ -2277,6 +2279,41 @@ public void addPartitions(String databaseName, String tableName, List<CatalogPro
...
+        " AND A.COLUMN_NAME IS NOT NULL " +
```
jihoonson: ```COLUMN_NAME``` is already defined with ```NOT NULL```.

tajo-catalog/tajo-catalog-server/src/main/java/org/apache/tajo/catalog/store/AbstractDBStore.java
```diff
@@ -2277,6 +2279,41 @@ public void addPartitions(String databaseName, String tableName, List<CatalogPro
...
+        " AND A.PARTITION_VALUE IS NOT NULL ";
```
jihoonson: Why are partition keys filtered if they have null partition value?

tajo-catalog/tajo-catalog-common/src/main/proto/CatalogProtos.proto
```diff
@@ -152,6 +152,12 @@ message TablePartitionProto {
   optional string path = 4;
 }
 
+message TablePartitionKeyProto {
+  required int32 partition_id = 1;
+  required string columnName = 2;
+  required string partitionValue = 3;
```
jihoonson: Partition value can be null.

tajo-catalog/tajo-catalog-common/src/main/proto/CatalogProtos.proto
```diff
@@ -152,6 +152,12 @@ message TablePartitionProto {
   optional string path = 4;
 }
 
+message TablePartitionKeyProto {
```
jihoonson: How is it different from ```PartitionKeyProto```?

tajo-catalog/tajo-catalog-server/src/main/java/org/apache/tajo/catalog/store/MemStore.java
```diff
@@ -598,6 +598,27 @@ public boolean existPartitionMethod(String databaseName, String tableName)
...
+    for (String partitionTable : partitionTables) {
+      Map<String, CatalogProtos.PartitionDescProto> entryMap = partitions.get(partitionTable);
+      for (Map.Entry<String, CatalogProtos.PartitionDescProto> proto : entryMap.entrySet()) {
+        CatalogProtos.PartitionDescProto partitionDescProto = proto.getValue();
+
+        for (PartitionKeyProto partitionKey : partitionDescProto.getPartitionKeysList()) {
+          TablePartitionKeyProto.Builder builder = TablePartitionKeyProto.newBuilder();
+          builder.setColumnName(partitionKey.getColumnName());
+          builder.setPartitionValue(partitionKey.getPartitionValue());
+          builder.setPartitionId(0);
```
jihoonson: partition id is always 0.

tajo-catalog/tajo-catalog-common/src/main/proto/CatalogProtos.proto
blrunner: PartitionKeyProto already exists.

jihoonson: Yes. I mean, what is the difference between PartitionKeyProto and TablePartitionKeyProto? Their name is similar, but contents are different. I wonder their proper usage.

blrunner:
Thanks @jihoonson 
A little while ago, as a fix again. :)

tajo-catalog/tajo-catalog-common/src/main/proto/CatalogProtos.proto
```diff
@@ -247,13 +240,15 @@ message PartitionDescProto {
   required string partitionName = 1;
   repeated PartitionKeyProto partitionKeys = 2;
   optional string path = 3;
-  optional int32 id = 4;
+  optional int32 partition_id = 4;
```
jihoonson: Partition id is the primary key of the ```partitions``` table. Why is it optional?

tajo-catalog/tajo-catalog-common/src/main/proto/CatalogProtos.proto
```diff
@@ -247,13 +240,15 @@ message PartitionDescProto {
...
+  optional int32 tid = 5;
```
jihoonson: tid is also mandatory field.

tajo-catalog/tajo-catalog-common/src/main/proto/CatalogProtos.proto
```diff
@@ -247,13 +240,15 @@ message PartitionDescProto {
   required string partitionName = 1;
   repeated PartitionKeyProto partitionKeys = 2;
   optional string path = 3;
```
jihoonson: Path also looks mandatory field. Is there any case that path is null? 
If not, it would be great if you also change it to ```required```.

tajo-catalog/tajo-catalog-common/src/main/proto/CatalogProtos.proto
```diff
@@ -247,13 +240,15 @@ message PartitionDescProto {
...
-  optional int32 id = 4;
+  optional int32 partition_id = 4;
+  optional int32 tid = 5;
 }
 
 message PartitionKeyProto {
   required string columnName = 1;
   optional string parentColumnName = 2;
```
jihoonson: parentColumnName is unused. Please remove it.

tajo-core-tests/src/test/java/org/apache/tajo/engine/query/TestAlterTable.java
```diff
@@ -25,6 +25,7 @@
 import org.apache.tajo.catalog.CatalogUtil;
 import org.apache.tajo.catalog.TableDesc;
 import org.apache.tajo.catalog.proto.CatalogProtos;
+import org.apache.tajo.util.TUtil;
```
jihoonson: Unused import.

tajo-core-tests/src/test/java/org/apache/tajo/engine/query/TestAlterTable.java
```diff
@@ -35,6 +36,7 @@
 
 @Category(IntegrationTest.class)
 public class TestAlterTable extends QueryTestCaseBase {
+
```
jihoonson: Meaningless change.

tajo-core-tests/src/test/java/org/apache/tajo/master/TestNonForwardQueryResultSystemScanner.java
```diff
@@ -18,9 +18,22 @@
 
 package org.apache.tajo.master;
 
+import org.apache.hadoop.fs.FileSystem;
+import org.apache.hadoop.fs.Path;
```
jihoonson: Unused imports.

tajo-core-tests/src/test/java/org/apache/tajo/master/TestNonForwardQueryResultSystemScanner.java
```diff
@@ -38,4 +51,89 @@ public void testGetNextRowsForTable() throws Exception {
...
+        && (partitionKey.getColumnName().equals("col3") && partitionKey.getPartitionValue().equals("1")
+        || partitionKey.getColumnName().equals("col4") && partitionKey.getPartitionValue().equals("2"))) {
+        resultPartitionKeys.add(partitionKey);
+      }
+    }
+    assertEquals(resultPartitionKeys.size(), 2);
+    assertEquals(resultPartitionKeys.get(0).getColumnName(), "col3");
+    assertEquals(resultPartitionKeys.get(0).getPartitionValue(), "1");
+    assertEquals(resultPartitionKeys.get(1).getColumnName(), "col4");
+    assertEquals(resultPartitionKeys.get(1).getPartitionValue(), "2");
```
jihoonson: Above tests looks duplicate with below tests. Only difference is that above tests use catalog client APIs while below tests use tajo client APIs. For consistency, I prefer to use tajo client APIs in this test class. 

In addition, I think that we need to test the case when information of a lot of partitions and partition keys is stored in catalog. So, it would be great if more partitions are added when initializing.

tajo-catalog/tajo-catalog-common/src/main/proto/CatalogProtos.proto
blrunner: When adding or dropping partition, CatalogClient calls alter table API with AlterTableDescProto. and the proto maintains partition informations to PartitionDescProto. When adding partition, the partition path will be not null. But when dropping partition, the path will be null because Tajo can delete partition informs in CatalogStore with partition name.

tajo-catalog/tajo-catalog-common/src/main/proto/CatalogProtos.proto
blrunner: PartitionDescProto is used for various catalog API, for examples, output parameter of getPartitionByPartitionName and getPartitionsByTableName and getAllPartitions, input parameter of alterTable and addPartitions. In case of output parameter, partition id always is not null. But in case of input parameter, partition id will be null because CatalogClient can get partition informs with table name and partition name.

tajo-catalog/tajo-catalog-common/src/main/proto/CatalogProtos.proto
blrunner: Please see above comments.

blrunner:
Thanks @jihoonson 
I updated the patch using your comments and leaved some comments about your questions.

tajo-catalog/tajo-catalog-common/src/main/proto/CatalogProtos.proto
jihoonson: This interface is very weird because it involves many potential bugs. We must handle the cases that some mandatory fields are missed, and this is naturally done by protocol buffer. The below problem of optionally declared path is also related to this weird design. 
A better design is as follows. PartitionDescProto corresponds to PartitionDesc, and should be used to represent a partition as TableDesc represents a table. 
To do so, PartitionDescProto should be used as only the result of the methods which retrieves partitions. For input params, a new data structure is required.

jihoonson:
@blrunner, thanks for update. I left a comment, but I know that my comment is not directly related to your patch. You may have two choices. One is updating the patch to improve the interface design, and the other is booking it as another issue in Jira. If you choose the second option, it should be resolved before our upcoming release, I think. 
Thanks.

blrunner:
@jihoonson 

Thank your for your kind reply. I've updated the patch using the first choice. :)

blrunner:
Sorry @jihoonson 

Honestly, I had made unnecessary adjustments to catalog protocol. As a result, we lost unnecessary times. So, I've removed unnecessary modification.

jihoonson:
Selecting partition keys is not necessary because partitions are searched with its partition name. It would be better to remove this function.

blrunner:
@jihoonson 

Okay, I'll remove it from this function at another issue.

PR #654, opened 2015-07-25
TAJO-1713: Change the type of edge cache in JoinGraphContext from HashMap to LRUMap

jihoonson:
I've also cleared variables which are unnecessary after join optimization.

[Review on Reviewable](https://reviewable.io/reviews/apache/tajo/654)


hyunsik:
LRUMap has different initial sizes according to whether it is used in tests or not. Is it necessary or is it enough for tests?

jihoonson:
It is enough for tests. According to my experience, the cache size is increased up to about 120 entries while inner joining 5 tables. So, I think it is enough to set the max cache size as 100 during tests.

jihoonson:
I've removed the codes for adjusting map size during tests because the overhead of map size seems not significant.

hyunsik:
LGTM! +1

jihoonson:
Thanks for the review!

PR #657, opened 2015-07-27
TAJO-1712: querytasks.jsp throws NPE occasionally when tasks are running.

jinossy:
Current implementation opens the all file in order to get the total size.
So I've remove the pagination in query page and add to the next link


jihoonson:
+1 LGTM!

PR #658, opened 2015-07-28
TAJO-1715: Precompute the hash value of various kinds of ids

jihoonson:
Changes are:
* Statically maintaining the hash value in TaskId and TaskAttemptId
 * This is because, unlike QueryId and ExecutionBlockId, it is expected there are many tasks and their attempts.
* Improving TupleMap.put()
 * Copying tuple is expensive, so I implemented the clone() method for KeyTuple.

[Review on Reviewable](https://reviewable.io/reviews/apache/tajo/658)


jinossy:
+1 LGTM

jihoonson:
Thanks for your review!

PR #660, opened 2015-07-29
TAJO-1332 Implement Python Client for Tajo

charsyam:
This is implement of python tajo rest client.

### DIFF
--- a/contrib/tajo-python/README.md
+++ b/contrib/tajo-python/README.md
@@ -1,0 +1,5 @@
+# python-tajo-rest-client
+Python Tajo Rest Client Module
+
+# requirements
+pip install httplib2
--- a/contrib/tajo-python/setup.py
+++ b/contrib/tajo-python/setup.py
@@ -1,0 +1,42 @@
+# Licensed to the Apache Software Foundation (ASF) under one or more
+# contributor license agreements.  See the NOTICE file distributed with
+# this work for additional information regarding copyright ownership.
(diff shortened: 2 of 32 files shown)


charsyam:
This is a PR just for discuss more.
we should discuss about testing and distribution and directory of this kind of clients.

I just put the python client into contrib/tajo-python
but it should be discussed for right position.

hyunsik:
Great work! I think that ``contrib`` seems to be a right position. I'll review it soon.

jihoonson:
Nice work! On the proper location, I agree with you guys.

hyunsik:
I'm sorry for late review. The protocol between client and TajoMaster has been rapidly changed for better error propagation (https://issues.apache.org/jira/browse/TAJO-1625). This changes will actually affect REST API specification. I thought that this patch is conflict to the on-going error propagation work.

The solution is that we have to improve REST API to reflect the error propagation before we can continue this Python client. I think that REST API improvements would be trivial. We can put this python client into the next release.

blrunner:
Hi @charsyam 

Python has an API to make it easy to move from one SQL-based DB to another called DB-API. Each DB may have multiple different modules available for it. And if python users use it, they can easy integrate with ORM. It's documented in https://www.python.org/dev/peps/pep-0249.

eminency:
Hi, @charsyam 

May I take this on to continue?

charsyam:
@eminency Thanks for your suggestion.
but, If you excuse. I will finish it ASAP
Thanks.

PR #664, opened 2015-07-29
TAJO-1711: Add basic monitoring page

yeongeon:
Add basic monitoring page


jinossy:
@hyunsik 
nv.d3.js license seems to apache 2.0. Should we add d3 license to LICENSE file ?
https://github.com/novus/nvd3/blob/master/LICENSE.md

tajo-core/src/main/resources/webapps/admin/header.jsp
```diff
@@ -26,6 +26,7 @@
     <li><a class='top_menu_item' href='query.jsp'>Query</a></li>
     <li><a class='top_menu_item' href='catalogview.jsp'>Catalog</a></li>
     <li><a class='top_menu_item' href='query_executor.jsp'>Execute Query</a></li>
+    <li><a class='top_menu_item' href='monitoring.jsp'>Basic Monitoring</a></li>
```
jihoonson: Simply ```Monitoring``` looks more appropriate because we don't provide any complex or advanced monitoring.

hyunsik:
@jinossy We don't need to add d3 license to LICENSE file if it is under ASF 2.0.

hyunsik:
For clear, it would be better if you add this license (https://github.com/novus/nvd3/blob/master/LICENSE.md) into LICENSE file.

hyunsik:
Hi @yeongeon,

This work looks great to me. But, it includes per-node monitoring feature. But, it is not fit in Tajo web page because Tajo basically aims at a number of nodes. 

If you want to put a single page monitoring to Tajo, it would be great if the page is a summary page to give an overview of cluster, query, and scheduling information. 

So, this work needs more work to remove worker parts and add some more information for an overview of them. For this, it would be great if you continue this work in a separate branch because 0.11.0 release is approaching.

I'll look forward to your comments. Thanks.

PR #667, opened 2015-07-30
TAJO-1725: TaskContainer can hang when RuntimeException occurs in TaskImpl

hyunsik:
[Review on Reviewable](https://reviewable.io/reviews/apache/tajo/667)

### DIFF
--- a/tajo-core/src/main/java/org/apache/tajo/worker/TaskContainer.java
+++ b/tajo-core/src/main/java/org/apache/tajo/worker/TaskContainer.java
@@ -66,7 +66,8 @@ public class TaskContainer implements Runnable {
         }
 
         task.cleanup();
-      } catch (Exception e) {
+
+      } catch (Throwable e) {
         LOG.error(e.getMessage(), e);
         if (task != null) {
           try {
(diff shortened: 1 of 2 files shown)


tajo-core/src/main/java/org/apache/tajo/worker/TaskImpl.java
```diff
@@ -400,10 +401,14 @@ public void run() throws Exception {
...
       }
     } catch (Throwable e) {
+      // clean up TaskImpl resource
       error = e ;
-      LOG.error(e.getMessage(), e);
       stopScriptExecutors();
       context.stop();
+
+      // rethrow the exception in order to propagate the abnormal status to TaskContainer
+      throw new TajoInternalError(e);
```
jinossy: If an exception causes in physical operator, it can be sent failure report in finally block. But rethrow the exception are sent report twice

tajo-core/src/main/java/org/apache/tajo/worker/TaskContainer.java
```diff
@@ -66,7 +66,8 @@ public void run() {
         }
 
         task.cleanup();
-      } catch (Exception e) {
+
+      } catch (Throwable e) {
```
jinossy: Nice catch!!

jinossy:
@hyunsik 
Query hang is fixed by TAJO-1731. Could you test again?

hyunsik:
This bug seems to be fixed in TAJO-1731. I close this issue. Thank you :)

jinossy:
@hyunsik 
TAJO-1726 is caused by rethrow the exception
Thanks.

PR #668, opened 2015-07-30
TAJO-1723: INSERT INTO statement should allow nested fields as target columns.

hyunsik:
This patch improve INSERT INTO statement to allow nested fields as target columns as follows:

``
CREATE TABLE X (rowkey TEXT, cf1 RECORD (v1 int, v2 int));
INSERT INTO X (rowkey, cf1.v1, cf2.v1) SELECT x, y, z FROM t2
``


hyunsik:
TestSelectNestedRecord contains some insertion tests. Its purpose is not matched to the class name. I wanted to merge all unit tests related to nested fields into one test. Later, I'll move them into more proper one.

jinossy:
+1 Looks great to me

PR #670, opened 2015-07-31
TAJO-1732: During filter push down phase, join conditions are not set properly

jihoonson:
[Review on Reviewable](https://reviewable.io/reviews/apache/tajo/670)


hyunsik:
+1 LGTM

jihoonson:
Thanks for your review.

PR #671, opened 2015-07-31
TAJO-1340: Change the default output file format.

jinossy:


tajo-core/src/main/java/org/apache/tajo/engine/planner/global/GlobalPlanner.java
```diff
@@ -163,7 +164,7 @@ public void build(QueryContext queryContext, MasterPlan masterPlan) throws IOExc
   private static void setFinalOutputChannel(DataChannel outputChannel, Schema outputSchema) {
     outputChannel.setShuffleType(NONE_SHUFFLE);
     outputChannel.setShuffleOutputNum(1);
-    outputChannel.setStoreType("TEXT");
+    outputChannel.setStoreType(BuiltinStorages.DRAW);
```
jihoonson: How about adding a final variable for the default store type and using it in the class?

tajo-core/src/main/java/org/apache/tajo/engine/planner/global/GlobalPlanner.java
```diff
@@ -991,7 +992,7 @@ public LogicalNode visitProjection(GlobalPlanContext context, LogicalPlan plan,
         for (DataChannel dataChannel : masterPlan.getIncomingChannels(execBlock.getId())) {
           // This data channel will be stored in staging directory, but RawFile, default file type, does not support
           // distributed file system. It needs to change the file format for distributed file system.
-          dataChannel.setStoreType("TEXT");
+          dataChannel.setStoreType(BuiltinStorages.DRAW);
```
jihoonson: How about adding a final variable for the default store type and using it in the class?

jihoonson:
@jinossy, this patch looks good to me, but it causes some minor conflicts against master. Please fix it. 
In addition, I left a very trivial comment. If you agree, please consider it.

hyunsik:
It needs rebase.

jinossy:
Thanks Guys for the review! 
But I found hangs in OffheapBlock. I am investigating the hangs in TAJO-1738
If TAJO-1738 is ready, I will do rebase

tajo-core/src/main/java/org/apache/tajo/engine/planner/global/GlobalPlanner.java
jinossy: I think that adding FINAL_OUTPUT_FORMAT in TajoConf is better.

jinossy:
I've tested JDBC performance on laptop because my testing network environment is 1Gbps
TPC-H scale-3 lineitem
Query: select * from lineitem where 1=1

Before
```
Serialize per row + Text
107 sec //avg 30MByte/sec
```
After
```
Serialize Row-Block + Text
80 sec 
Serialize Row-Block + DRAW
30 sec
```

jinossy:
I've add async fetch.
Here is some benchmark
```
//Query: select * from lineitem
TajoMaster: text + TajoClient: async-fetch
81 sec (Deserialize), 76 sec (Non-Deserialize)

TajoMaster: text compress + TajoClient: async-fetch , decompress
75 sec (Deserialize), 76 sec (Non-Deserialize)

//Query: select * from lineitem where 1=1
TajoMaster: draw + TajoClient: async-fetch
35 sec (Deserialize), 25 sec (Non-Deserialize)

TajoMaster: draw, compress + TajoClient: async-fetch, decompress
53 sec (Deserialize), 51 sec (Non-Deserialize)
```

jinossy:
Change to pre-fetch both TajoMaster and TajoClient

Pre-Fetch, Network: Loopback, Data: lineitem 3GB
```
TEXT format
52 sec (Deserialize), 52 sec (Non-Deserialize)

TEXT format + snappy compression
54 sec (Deserialize), 54 sec (Non-Deserialize)

DRAW format
32 sec (Deserialize), 22 sec (Non-Deserialize)

DRAW format + snappy compression
47 sec (Deserialize), 47 sec (Non-Deserialize)
```

jinossy:
This PR is ready for review

hyunsik:
I've started reviewing this patch.

tajo-client/src/main/java/org/apache/tajo/jdbc/FetchResultSet.java
```diff
@@ -25,11 +25,13 @@
 
 import java.io.IOException;
 import java.sql.SQLException;
+import java.util.concurrent.Future;
 
 public class FetchResultSet extends TajoResultSetBase {
   protected QueryClient tajoClient;
   private int fetchRowNum;
   private TajoMemoryResultSet currentResultSet;
+  private Future<TajoMemoryResultSet> nextResultSet;
```
jihoonson: It looks that ```nextResultSet``` should also be closed when FetchResultSet.close() is called.

tajo-client/src/main/java/org/apache/tajo/client/QueryClient.java
```diff
@@ -108,6 +109,8 @@
 
   TajoMemoryResultSet fetchNextQueryResult(final QueryId queryId, final int fetchRowNum) throws TajoException;
```
jihoonson: It looks that this method doesn't need to be exposed as a client interface.

jihoonson:
Nice work! I left some trivial comments.

jinossy:
@jihoonson 
I've update the patch that reflects your comments

tajo-client/src/main/java/org/apache/tajo/client/QueryClient.java
```diff
@@ -106,7 +107,7 @@
 
   GetQueryResultResponse getResultResponse(QueryId queryId) throws TajoException;
 
-  TajoMemoryResultSet fetchNextQueryResult(final QueryId queryId, final int fetchRowNum) throws TajoException;
+  Future<TajoMemoryResultSet> asyncFetchNextQueryResult(final QueryId queryId, final int fetchRowNum);
```
hyunsik: How about renaming it into ``fetchNextQueryResultAsync()``?

It would be more consistent for a method name, starting a verb. So, the method's full meaning will be 'fetch the next query result asynchronously'.

tajo-core-tests/src/test/resources/results/TestGroupByQuery/testDistinctAggregation_case10.result
```diff
@@ -1,3 +1,3 @@
 ?sum,?sum_1
 -------------------------------
-3,414440.9
\ No newline at end of file
+3,414440.89999999997
```
hyunsik: What is the reason why different precision is used here?

jinossy: I tested in IDE debug tool
The precision is changed in SumContext and OffheapWriter keeps number with precision.

tajo-client/src/main/java/org/apache/tajo/client/QueryClient.java
jinossy: You're right. Thanks

tajo-common/src/main/java/org/apache/tajo/SessionVars.java
```diff
@@ -149,6 +149,8 @@
       Integer.class, Validators.min("0")),
   BLOCK_ON_RESULT(ConfVars.$RESULT_SET_BLOCK_WAIT, "Whether to block result set on query execution", DEFAULT,
       Boolean.class, Validators.bool()),
+  RESULT_COMPRESS(ConfVars.$RESULT_SET_COMPRESS, "Enable resultSet compression", CLI_SIDE_VAR,
```
hyunsik: How about renaming it to RESULT_COMPRESSION?

tajo-common/src/main/java/org/apache/tajo/conf/TajoConf.java
```diff
@@ -377,6 +381,7 @@ public static int setDateOrder(int dateOrder) {
       // ResultSet ---------------------------------------------------------
     $RESULT_SET_FETCH_ROWNUM("tajo.resultset.fetch.rownum", 200),
     $RESULT_SET_BLOCK_WAIT("tajo.resultset.block.wait", true),
+    $RESULT_SET_COMPRESS("tajo.resultset.compress", false),
```
hyunsik: How about ``tajo.resultset.compression`` ?

tajo-core-tests/src/test/resources/results/TestTajoCli/testHelpSessionVars.result
```diff
@@ -42,4 +42,5 @@ Available Session Variables:
 \set ARITHABORT [true or false] - If true, a running query will be terminated when an overflow or divide-by-zero occurs.
 \set FETCH_ROWNUM [int value] - Sets the number of rows at a time from Master
 \set BLOCK_ON_RESULT [true or false] - Whether to block result set on query execution
+\set RESULT_COMPRESS [true or false] - Enable resultSet compression
```
hyunsik: I have some comments.

* ``RESULT_COMPRESSION`` would be better in terms of naming convention.
* Probably, ``RESULT_COMPRESS`` and its description may seem to enable the compressed query result. 
* Actually, this feature compresses the transfer stream for ResultSet. So, I'd like to suggest the description as ``Use compression to optimize result transmission." 
* Also, I'd like to suggest this key ``COMPRESSED_RESULT_TRENSFER`` for session variable. It's clear, but it is somewhat longer than yours. It's up to you.

Also, I'm going to improve JDBC URL to take defined options instead of taking directly session variables. For compressed data transfer, I'll add ``?useCompression=true`` like MySQL.

tajo-core/src/main/java/org/apache/tajo/engine/planner/physical/SeqScanExec.java
```diff
@@ -198,7 +201,34 @@ public void init() throws IOException {
       // for non-projected fields.
       Schema actualInSchema = scanner.isProjectable() ? projectedFields : inSchema;
 
-      this.projector = new Projector(context, actualInSchema, outSchema, plan.getTargets());
+      Target[] realTargets;
```
hyunsik: It would be better if this change is extracted to one separate method like ``initializeProjector``.

tajo-common/src/main/proto/DataTypes.proto
```diff
@@ -114,3 +114,7 @@ message DataType {
    */
   optional int32 num_nested_fields = 4;
 }
+
+enum CodecType {
```
hyunsik: DataTypes.proto seems to be not proper for CodecType, which is used for client communication.

tajo-common/src/main/java/org/apache/tajo/conf/TajoConf.java
```diff
@@ -377,6 +381,7 @@ public static int setDateOrder(int dateOrder) {
       // ResultSet ---------------------------------------------------------
     $RESULT_SET_FETCH_ROWNUM("tajo.resultset.fetch.rownum", 200),
```
hyunsik: When we use compression, ``200`` may be too small to get some performance benefits from compression. Later, we may need some experiments to find the best row number for compression or non-compression mode.

tajo-client/src/main/java/org/apache/tajo/jdbc/TajoMemoryResultSet.java
```diff
@@ -18,46 +18,78 @@
...
                              Map<String, String> clientSideSessionVars) {
     super(queryId, schema, clientSideSessionVars);
-    this.totalRow = maxRowNum;
-    this.serializedTuples = serializedTuples;
-    this.decoder = RowStoreUtil.createDecoder(schema);
+    if(resultSet != null && resultSet.getRows() > 0) {
+      this.totalRow = resultSet.getRows();
+
+      try {
+        // decompress if has a codec
```
hyunsik: It should be ``decompress if it has a codec`` or ``decompress if a codec is specified``.

tajo-common/src/main/java/org/apache/tajo/conf/TajoConf.java
```diff
@@ -215,6 +216,9 @@ public static int setDateOrder(int dateOrder) {
     SHUFFLE_HASH_APPENDER_PAGE_VOLUME("tajo.shuffle.hash.appender.page.volumn-mb", 30),
     HASH_SHUFFLE_PARENT_DIRS("tajo.hash.shuffle.parent.dirs.count", 10),
 
+    // Final output Configuration --------------------------------------------------
+    FINAL_OUTPUT_FILE_FORMAT("tajo.final.output.file-format", BuiltinStorages.TEXT, Validators.javaString()),
```
hyunsik: Each key level should be meaningful, and left keys should be more general than those of right ones. 
I'd like to recommend ``DEFAULT_OUTPUT_FILE_FORMAT`` and ``tajo.output.file-format``.

hyunsik:
Excellent work! I leaved some comments.

tajo-common/src/main/java/org/apache/tajo/conf/TajoConf.java
jinossy: @hyunsik 
I think it should be a size. If a row size is too big, probably it can be OOM

tajo-common/src/main/proto/DataTypes.proto
jinossy: I'll move CodecType to `tajo_protos.proto`

tajo-core-tests/src/test/resources/results/TestTajoCli/testHelpSessionVars.result
jinossy: Thanks for your detailed review
"useCompression=true" looks great to me. also I’m going to change to COMPRESSED_RESULT_TRENSFER for session variable

tajo-common/src/main/java/org/apache/tajo/conf/TajoConf.java
hyunsik: Yes, you are right. Too big rows will cause OOM. If we find proper default row number, it will result in better compression ratio.

tajo-common/src/main/java/org/apache/tajo/conf/TajoConf.java
```diff
@@ -377,6 +381,7 @@ public static int setDateOrder(int dateOrder) {
       // ResultSet ---------------------------------------------------------
     $RESULT_SET_FETCH_ROWNUM("tajo.resultset.fetch.rownum", 200),
     $RESULT_SET_BLOCK_WAIT("tajo.resultset.block.wait", true),
+    $RESULT_SET_COMPRESSION("tajo.resultset.compression", false),
```
hyunsik: It also should be ``$COMPRESSED_RESULT_TRANSFER`` for consistency of ``COMPRESSED_RESULT_TRANSFER`` in SessionVars.

hyunsik:
Here is my +1. The patch looks great to me. I leaved one trivial comment. You can commit it after you reflect my comment if you agree.

tajo-common/src/main/java/org/apache/tajo/conf/TajoConf.java
jinossy: I've change to `$COMPRESSED_RESULT_TRANSFER`
Thanks

PR #672, opened 2015-07-31
TAJO-1733: Finished query occasionally does not appear in Web-UI

jinossy:


hyunsik:
+1

The patch looks good to me. Ship it.

PR #676, opened 2015-08-05
TAJO-1742: Remove hadoop dependency in DatumFactory

jinossy:


hyunsik:
+1
The patch looks straightforward.

PR #678, opened 2015-08-05
TAJO-1743: Improve calculation of intermediate table statistics

jinossy:


tajo-core/src/main/java/org/apache/tajo/engine/planner/physical/RangeShuffleFileWriteExec.java
```diff
@@ -56,14 +57,19 @@
   private KeyProjector keyProjector;
 
   public RangeShuffleFileWriteExec(final TaskAttemptContext context,
-                                   final PhysicalExec child, final Schema inSchema, final Schema outSchema,
-                                   final SortSpec[] sortSpecs) throws IOException {
-    super(context, inSchema, outSchema, child);
+                                   final ShuffleFileWriteNode plan,
```
blrunner: Could you explain why you remove output schema from parameters of the constructor?

jinossy:
You can find it in PhysicalPlannerImpl. in-out schema is same in RangeShuffleWriter

blrunner:
Thanks @jinossy 

I found following codes using your comments:
``` 
return new RangeShuffleFileWriteExec(ctx, subOp, plan.getInSchema(), plan.getInSchema(), sortSpecs); 
```

blrunner:
+1

LGTM. :)

PR #685, opened 2015-08-12
TAJO-1561: Query which contains join condition in "OR" clause does not finish.

jihoonson:
I added a new query rewriting rule. I've also tested the TPC-DS query reported at Jira.

[Review on Reviewable](https://reviewable.io/reviews/apache/tajo/685)


hyunsik:
The patch looks straightforward. Here is my +1.

jihoonson:
Thanks for the review!

PR #690, opened 2015-08-14
TAJO-1761: Separate an integration unit test kit into an independent module.

hyunsik:
This patch simply moves some packages into separate modules. In detail, it does as follows:
 * add tajo-cluster-tests and tajo-core-tests modules
 * Move TajoTestingCluster, QueryTestBase, ExprTestBase, BackendTestingUtil, HBaseTestingCluster, IntegrationTest, TpchTestBase into tajo-cluster-tests module.
 * Move all unit tests in tajo-core into tajo-core-tests

They are at all.


hyunsik:
rebased.

jihoonson:
Thanks @hyunsik. This patch is simple and good. I have two comments.
* Please fix the indent of BackendTestingUtil.
* core/pom.xml should not include dependencies on packages required for tests.

jihoonson:
+1 LGTM!

PR #691, opened 2015-08-16
TAJO-1763: tpch/*.tbl files cannot be founded in maven modules except for core-tests.

hyunsik:
I also attached the patch to the jira.
https://issues.apache.org/jira/browse/TAJO-1763

[Review on Reviewable](https://reviewable.io/reviews/apache/tajo/691)


hyunsik:
Do not review this issue. I found more essential problem. It would be better if we remove legacy org.apache.tajo.benchmark package and improve QueryTestBase.

PR #693, opened 2015-08-16
TAJO-1744: Porting bash shell scripts to Windows command shell scripts.

hys9958:
Hi,
I have ported to CMD the bash shell.
And i attach pdf to explain about getting started for Windows Tajo to ticket(TAJO-1744) of Apache tajo.
https://issues.apache.org/jira/secure/attachment/12750713/tajo_cmd.pdf

- conf
    - tajo-env.cmd
- bin
    - start-tajo.cmd
    - stop-tajo.cmd
    - tajo-daemon.cmd
    - tajo.cmd
    - tajo-config.cmd
    - tsql.cmd

### DIFF
--- a/tajo-dist/src/main/bin/start-tajo.cmd
+++ b/tajo-dist/src/main/bin/start-tajo.cmd
@@ -1,0 +1,34 @@
+@echo off
+@rem Licensed to the Apache Software Foundation (ASF) under one or more
+@rem contributor license agreements.  See the NOTICE file distributed with
+@rem this work for additional information regarding copyright ownership.
+@rem The ASF licenses this file to You under the Apache License, Version 2.0
+@rem (the "License"); you may not use this file except in compliance with
+@rem the License.  You may obtain a copy of the License at
+@rem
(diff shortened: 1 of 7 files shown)


jihoonson:
@hys9958, thanks for your work. Would you trigger ci again?

hys9958:
@jihoonson, of couse.
I have updated a tajo.cmd file to add some log message.
So, May be run CI.

dkhwangbo:
@hys9958 
Thanks for your effort.
I'm trying to check suitability of your work
and some inappropriate operation is occured like below picture.
capture01
after quit of tsql, some other word is printed.
Please check this part.

jihoonson:
Honestly, it seems that this problem is not directly related to this ticket. It would be better to book at another jira, I think.

dkhwangbo:
Testing this commit is done. I can run tajo in Windows 7 via above files.
I also test tsql.cmd via TPC-H set in same environment. TAJO can run all query of this set.

jihoonson:
@dkhwangbo, thank you for your test. 
Here is my +1.
@hys9958, thank you for your work.

PR #694, opened 2015-08-17
TAJO-1775: HCatalogStore need to be deprecated.

blrunner:
If old version users use tajo 0.11.0 using existing hcatalog configuration, they will fail to run TajoMaster with ClassNotFoundException. So, we need to deprecate HCatalogStore.


hyunsik:
I think that the unit test seems to be not necessary because it is striaghtforward.

blrunner:
Thanks @hyunsik 
I've removed it. :-)

jihoonson:
Looks good to me.

hyunsik:
+1

PR #698, opened 2015-08-18
TAJO-1781: Join condition is still not found when it exists in OR clause

jihoonson:
The problem is some hard code in CommonConditionReduceRule.
I've fixed this problem and added more tests.

[Review on Reviewable](https://reviewable.io/reviews/apache/tajo/698)


tajo-plan/src/main/java/org/apache/tajo/plan/rewrite/rules/CommonConditionReduceRule.java
```diff
@@ -162,58 +162,72 @@ protected EvalNode visitDefaultFunctionEval(Object context, Stack<EvalNode> stac
     private EvalNode rewrite(BinaryEval evalNode) {
       // Example qual: ( a | b ) ^ ( a | c )
```
hyunsik: Even through it is not change in this patch, it would be great if you change it to more intuitive example.

jihoonson:
Thanks. I've fixed comments.

hyunsik:
+1
The patch looks good to me, and the examples included in comments look comprehensive. Also, the unit tests are enough to verify this feature.

PR #699, opened 2015-08-19
TAJO-1741: Two tables having same time zone display different timestamps

eminency:
includeTimezone flag is for TIMESTAMPZ type, so changed to 'false'

[Review on Reviewable](https://reviewable.io/reviews/apache/tajo/699)


hyunsik:
+1
The change looks straightforward. I'll change your comments into a java doc format before committing.

hyunsik:
@eminency 

I missed one thing. Could you add insert tests using timezone?

eminency:
@hyunsik 

Thank you for review.
I applied your suggestion right before.

hyunsik:
+1
The patch looks good to me, and it looks enough for the unit test to verify this bug fix.

PR #700, opened 2015-08-19
TAJO-1758: Some TajoRuntimeException should be restored as TajoException.

hyunsik:
See https://issues.apache.org/jira/browse/TAJO-1758.


hyunsik:
I've updated the patch. It makes exception handling ways in client side consistent.

hyunsik:
I fixed test failures.

jihoonson:
+1, thanks for the nice patch!

PR #701, opened 2015-08-19
TAJO-1783: Query result is not returned by invalid output path

jinossy:


hyunsik:
I just reviewed the patch. The change and its clean up look great to me.

I tried to find non-forwarded tests on partitioned tables. I haven't found them. If they exist, could you let me know where they are?

Actually, it's not your problem. But, I'd like to ensure that your change guarantees that non-forwarded scans work well on partitioned tables.

jinossy:
You can find it like testColumnPartitionedTableByThreeColumns in TestTablePartitions.
https://github.com/apache/tajo/pull/431 does not allow the managed table. so it does not test the non-forwarded query.
Anyway,  I’m going to add TestSimpleQuery

hyunsik:
+1
The patch looks great  to me. The patch seems to include unit tests to verify the things you changed.

PR #702, opened 2015-08-19
TAJO-1787: Remove unused and legacy exceptions.

hyunsik:
[Review on Reviewable](https://reviewable.io/reviews/apache/tajo/702)

PR #706, opened 2015-08-20
TAJO-1766: Improve the performance of cross join

jihoonson:
[Review on Reviewable](https://reviewable.io/reviews/apache/tajo/706)


jihoonson:
Please don't start review until #702 is committed.

jihoonson:
This patch is not ready for review.

jihoonson:
This patch is ready for review. Sorry for a large patch. 

Here are some highlights of changes.
* Added a session variable to set the limitation of broadcast table size for cross join. This value is valid only when ```TEST_BROADCAST_JOIN_ENABLED``` is set.
* Cross join is always executed with broadcast join. To do so, at least one input of cross join should be the relation which is smaller than ```BROADCAST_CROSS_JOIN_THRESHOLD```.
* Added ```PostLogicalPlanVerifier``` to verify that cross join is executable or not.
* Fixed some bugs in BroadcastJoinRule.
* Fixed some bugs in QueryTestCaseBase.
* Removed BNL and NL join executors. Instead, each task executes cross join with hash join. This is because one of inputs of cross join is always cached in the broadcast cache holder.
* Improved unique key generation for a scan executor in ```TaskAttemptContext```.

I've tested cross join performance with a cluster consisting of a master and 5 workers. Each worker equips 48 cores, 80 GB memory, and 24 disks. 

#### Data
* partsupp: 80000000 rows (12.2 GB in TEXT file)
* supplier_small: 100000 rows (14.1 MB in TEXT file)

#### Query
```
select count(*) from partsupp, supplier_small
```

#### Result
1 hrs, 14 mins, 33 sec is taken with this patch. The above query runs forever without this patch because a single worker executes cross join.

jinossy:
Awesome!!

I've test your query on Tez and same cluster. but I've stop the query after 5h
```
hive> select count(*) from partsupp, supplier_small;
Warning: Shuffle Join MERGEJOIN[11][tables = [partsupp, supplier_small]] in Stage 'Reducer 2' is a cross product
Query ID = hadoop_20150825100744_8e11a2d9-d918-473f-bad2-36e403d6eca3
Total jobs = 1
Launching Job 1 out of 1


Status: Running (Executing on YARN cluster with App id application_1440464539141_0001)

--------------------------------------------------------------------------------
        VERTICES      STATUS  TOTAL  COMPLETED  RUNNING  PENDING  FAILED  KILLED
--------------------------------------------------------------------------------
Exiting the JVM.   SUCCEEDED    224        224        0        0       0       0
Interrupting... Be patient, this might take some time.0        0       0       0
Map 1 ..........   SUCCEEDED    224        224        0        0       0       0
Map 4 ..........   SUCCEEDED      1          1        0        0       0       0
Reducer 2            RUNNING      1          0        1        0       0       0
Reducer 3             INITED      1          0        0        1       0       0
--------------------------------------------------------------------------------
VERTICES: 02/04  [=========================>>-] 99%   ELAPSED TIME: 22510.28 s
```

hyunsik:
Here is my +1. The patch looks good to me.

jihoonson:
Thanks guys!

PR #708, opened 2015-08-24
TAJO-1782: Check ON_ERROR_STOP flag in TSQL when error is occured

dkhwangbo:
Since the bug tsql suicide itself by any error, check the ON_ERROR_STOP. If it is TRUE, TSQL quit itself with exitcode returned.


jihoonson:
Looks good to me.
@hyunsik, would you mind reviewing this patch? It looks to be related to your recent patches.

hyunsik:
The patch looks good to me. If possible, could you add some unit tests to guarantee the bug fix?

PR #712, opened 2015-08-27
TAJO-1738: Improve off-heap RowBlock

jinossy:


jinossy:
I’ve move the offheap package to the tajo-common and I’ve change the OffHeapMemory to MemoryBlock in order to the heap memory

hyunsik:
+1
The patch looks good to me.

PR #714, opened 2015-08-29
TAJO-1803: Use in-memory derby as the default catalog for unit tests

jihoonson:
[Review on Reviewable](https://reviewable.io/reviews/apache/tajo/714)


jihoonson:
I've also fixed some bugs in catalog.

jihoonson:

tajo-core-tests/src/test/java/org/apache/tajo/engine/query/TestSelectQuery.java
```diff
@@ -26,9 +26,11 @@
 import org.apache.tajo.TajoConstants;
 import org.apache.tajo.TajoProtos.QueryState;
 import org.apache.tajo.TajoTestingCluster;
+import org.apache.tajo.catalog.CatalogConstants;
```
hyunsik: unused imports

tajo-common/src/main/proto/errors.proto
```diff
@@ -157,6 +157,9 @@ enum ResultCode {
   CAT_UPGRADE_REQUIRED                  = 901;  // Migration
   CAT_CANNOT_CONNECT                    = 902;  // Cannot connect metadata server
 
+  // Catalog store
+  UNSUPPORTED_CATALOG_STORE             = 905; // Unsupported type of catalog store
```
hyunsik: It seems to be necessary the prefix ``CAT``.

hyunsik:
I used some trivial comments. You can commit the patch after you reflect them if you agree with my comments.

jinossy:
+1 LGTM

jihoonson:
Thanks for quick review!

PR #715, opened 2015-08-31
TAJO-1802: PythonScriptEngine copies controller and tajo util whenever it is initialized

jihoonson:
I've changed to copy python files exactly once before script engine is started in master. In worker, python files are treated as one of shared resources.

[Review on Reviewable](https://reviewable.io/reviews/apache/tajo/715)


jinossy:
In my opinion, once copied controller.py and tajo_util.py, It does not need to overwrite.
Can you move to TajoWorker.serviceInit and TajoMaster.serviceInit ?

jihoonson:
Thank you for the nice comment. I'll change it.

jinossy:
+1 LGTM!

PR #717, opened 2015-09-02
TAJO-1707: Rack local count can be more than actual number of tasks.

jinossy:


jinossy:
I've remove the cancellation count

jihoonson:
This patch looks good, but totalAssigned is the sum of only hostLocalAssigned and rackLocalAssigned. How about adding a new variable to keep the total number of attempts including cancelled attempts?

jinossy:
Good idea. I'm going to add totalAttempts

jihoonson:
+1 LGTM!

PR #718, opened 2015-09-02
TAJO-1809: Change default value of several configurations

jihoonson:


jinossy:
Great work! 
I ran queries with default configuration successfully 
tajo.task.resource.min.memory-mb 1G

hyunsik:
+1

The patch looks good to me.

jihoonson:
Thanks guys! I've just committed.

PR #720, opened 2015-09-02
TAJO-993: Cleanup the result data in HDFS after query finished.

jinossy:


hyunsik:
+1
The change is straightforward. Ship it.

PR #725, opened 2015-09-04
TAJO-1814: Add some positive/negative tests for CREATE TABLE IF NOT E…

hyunsik:
…XISTS.


hyunsik:
This is a set of unit tests to guarantee the fix of the bug of TAJO-1606.  I rebased it.

jihoonson:
+1. Test failure looks not related to this patch.

PR #728, opened 2015-09-04
TAJO-1740: Update Partition Table document

blrunner:
Currently, Tajo doesn't provide enough informations about partition table. Thus, we need to add more informations to following documentation.
http://tajo.apache.org/docs/current/partitioning/column_partitioning.html


tajo-docs/src/main/sphinx/partitioning/column_partitioning.rst
```diff
@@ -11,22 +11,40 @@ How to Create a Column Partitioned Table
 You can create a partitioned table by using the ``PARTITION BY`` clause. For a column partitioned table, you should use
 the ``PARTITION BY COLUMN`` clause with partition keys.
 
-For example, assume there is a table ``orders`` composed of the following schema. ::
+For example, assume there is a table ``student`` composed of the following schema.
```
hyunsik: ``assume a table with the following schema`` would be better.

blrunner:
Thanks @hyunsik 
I've reflected your comments. :)

tajo-docs/src/main/sphinx/partitioning/column_partitioning.rst
```diff
@@ -11,22 +11,40 @@ How to Create a Column Partitioned Table
...
+For example, assume a table with the following schema.
 
-  id          INT,
-  item_name   TEXT,
-  price       FLOAT
+.. code-block:: sql
+
+  id     INT,
+  name   TEXT,
+  grade  TEXT
```
jihoonson: Maybe you intend that this table has the five columns of id, name, grade, country, and city. If so, how about showing all columns as the below example of ```student_source```? It will be much easier to understand.

blrunner:
Thanks @jihoonson 

I reflected your comments as follows.
https://github.com/blrunner/tajo/commit/37283a0164a9fba298d1bd64e9b7f38b74d2dc10

jihoonson:
Thanks, but your patch looks to be include other patches. Would you fix it?

blrunner:
Thanks @jihoonson .
 Appears to be twisted. I'll create new PR soon.

PR #731, opened 2015-09-05
TAJO-1816: Refactor SQL parser tests.

hyunsik:


tajo-core-tests/src/test/java/org/apache/tajo/engine/parser/TestSQLAnalyzer.java
```diff
@@ -18,805 +18,192 @@
...
+   *
+   * @throws IOException
+   * @throws URISyntaxException
+   */
+  @Test
+  public void testPositiveTests() throws IOException, URISyntaxException {
+    for (Pair<String, String> pair : getFileContents("positive")) {
+      try {
+        assertNotNull(parseQuery(pair.getSecond()));
+        System.out.println(pair.getFirst() + " test passed...");
```
blrunner: How about use log4j instead of PrintStream?

tajo-core-tests/src/test/java/org/apache/tajo/engine/parser/TestSQLAnalyzer.java
```diff
@@ -18,805 +18,192 @@
...
+        expectedResult = FileUtil.readTextFileFromResource("results/TestSQLAnalyzer/" + fileName + ".result");
+      } catch (FileNotFoundException ioe) {
+        expectedResult = "";
+      } catch (Throwable t) {
+        fail(t.getMessage());
+      }
+
+      assertEquals(pair.getFirst() + " is different from " + fileName + ".result",
+          expectedResult.trim(), expr.toJson().trim());
+      System.out.println(pair.getFirst() + " test passed..");
```
blrunner: It would be better if log4j is used instead of PrintStream.

blrunner:
+1

I left a few trivial comments. You can commit the patch after you reflect them if you agree with my comments.

hyunsik:
Thank you for the quick review. In my opinion, these prints will not be problematic because there are no concerns about log level or buffer flush problem of System.out. I'll commit it soon. Thank you for the quick review.

PR #732, opened 2015-09-05
TAJO-1727: Avoid to create external table using TableSpace

blrunner:
When creating external table, add setting tablespace.


hyunsik:
External table must requires a location URL. So, it does not need a tablespace clause.

blrunner:
Thanks @hyunsik .

tajo-core/src/main/java/org/apache/tajo/parser/sql/SQLAnalyzer.java
```diff
@@ -1285,6 +1285,10 @@ public Expr visitCreate_table_statement(Create_table_statementContext ctx) {
     if (checkIfExist(ctx.EXTERNAL())) {
       createTable.setExternal();
 
+      if (checkIfExist(ctx.TABLESPACE())) {
+        throw new TajoRuntimeException(new SQLSyntaxError("CREATE EXTERNAL TABLE cannot be used with tablespace"));
```
hyunsik: It would be better if Tablespace clause is a subject. So, I'd like to recommend ``Tablespace clause is not allowed for an external table``.

tajo-core/src/main/java/org/apache/tajo/parser/sql/SQLAnalyzer.java
```diff
@@ -1293,6 +1297,9 @@ public Expr visitCreate_table_statement(Create_table_statementContext ctx) {
       if (checkIfExist(ctx.LOCATION())) {
         String uri = stripQuote(ctx.uri.getText());
         createTable.setLocation(uri);
+      } else {
+        throw new TajoRuntimeException(new SQLSyntaxError("EXTERNAL and LOCATION clauses need to be specified " +
```
hyunsik: How about ``LOCATION clause must be required for an external table.``?

hyunsik:
+1

The patch looks good to me. It's right validation. If you agree with my comments, you can commit it after you reflect my comments.

jihoonson:
Please hold on a second.
This patch lacks any test. Would you add it?
In addition, the issue title is mismatched with the patch.

hyunsik:
I agree with Jihoon's comment. The title may be matter. For unit tests, I'm going to add some unit tests after I improve the exception system with the better error message. So, it can be skip if you guys agree with my plan.

blrunner:
Thanks guys. I also think that the tile is mismatched.

blrunner:
I updated the title and added unit test cases.

blrunner:
Thanks @jihoonson 

I've reflected your comments and rebased. :-)

jihoonson:
+1 looks good to me!

blrunner:
Thanks @jihoonson 

I'll ship it. :)

PR #733, opened 2015-09-05
TAJO-1818: Separate sql parser into an independent maven module.

hyunsik:
This patch includes TAJO-1816 changes. It should be reviewed and committed after TAJO-1816.


hyunsik:
This patch just moves engine.parser pakcage into a separate maven module ``tajo-parser-sql``. So, its review will be easy.

hyunsik:
rebased

blrunner:
+1

LGTM

PR #734, opened 2015-09-07
TAJO-1739: Add a statement for adding partition to TajoDump

blrunner:
Add a statement for adding partition to TajoDump

### DIFF
--- a/tajo-catalog/tajo-catalog-common/src/main/proto/CatalogProtos.proto
+++ b/tajo-catalog/tajo-catalog-common/src/main/proto/CatalogProtos.proto
@@ -407,3 +407,8 @@ message IndexResponse {
   required ReturnState state = 1;
   optional IndexDescProto indexDesc = 2;
 }
+
+message GetPartitionsResponse {
+  required ReturnState state = 1;
+  repeated PartitionDescProto partition = 2;
+}
--- a/tajo-catalog/tajo-catalog-common/src/main/proto/CatalogProtos.proto
+++ b/tajo-catalog/tajo-catalog-common/src/main/proto/CatalogProtos.proto
@@ -408,7 +408,7 @@ message IndexResponse {
   optional IndexDescProto indexDesc = 2;
 }
 
-message GetPartitionsResponse {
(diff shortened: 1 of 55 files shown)


tajo-catalog/tajo-catalog-common/src/main/proto/CatalogProtos.proto
```diff
@@ -407,3 +407,8 @@ message IndexResponse {
   required ReturnState state = 1;
   optional IndexDescProto indexDesc = 2;
 }
+
+message GetPartitionsResponse {
```
jihoonson: Every message for rpc communication with catalog should locate in CatalogProtocol.proto.

jihoonson: You seem to move this message to here to use in TajoMasterClientProtocol. If so, rename it like other messages.

tajo-client/src/main/java/org/apache/tajo/client/CatalogAdminClientImpl.java
```diff
@@ -241,6 +241,22 @@ public TableDesc getTableDesc(final String tableName) throws UndefinedTableExcep
...
+
+    GetPartitionsResponse response;
+    try {
+      response = stub.getPartitionsByTableName(null,
+        conn.getSessionedString(tableName));
+    } catch (ServiceException e) {
+      throw new RuntimeException(e);
+    }
+
+    ensureOk(response.getState());
```
jihoonson: ```UndefinedTableException``` and ```UndefinedDatabaseException``` must be handled.

blrunner:
@jihoonson 

I reflected your comments.

jihoonson:
I missed that AbstractDBStore.getPartitions() already throws UndefinedTableException and UndefinedDatabaseException properly. Sorry for the last comment.

jihoonson:
+1 looks good to me!

blrunner:
Thanks @jihoonson 
I'll ship it. :)

PR #735, opened 2015-09-07
TAJO-1610: Cannot find column when the same name is used for table and database.

hyunsik:
See https://issues.apache.org/jira/browse/TAJO-1610.


jihoonson:
+1 LGTM!

PR #737, opened 2015-09-07
TAJO-1821: Temporary data is not cleared after TestCatalog

jihoonson:


tajo-catalog/tajo-catalog-server/src/test/java/org/apache/tajo/catalog/TestCatalog.java
```diff
@@ -64,18 +62,10 @@
...
-  }
-
-  public static File setupClusterTestBuildDir() throws IOException {
-    String randomStr = UUID.randomUUID().toString();
-    String dirStr = CommonTestingUtil.getTestDir(randomStr).toString();
-    File dir = new File(dirStr).getAbsoluteFile();
-    // Have it cleaned up on exit
-    dir.deleteOnExit();
-    return dir;
+  public static TajoConf newTajoConfForCatalogTest(String testDir) throws IOException, UnsupportedCatalogStore {
```
hyunsik: It would be better if it returns a pair of TajoConf and auto-generated directory instead of taking ``testDir``, especially if it is used to make a temporary directory rather than a designated directory.

hyunsik:
The patch looks good to me.

jihoonson:
Thanks. I've reflected your comment.

hyunsik:
+1 LGTM!

jinossy:
I'm reviewing. Hold on please

jinossy:
+1 LGTM

jihoonson:
Thank you for your review!

PR #738, opened 2015-09-07
TAJO-1641: Add window function documentation

jihoonson:


hyunsik:
The unit test failed, but it seems to be not related to this patch. Here is my +1. The patch looks good to me.

PR #739, opened 2015-09-07
TAJO-1805: In the 'Execute Query' page of web UI, default database should be set as 'default'

dkhwangbo:
See the title. 
Before this patch, 'information_schema' is setted to default database in 'Execute Query' page. 
this patch fix this part.


hyunsik:
+1

PR #740, opened 2015-09-07
TAJO-1823: Can't start TajoMaster.

hyunsik:


jinossy:
Thanks for your quick fix but I can't compile on IDE
The problem looks like that incorrect generated source path and module order.

Could you create the package directory ?
```
tajo-sql-parser/src/main/antlr4/org.apache.tajo.parser.sql
```

Please change module order as follows : 
```
...
<module>tajo-algebra</module>
<module>tajo-sql-parser</module>
...
```

hyunsik:
Thank you for the comment. I'll check out them.

hyunsik:
I fixed the wrong package name and rebased the patch.

jinossy:
+1

PR #742, opened 2015-09-08
TAJO-1729: No handling of default case in DDLExecutor

dkhwangbo:
As you can see, there is no handling code of default case in alterTable in DDLExecutor.java.
so add line like execute() in same file.


jinossy:
+1
Thanks for your contribution!

PR #743, opened 2015-09-08
TAJO-1824: Remove partition_keys table from information_schema

blrunner:
Selecting partition keys is not necessary because partitions are searched with its partition name. It would be better to remove from information_schema.


tajo-core-tests/src/test/java/org/apache/tajo/master/TestNonForwardQueryResultSystemScanner.java
```diff
@@ -38,4 +47,20 @@ public void testGetNextRowsForTable() throws Exception {
   public void testGetClusterDetails() throws Exception {
     assertQueryStr("SELECT TYPE FROM INFORMATION_SCHEMA.CLUSTER");
   }
+
```
hyunsik: I think that this test is not necessary.

hyunsik:
This change looks straightforward. So, the added unit test is not necessary.

blrunner:
Thanks @hyunsik 

I've just removed the unit test.

hyunsik:
+1

PR #744, opened 2015-09-08
TAJO-1825: Remove zero length fragments when file length is zero.

jinossy:


jihoonson:
+1 thank you for the nice work!

PR #747, opened 2015-09-08
TAJO-1829: Fix DelimitedTextFileAppender NPE in negative tests.

jinossy:


hyunsik:

PR #748, opened 2015-09-09
TAJO-1830: Fix race condition in HdfsServiceTracker.

jinossy:


tajo-core/src/main/java/org/apache/tajo/ha/HdfsServiceTracker.java
```diff
@@ -458,21 +454,37 @@ public InetSocketAddress getMasterHttpInfo() throws ServiceTrackerException {
...
           this.wait(pause);
         } catch (InterruptedException e) {
           throw new ServiceTrackerException(e);
         }
+
         files = fs.listStatus(activeMasterBaseDir);
       }
 
+      if (activeMasterEntry == null) {
+        throw new ServiceTrackerException("No such active master in : " + activeMasterBaseDir);
```
hyunsik: It would be better if the exception message is changed to 'Active master entry cannot be founded in '.

tajo-core/src/main/java/org/apache/tajo/ha/HdfsServiceTracker.java
```diff
@@ -458,21 +454,37 @@ public InetSocketAddress getMasterHttpInfo() throws ServiceTrackerException {
...
       }
 
-      FileStatus[] files = fs.listStatus(activeMasterBaseDir);
       /* wait for active master from HDFS */
       int pause = conf.getIntVar(ConfVars.TAJO_MASTER_HA_CLIENT_RETRY_PAUSE_TIME);
       int maxRetry = conf.getIntVar(ConfVars.TAJO_MASTER_HA_CLIENT_RETRY_MAX_NUM);
       int retry = 0;
 
-      while (files.length < 2 && retry < maxRetry) {
+      FileStatus[] files = fs.listStatus(activeMasterBaseDir);
```
hyunsik: It would be better if it is moved into the inner loop. If so, it will handle the case where the master entry is created during retrying.

jinossy:
Thanks for the review
I've update the patch that reflects your comments

hyunsik:
Could you trigger it?

jinossy:
triggered it

hyunsik:
+1

The patch looks good to me.

PR #749, opened 2015-09-09
TAJO-1813: Allow external catalog store for unit testing

jihoonson:


jihoonson:
This patch is only to test catalog.
I've added two more unit tests, i.e., ```TestCatalogAgainstCaseSensitivity``` and ```TestCatalogExceptions```. As shown in those names, one of them is to test catalog functionalities against text case sensitivity and the other is to test if catalog throws exceptions properly.
I've verified every unit test of ```tajo-catalog-server``` with pgsql, mysql, mariadb, and derby.

jihoonson:
I miss tests for some partition APIs and nested fields. I'll add them.

jihoonson:

tajo-catalog/tajo-catalog-client/src/main/java/org/apache/tajo/catalog/AbstractCatalogClient.java
```diff
@@ -598,18 +604,15 @@ public final boolean existsTable(final String tableName) {
   }
 
   @Override
-  public final boolean createIndex(final IndexDesc index) {
+  public final void createIndex(final IndexDesc index) throws DuplicateIndexException {
```
hyunsik: If index creation is based on table, it may need ``Undefined(Table|Database)Exception``.

tajo-catalog/tajo-catalog-common/src/main/java/org/apache/tajo/catalog/CatalogService.java
```diff
@@ -169,7 +169,7 @@ PartitionDescProto getPartition(String databaseName, String tableName, String pa
   void addPartitions(String databaseName, String tableName, List<PartitionDescProto> partitions
     , boolean ifNotExists) throws UndefinedTableException, DuplicatePartitionException, UndefinedPartitionMethodException, UndefinedDatabaseException;
 
-  boolean createIndex(IndexDesc index);
+  void createIndex(IndexDesc index) throws DuplicateIndexException;
```
hyunsik: If index creation is based on table, it may need ``Undefined(Table|Database)Exception``.

tajo-catalog/tajo-catalog-server/pom.xml
```diff
@@ -127,6 +127,15 @@
         <groupId>org.apache.maven.plugins</groupId>
         <artifactId>maven-surefire-report-plugin</artifactId>
       </plugin>
+      <plugin><groupId>org.apache.maven.plugins</groupId><artifactId>maven-surefire-plugin</artifactId><configuration>
+        <systemProperties>
+          <tajo.test>TRUE</tajo.test>
+        </systemProperties>
+        <additionalClasspathElements>
+          <additionalClasspathElement>/home/jihoon/다운로드/postgresql-9.4-1201.jdbc4.jar</additionalClasspathElement>
```
hyunsik: It seems to be a mistake?

tajo-catalog/tajo-catalog-server/src/main/resources/schemas/mariadb/mariadb.xml
```diff
@@ -19,6 +19,7 @@
 <tns:store xmlns:tns="http://tajo.apache.org/catalogstore" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://tajo.apache.org/catalogstore ../DBMSSchemaDefinition.xsd ">
   <!--
       Catalog base version history
+      * 9 - TODO
```
hyunsik: You may misse this description.

tajo-catalog/tajo-catalog-server/src/main/resources/schemas/mysql/mysql.xml
```diff
@@ -19,6 +19,7 @@
 <tns:store xmlns:tns="http://tajo.apache.org/catalogstore" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://tajo.apache.org/catalogstore ../DBMSSchemaDefinition.xsd ">
   <!--
     Catalog base version history
+    * 9 - TODO
```
hyunsik: You may miss this description.

tajo-catalog/tajo-catalog-server/src/main/resources/schemas/oracle/oracle.xml
```diff
@@ -19,6 +19,7 @@
 <tns:store xmlns:tns="http://tajo.apache.org/catalogstore" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://tajo.apache.org/catalogstore ../DBMSSchemaDefinition.xsd ">
   <!--
       Catalog base version history
+      * 9 - TODO
```
hyunsik: You seem to miss this description.

tajo-catalog/tajo-catalog-server/src/main/resources/schemas/postgresql/postgresql.xml
```diff
@@ -21,6 +21,7 @@ xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
 xsi:schemaLocation="http://tajo.apache.org/catalogstore ../DBMSSchemaDefinition.xsd ">
   <!--
       Catalog base version history
+      * 9 - TODO
```
hyunsik: You seem to miss this description.

tajo-catalog/tajo-catalog-server/src/test/java/org/apache/tajo/catalog/TestCatalogExceptions.java
```diff
@@ -0,0 +1,293 @@
...
+  @AfterClass
+  public static void tearDown() throws Exception {
+    CatalogTestingUtil.cleanupBaseData(catalog);
+    server.stop();
+  }
+
+  @Test
+  public void testCreateTablespaceWithWrongUri() throws Exception {
+    // TODO
+    catalog.createTablespace("wrong", "hdfs:");
```
hyunsik: It seems to throw some exception, but it does not catch or handle the exception.

tajo-catalog/tajo-catalog-server/src/test/java/org/apache/tajo/catalog/TestCatalogExceptions.java
```diff
@@ -0,0 +1,293 @@
...
+  }
+
+  @Test(expected = InsufficientPrivilegeException.class)
+  public void testDropDefaultDatabase() throws Exception {
+    catalog.dropDatabase(TajoConstants.DEFAULT_DATABASE_NAME);
+  }
+
+  @Test()
+  public void testCreateTableWithWrongUri() throws Exception {
+    // TODO
```
hyunsik: It seems to throw some exception, but it does not catch or handle the exception.

tajo-catalog/tajo-catalog-server/src/test/java/org/apache/tajo/catalog/TestCatalogExceptions.java
```diff
@@ -0,0 +1,293 @@
...
+                TableStatsProto.newBuilder().
+                    setNumRows(0).
+                    setNumBytes(0).
+                    build()).
+            build());
+  }
+
+  @Test
+  public void testAddPartitionWithWrongUri() throws Exception {
+    // TODO
```
hyunsik: It seems to throw some exception, but it does not catch or handle the exception.

hyunsik:
It's excellent! I leaved some comments.

jihoonson:
@hyunsik thank you for your review. I updated the patch according to your comment except wrong URI tests. The problem is that we cannot test against wrong URI input because it doesn't cause any errors. I think we need to handle this problem at another issue.

hyunsik:
+1 
The patch looks good to me. Ship it!

PR #752, opened 2015-09-10
TAJO-1833: Refine LogicalPlanPreprocessor to add new rules easily

jihoonson:


hyunsik:
+1
The patch looks good to me.

jihoonson:
Thanks for quick review!

PR #756, opened 2015-09-12
TAJO-1832: Well support for self-describing data formats

jihoonson:
This is still on-going work, but mostly finished.
Honestly, I think it will be finished if I improve unit tests and cleanup some code lines.
Before finishing it, I'd like to share my approach for this issue.


jihoonson:
This test failure is due to #757. Please review it first.

jihoonson:

tajo-catalog/tajo-catalog-common/src/main/java/org/apache/tajo/catalog/TableDesc.java
```diff
@@ -35,34 +36,43 @@
 
 public class TableDesc implements ProtoObject<TableDescProto>, GsonObject, Cloneable {
 	@Expose protected String tableName;                        // required
-  @Expose protected Schema schema;
+  @Expose protected Schema schema;                           // optional for self-describing tables
   @Expose protected TableMeta meta;                          // required
   /** uri is set if external flag is TRUE. */
-  @Expose protected URI uri;                                // optional
+  @Expose protected URI uri;                                 // optional
```
hyunsik: In the last work, I changed URI to be required. Regardless of whether it is manged or external, every TableDescs has the table URI. The evidence is unused ``TableDesc::hasUri()``, which can be removed. Additional evidence is that there is no nullity check around ``getUri()``.

Otherwise, it is a bug.

tajo-catalog/tajo-catalog-common/src/main/proto/CatalogProtos.proto
```diff
@@ -98,10 +98,11 @@ message TableDescProto {
   required string table_name = 2;
   optional string path = 3;
   required TableProto meta = 4;
-  required SchemaProto schema = 5;
+  optional SchemaProto schema = 5;
```
hyunsik: If we can ensure that the TableDescs of all normal tables have Schema, we can use its ``optional`` constraint to indicate either self-describing or not.

tajo-rpc/tajo-rpc-protobuf/src/main/java/org/apache/tajo/rpc/RpcClientManager.java
```diff
@@ -199,7 +199,6 @@ public static void close() {
    * After it is shutdown it is not possible to reuse it again.
    */
   public static void shutdown() {
-    close();
```
hyunsik: Is it relevant in this patch? If so, could you explain the behind reason?

tajo-sql-parser/src/main/antlr4/org/apache/tajo/parser/sql/SQLParser.g4
```diff
@@ -106,9 +106,9 @@ if_exists
   ;
 
 create_table_statement
-  : CREATE EXTERNAL TABLE (if_not_exists)? table_name table_elements (TABLESPACE spacename=identifier)? USING storage_type=identifier
+  : CREATE EXTERNAL TABLE (if_not_exists)? table_name (table_elements)? (TABLESPACE spacename=identifier)? USING storage_type=identifier
```
hyunsik: This grammar allows to create table without parenthesis ``(..)``. It looks good to me.

I also have another suggestion. How do you think about a special notation with parenthesis like ``CREATE EXTERNAL TABLE (*) USING JSON``. I used asterisk ``*`` as a meaning of becoming anything and the infinite number of fields, which can be zero to up to be infinite.

tajo-plan/src/main/java/org/apache/tajo/plan/rewrite/SelfDescSchemaBuildPhase.java
```diff
@@ -0,0 +1,479 @@
...
+        String simpleName = eachColumn.getSimpleName();
+        if (NestedPathUtil.isPath(simpleName)) {
+          String[] paths = simpleName.split(NestedPathUtil.PATH_DELIMITER);
+          for (int i = 0; i < paths.length-1; i++) {
+            String parentName = paths[i];
+            if (i == 0) {
+              parentName = CatalogUtil.buildFQName(eachColumn.getQualifier(), parentName);
+            }
+            // Leaf column type is TEXT; otherwise, RECORD.
+            Type childDataType = (i == paths.length-2) ? Type.TEXT : Type.RECORD;
```
hyunsik: These lines exceed 120 column width. In addition, ``visitCreateTable()`` in this file exceeds 120 too.

tajo-client/src/main/proto/ClientProtos.proto
```diff
@@ -198,6 +198,7 @@ message CreateTableRequest {
   required TableProto meta = 4;
   required string path = 5;
   optional PartitionMethodProto partition = 6;
+  required bool hasSelfDescSchema = 7;
```
hyunsik: In this message scheme, SchemaProto is required. it may have an empty Schema? This semantic seems to be different from TableDesc. It's trivial thing. If we see more consistent semantic, the code would be more comprehensive.

hyunsik:
It is an interesting feature. I leaved comments.

tajo-rpc/tajo-rpc-protobuf/src/main/java/org/apache/tajo/rpc/RpcClientManager.java
jihoonson: This is not directly related to this ticket. It is to solve the hang problem after executing any unit test.

tajo-catalog/tajo-catalog-common/src/main/proto/CatalogProtos.proto
jihoonson: I agree.

tajo-sql-parser/src/main/antlr4/org/apache/tajo/parser/sql/SQLParser.g4
jihoonson: Nice idea!

jihoonson:
I reflected comments.

hyunsik:
You missed to update ``<tns:base version="9">`` in ``resources/schemas/*/*.xml`` files.

tajo-core-tests/src/test/java/org/apache/tajo/engine/query/TestQueryOnSelfDescTable.java
```diff
@@ -0,0 +1,185 @@
...
+    executeString(String.format("create external table self_desc_table3 (*) using json location '%s'",
+        getDataSetFile("tweets"))).close();
+
+    executeString(String.format("create external table github (*) using json location '%s'",
+        getDataSetFile("github"))).close();
+  }
+
+  @After
+  public void teardown() throws TajoException, SQLException {
+    executeString("drop table self_desc_table1").close();
```
hyunsik: It's a trivial comment. It would be better if you put ``IF NOT EXISTS`` to them. It will suppress error messages in unexpected cases.

tajo-core-tests/src/test/java/org/apache/tajo/engine/query/TestQueryOnSelfDescTable.java
```diff
@@ -0,0 +1,185 @@
...
+package org.apache.tajo.engine.query;
+
+import org.apache.tajo.QueryTestCaseBase;
+import org.apache.tajo.exception.AmbiguousColumnException;
+import org.apache.tajo.exception.TajoException;
+import org.junit.After;
+import org.junit.Test;
+
+import java.io.IOException;
+import java.sql.ResultSet;
```
hyunsik: Unused import

tajo-docs/src/main/sphinx/sql_language/ddl.rst
```diff
@@ -40,7 +40,7 @@ CREATE TABLE
   CREATE TABLE [IF NOT EXISTS] <table_name> [(<column_name> <data_type>, ... )] [TABLESPACE tablespace_name]
   [using <storage_type> [with (<key> = <value>, ...)]] [AS <select_statement>]
 
-  CREATE EXTERNAL TABLE [IF NOT EXISTS] <table_name> (<column_name> <data_type>, ... )
+  CREATE EXTERNAL TABLE [IF NOT EXISTS] <table_name> [(<column_name> <data_type>, ... )]
```
hyunsik: The documentation needs to reflect the changed DDL for ``(*)``.

tajo-docs/src/main/sphinx/table_management/json.rst
```diff
@@ -0,0 +1,100 @@
...
+****
+
+JSON(JavaScript Object Notation) is an open standard format for data (de)serialization. Since it is simple and human-readable, it is popularly used in many fields.
+Tajo supports JSON as its data format. In this section, you will get an overview of how to create JSON tables and query on them.
+
+============================
+How to Create a JSON Table ?
+============================
+
+You can create JSON tables using ``CREATE TABLE`` statement. (For more information, please refer to :doc:`/sql_language/ddl`.)
```
hyunsik: ``JSON tables`` should be ``a JSON table``.

tajo-docs/src/main/sphinx/table_management/json.rst
```diff
@@ -0,0 +1,100 @@
...
+For example, please consider an example data as follows:
+
+.. code-block:: bash
+
+  $ hdfs dfs -cat /table1/table.json
+  { "title" : "Hand of the King", "name" : { "first_name": "Eddard", "last_name": "Stark"}}
+  { "title" : "Assassin", "name" : { "first_name": "Arya", "last_name": "Stark"}}
+  { "title" : "Dancing Master", "name" : { "first_name": "Syrio", "last_name": "Forel"}}
+
+Tajo provides two alternatives to create a table for this data. First is a traditional way to create tables. Here is an example.
```
hyunsik: How about ``two ways`` instead of ``two alternatives``?

tajo-docs/src/main/sphinx/table_management/json.rst
```diff
@@ -0,0 +1,100 @@
...
+      last_name TEXT
+    )
+  ) USING JSON LOCATION '/table1/table.json';
+
+With this way, you need to specify every column which they want to use. This will be a tedious work, and not appropriate for flexible JSON schema.
+Second is a simpler alternative to alleviate this problem. When you create an external table of JSON format, you can simply omit the column specification as follows:
+
+.. code-block:: sql
+
+  CREATE EXTERNAL TABLE table1 USING JSON LOCATION '/table1/table.json';
```
hyunsik: It needs to reflect the changed schema description way; i.e., ``(*)``.

hyunsik:
The patch looks nice to me. I leaved some comments.

I have one question. Could you briefly explain the priority of name resolving in the cases where nested column names are duplicated to database and table names?

jihoonson:
Thank you for your review. 

As the answer for your question, the ```already known information``` has a higher priority. For example, if the nested column name of a table of self-describing format has the same name with a database or a table, the database name (or the table name) has a higher priority than nested column name because it is already known information.
However, the table has a predefined schema, NameResolver follows the existing routine even though the query involves a column of nested type.

hyunsik:
Excellent! The policy sounds reasonable. Here is my +1. Ship it.

jihoonson:
Thanks!

PR #757, opened 2015-09-12
TAJO-1845: Enforcers in the master plan should be printed in a fixed order

jihoonson:


jihoonson:
I've changed some test queries for more deterministic test result.
In these queries, a single table is used twice for join, that is self join. Since the join tables have the same volume, broadcast table can be changed.

tajo-core-tests/src/test/resources/results/TestInnerJoinWithSubQuery/testComplexJoinCondition5.result
```diff
@@ -5,25 +5,3 @@ n_nationkey,name1,name2
 2,BRAZ,BRAZ
 3,CANA,CANA
 4,EGYP,EGYP
-5,ETHI,ETHI
```
jinossy: @jihoonson 
Is this correct result ?

jihoonson: Yes, because I changed the query. The original query performs the self-join on ```nation```, which makes the query plan non-deterministic. So, I changed it to the join of ```nation``` and ```small_nation``` which is derived from the ```nation``` table.

jinossy: Oops! the table is small_nation. I understood

jinossy:
+1 LGTM

jihoonson:
Thank you for the quick review!

PR #759, opened 2015-09-14
TAJO-1848: ShutdownHook in TajoMaster can throw NPE if serviceInit() is failed.

hyunsik:
…is failed.


blrunner:
+1

Ship it! :)

PR #761, opened 2015-09-16
TAJO-1846: Write the directory address abide by user's OS

dkhwangbo:
in PythonScriptEngine, Some of variable has no consideration in user's os.
Cannot run start-tajo.cmd in Windows 7 because of this problem.


jihoonson:
+1 I'll commit shortly.

PR #765, opened 2015-09-17
TAJO-1864: CatalogStore::getAllPartitions need to throw PartitionNotFoundException if there is no partitions on catalog.

blrunner:
Currently, CatalogStore::getAllPartitions doesn't throw PartitionNotFoundException if there is not partitions on catalog. As a result, if users execute a select statement for partition table without filter, they can't get expected result.


blrunner:
This issue will be resolved in https://issues.apache.org/jira/browse/TAJO-1493

PR #766, opened 2015-09-17
TAJO-1861: TSQL should change line after printout error message during connecting other database

dkhwangbo:
TSQL should change line after printout this error message.


jihoonson:
+1. This patch is straightforward.

PR #767, opened 2015-09-17
TAJO-1867: TajoMaster should handle the change of ${tajo.root}.

hyunsik:


hyunsik:
It may be hard to make unit tests. I verified it in a local cluster.

jinossy:
+1 LGTM

PR #771, opened 2015-09-18
TAJO-1871: '-DskipTests' flag does not work

jihoonson:


tajo-storage/tajo-storage-pgsql/pom.xml
```diff
@@ -236,7 +236,6 @@
             <groupId>org.apache.maven.plugins</groupId>
             <artifactId>maven-surefire-plugin</artifactId>
             <configuration>
```
jinossy: It is my fault. but your patch can't run tests
please override configuration as follows:
<configuration combine.self="override">

tajo-storage/tajo-storage-pgsql/pom.xml
```diff
@@ -224,10 +224,11 @@
   <profiles>
     <profile>
       <!-- testing-postgresql-server only supports x86-64 -->
-      <id>64bit-tests</id>
+      <id>64bit-intel-tests</id>
       <activation>
         <os>
-         <arch>x86_64</arch>
+          <name>x86_64</name>
```
hyunsik: As far as I know, ``name`` is intended to describe operation system name like ``linux`` or ``mac os x``.

jihoonson:
I've tested with my linux desktop and MBP.

jinossy:
+1 LGTM

jihoonson:
Thanks for your review.

PR #772, opened 2015-09-21
TAJO-1493: Make partition pruning based on catalog informations.

blrunner:
Reopen PR which contains following features.
* Add columns for partition volume and file count to Catalog
* Allow to scan partitioned table which contains removed partition directories.
* Implement PartitionNotFoundException handling
* Implement UnsupportedException handling
* Add unit test cases for abnormal partition directories

*NOTE: old PR is https://github.com/apache/tajo/pull/624.*


jihoonson:
I'm reviewing.

tajo-catalog/tajo-catalog-common/src/main/java/org/apache/tajo/catalog/CatalogUtil.java
```diff
@@ -812,6 +812,22 @@ public static AlterTableDesc setProperty(String tableName, KeyValueSet params, A
...
+   * *
+   * @param tableName
+   * @param columns
+   * @param values
+   * @param location
+   * @param alterTableType
+   * @return
+   */
+  public static AlterTableDesc addOrDropPartition(String tableName, String[] columns,
+                                                  String[] values, String location, AlterTableType alterTableType) {
```
jihoonson: ```location``` looks confused to me. How about using 'path'?

tajo-catalog/tajo-catalog-common/src/main/java/org/apache/tajo/catalog/CatalogUtil.java
```diff
@@ -821,7 +837,8 @@ public static AlterTableDesc setProperty(String tableName, KeyValueSet params, A
    * @return AlterTableDesc
    */
   public static AlterTableDesc addOrDropPartition(String tableName, String[] columns,
-                                            String[] values, String location, AlterTableType alterTableType) {
+      String[] values, String location, AlterTableType alterTableType,
```
jihoonson: ```location``` looks confused to me. How about using 'path'?

tajo-catalog/tajo-catalog-common/src/main/proto/CatalogProtos.proto
```diff
@@ -262,6 +263,18 @@ message PartitionIdentifierProto {
   optional string partitionName = 3;
 }
 
+message PartitionsByAlgebraProto {
+  required string databaseName = 1;
+  required string tableName = 2;
+  required string algebra = 3;
```
jihoonson: It would be better if you leave a comment of how the algebra is serialized into a string.

tajo-catalog/tajo-catalog-common/src/main/proto/CatalogProtos.proto
```diff
@@ -262,6 +263,18 @@ message PartitionIdentifierProto {
...
+}
+
+message PartitionsByFilterProto {
+  required string databaseName = 1;
+  required string tableName = 2;
+  required string filter = 3;
```
jihoonson: It would be better if you leave a comment of how the filter is serialized into a string.

tajo-catalog/tajo-catalog-drivers/tajo-hive/src/main/java/org/apache/tajo/catalog/store/HiveCatalogStore.java
```diff
@@ -845,11 +856,155 @@ public boolean existPartitionMethod(String databaseName, String tableName) throw
...
+
+    try {
+      String databaseName = request.getDatabaseName();
+      String tableName = request.getTableName();
+
+      TableDescProto tableDesc = getTable(databaseName, tableName);
+      String filter = getFilter(databaseName, tableName, tableDesc.getPartition().getExpressionSchema().getFieldsList()
+        , request.getAlgebra());
+      list = getPartitionsByFilterFromHiveMetaStore(databaseName, tableName, filter);
+    } catch (UnsupportedException ue) {
```
jihoonson: Who does throw this exception?

tajo-catalog/tajo-catalog-drivers/tajo-hive/src/main/java/org/apache/tajo/catalog/store/HiveCatalogStore.java
```diff
@@ -845,11 +856,155 @@ public boolean existPartitionMethod(String databaseName, String tableName) throw
...
+    for (int i = 0; i < partitionColumns.size(); i++) {
+      target = new Column(partitionColumns.get(i));
+
+      if (!(filters[i] instanceof IsNullPredicate)) {
+        visitor.setColumn(target);
+        visitor.visit(null, new Stack<Expr>(), filters[i]);
+        result = visitor.getResult();
+
+        // If visitor build filter successfully, add filter to be used for executing hive api.
+        if (result.length() > 0) {
```
jihoonson: When does the result have zero-length?
We should handle this situation.

tajo-catalog/tajo-catalog-server/src/main/java/org/apache/tajo/catalog/CatalogServer.java
```diff
@@ -1043,6 +1043,90 @@ public GetTablePartitionsResponse getAllPartitions(RpcController controller, Nul
...
     @Override
+    public GetPartitionsResponse getPartitionsByAlgebra(RpcController controller,
+      PartitionsByAlgebraProto request) throws ServiceException {
+      String dbName = request.getDatabaseName();
+      String tbName = request.getTableName();
+
+      try {
+        // linked meta data do not support partition.
+        // So, the request that wants to get partitions in this db will be failed.
+        if (linkedMetadataManager.existsDatabase(dbName)) {
```
jihoonson: These codes can be much simplified. Please refer to the ```existsTable()``` method of this class.

tajo-catalog/tajo-catalog-server/src/main/java/org/apache/tajo/catalog/CatalogServer.java
```diff
@@ -1043,6 +1043,90 @@ public GetTablePartitionsResponse getAllPartitions(RpcController controller, Nul
...
+    @Override
+    public GetPartitionsResponse getPartitionsByFilter(RpcController controller,
+                                                 PartitionsByFilterProto request) throws ServiceException {
+      String dbName = request.getDatabaseName();
+      String tbName = request.getTableName();
+
+      try {
+        // linked meta data do not support partition.
+        // So, the request that wants to get partitions in this db will be failed.
+        if (linkedMetadataManager.existsDatabase(dbName)) {
```
jihoonson: These codes can be much simplified. Please refer to the ```existsTable()``` method of this class.

tajo-catalog/tajo-catalog-server/src/main/java/org/apache/tajo/catalog/store/AbstractDBStore.java
```diff
@@ -2147,6 +2154,8 @@ private void ensurePartitionTable(String tbName, int tableId)
         builder.setId(res.getInt(COL_PARTITIONS_PK));
         builder.setPath(res.getString("PATH"));
         builder.setPartitionName(partitionName);
+        builder.setNumBytes(res.getLong(COL_PARTITION_BYTES));
+        builder.setNumFiles(res.getLong(COL_PARTITION_FILES));
```
jihoonson: I wonder why number of files in a partition is required.
Can you tell me an example?

tajo-catalog/tajo-catalog-server/src/main/java/org/apache/tajo/catalog/store/AbstractDBStore.java
```diff
@@ -2228,6 +2244,257 @@ private void setPartitionKeys(int pid, PartitionDescProto.Builder partitionDesc)
...
+   * @return
+   */
+  public boolean existPartitionsOnCatalog(int tableId) {
+    Connection conn = null;
+    ResultSet res = null;
+    PreparedStatement pstmt = null;
+    boolean result = false;
+
+    try {
+      String sql = "SELECT COUNT(*) CNT FROM "
```
jihoonson: This is trivial, but you can perform a simple select rather than count aggregation which is more expensive operation.

tajo-catalog/tajo-catalog-drivers/tajo-hive/src/main/java/org/apache/tajo/catalog/store/HiveCatalogStore.java
jihoonson: I found it.

tajo-catalog/tajo-catalog-server/src/main/java/org/apache/tajo/catalog/store/CatalogStore.java
```diff
@@ -97,13 +97,31 @@ boolean existPartitionMethod(String databaseName, String tableName) throws Undef
...
+   *
+   * @param request contains database name, table name, algebra expressions, parameter for executing PrepareStatement
+   * @return list of TablePartitionProto
+   * @throws UndefinedDatabaseException, UndefinedTableException, UndefinedPartitionMethodException
+   */
+  List<PartitionDescProto> getPartitionsByAlgebra(PartitionsByAlgebraProto request) throws
+    UndefinedDatabaseException, UndefinedTableException, UndefinedPartitionMethodException,
+    UndefinedOperatorException, PartitionNotFoundException, UnsupportedException;
+
+  List<PartitionDescProto> getPartitionsByFilter(PartitionsByFilterProto request) throws
```
jihoonson: You need to add a description for this method.

tajo-catalog/tajo-catalog-server/src/main/java/org/apache/tajo/catalog/store/AbstractDBStore.java
```diff
@@ -2228,6 +2244,257 @@ private void setPartitionKeys(int pid, PartitionDescProto.Builder partitionDesc)
...
+      throw new TajoInternalError(se);
+    } finally {
+      CatalogUtil.closeQuietly(pstmt, res);
+    }
+    return result;
+  }
+
+  @Override
+  public List<PartitionDescProto> getPartitionsByFilter(PartitionsByFilterProto request) {
+    throw new TajoRuntimeException(new UnsupportedException());
```
jihoonson: ```UnsupportedException``` means that we don't support this feature. But, this is not true because we support partition pruning with simple filters as well as more complex conditions.
So, it would be good if this method just call ```getPartitionsByAlgebra()``` in the below.

tajo-catalog/tajo-catalog-server/src/main/java/org/apache/tajo/catalog/store/AbstractDBStore.java
blrunner: Honestly, I added num of bytes and number of files by reference to PartitionedTableRewriter::updateTableStat. But now, it looks like unnecessary information.

tajo-catalog/tajo-catalog-server/src/main/java/org/apache/tajo/catalog/store/AbstractDBStore.java
```diff
@@ -2228,6 +2244,257 @@ private void setPartitionKeys(int pid, PartitionDescProto.Builder partitionDesc)
...
+        builder.setId(res.getInt(COL_PARTITIONS_PK));
+        builder.setPartitionName(res.getString("PARTITION_NAME"));
+        builder.setPath(res.getString("PATH"));
+        builder.setNumBytes(res.getLong(COL_PARTITION_BYTES));
+        builder.setNumFiles(res.getLong(COL_PARTITION_FILES));
+
+        partitions.add(builder.build());
+      }
+    } catch (TajoException se) {
+      throw new TajoInternalError(se);
```
jihoonson: All UndefinedDatabaseException, UndefinedTableException, UndefinedPartitionMethodException, and PartitionNotFoundException exceptions are subclass of TajoException. This code will prevent these exceptions from propagating properly.

tajo-catalog/tajo-catalog-server/src/main/java/org/apache/tajo/catalog/store/AbstractDBStore.java
```diff
@@ -2228,6 +2244,257 @@ private void setPartitionKeys(int pid, PartitionDescProto.Builder partitionDesc)
...
+   *
+   * @param partitionColumns
+   * @param json
+   * @param filterSets
+   * @return
+   * @throws TajoException
+   * @throws SQLException
+   */
+  private String getSelectStatementForPartitions(String tableName, List<ColumnProto> partitionColumns, String json,
+    List<PartitionFilterSet> filterSets) throws TajoException {
```
jihoonson: At least, you should specify that ```filterSets``` is one of the results of this method.

tajo-catalog/tajo-catalog-server/src/main/java/org/apache/tajo/catalog/store/AbstractDBStore.java
```diff
@@ -2228,6 +2244,257 @@ private void setPartitionKeys(int pid, PartitionDescProto.Builder partitionDesc)
...
+   *  SELECT A.PARTITION_ID, A.PARTITION_NAME, A.PATH FROM PARTITIONS A
+   *  WHERE A.TID = ?
+   *  AND A.PARTITION_ID IN (
+   *    SELECT T1.PARTITION_ID FROM PARTITION_KEYS T1
+   *    JOIN PARTITION_KEYS T2 ON T1.TID=T2.TID AND T1.PARTITION_ID = T2.PARTITION_ID AND T2.TID = ?
+   *    AND ( T2.COLUMN_NAME = 'col2' AND T2.PARTITION_VALUE IS NOT NULL )
+   *    JOIN PARTITION_KEYS T3 ON T1.TID=T3.TID AND T1.PARTITION_ID = T3.PARTITION_ID AND T3.TID = ?
+   *    AND ( T3.COLUMN_NAME = 'col3' AND T3.PARTITION_VALUE > ? )
+   *    WHERE T1.TID = ? AND ( T1.COLUMN_NAME = 'col1' AND T1.PARTITION_VALUE = ? )
+   *    OR ( T1.COLUMN_NAME = 'col1' AND T1.PARTITION_VALUE = ? )
```
jihoonson: I cannot understand this example. Would you tell me why did you use expensive self-join on partition_keys? This definitely degrades the performance as the more partitions are stored.

tajo-catalog/tajo-catalog-server/src/main/java/org/apache/tajo/catalog/store/CatalogStore.java
```diff
@@ -97,13 +97,31 @@ boolean existPartitionMethod(String databaseName, String tableName) throws Undef
    * @return
    * @throws TajoException
    */
-  List<CatalogProtos.PartitionDescProto> getPartitions(String databaseName, String tableName) throws
-      UndefinedDatabaseException, UndefinedTableException, UndefinedPartitionMethodException;
+  List<CatalogProtos.PartitionDescProto> getAllPartitions(String databaseName, String tableName) throws
```
jihoonson: I think ```getPartitionsOfTable()``` is a more appropriate name.

tajo-catalog/tajo-catalog-server/src/main/resources/schemas/derby/derby.xml
```diff
@@ -173,7 +174,9 @@
   				TID INT NOT NULL REFERENCES TABLES (TID) ON DELETE CASCADE,
   				PARTITION_NAME VARCHAR(767),
   				PATH VARCHAR(1024),
-  				CONSTRAINT C_PARTITIONS_PK PRIMARY KEY (PARTITION_ID),
+  				NUM_BYTES BIGINT,
+  				NUM_FILES BIGINT,
+					CONSTRAINT C_PARTITIONS_PK PRIMARY KEY (PARTITION_ID),
```
jihoonson: Please fix the indent.

tajo-catalog/tajo-catalog-server/src/main/resources/schemas/mariadb/mariadb.xml
```diff
@@ -150,6 +151,8 @@
           TID INT NOT NULL,
           PARTITION_NAME VARCHAR(255) BINARY,
           PATH VARCHAR(4096) BINARY,
+  				NUM_BYTES BIGINT,
+  				NUM_FILES BIGINT,
```
jihoonson: Please fix the indent.

tajo-catalog/tajo-catalog-server/src/main/resources/schemas/mysql/mysql.xml
```diff
@@ -154,6 +155,8 @@
           TID INT NOT NULL,
           PARTITION_NAME VARCHAR(255) BINARY,
           PATH VARCHAR(4096) BINARY,
+  				NUM_BYTES BIGINT,
```
jihoonson: Please fix the indent.

tajo-core/src/main/java/org/apache/tajo/querymaster/Query.java
```diff
@@ -509,6 +510,18 @@ private QueryState finalizeQuery(Query query, QueryCompletedEvent event) {
...
+            List<PartitionDescProto> finalPartitions = TUtil.newList();
+            FileSystem fileSystem = finalOutputDir.getFileSystem(query.systemConf);
+            for (PartitionDescProto partition : partitions) {
+              PartitionDescProto.Builder builder = partition.toBuilder();
+              Path partitionPath = new Path(finalOutputDir, partition.getPath());
+              ContentSummary contentSummary = fileSystem.getContentSummary(partitionPath);
+              builder.setNumBytes(contentSummary.getLength());
+              builder.setNumFiles(contentSummary.getFileCount());
+              finalPartitions.add(builder.build());
+            }
```
jihoonson: It would be better to modularize this routine.

tajo-plan/src/main/java/org/apache/tajo/plan/expr/AlgebraicUtil.java
```diff
@@ -493,4 +497,190 @@ public Object visitTimeLiteral(Object ctx, Stack<Expr> stack, TimeLiteral expr)
...
+    ExprFinder finder = new ExprFinder(type);
+    finder.visit(null, new Stack<Expr>(), expr);
+
+    if (finder.getFoundExprs().size() == 0) {
+      return null;
+    }
+    return (T) finder.getFoundExprs().get(0);
+  }
+
+  private static class ExprFinder extends SimpleAlgebraVisitor<Object, Expr> {
```
jihoonson: We have already ExprFinder.

tajo-plan/src/main/java/org/apache/tajo/plan/expr/AlgebraicUtil.java
```diff
@@ -493,4 +497,190 @@ public Object visitTimeLiteral(Object ctx, Stack<Expr> stack, TimeLiteral expr)
...
+      return super.visit(ctx, stack, expr);
+    }
+
+    public List<Expr> getFoundExprs() {
+      return list;
+    }
+
+  }
+
+  public static Expr[] toConjunctiveNormalFormArray(Expr expr) {
```
jihoonson: Unnecessary change.

tajo-plan/src/main/java/org/apache/tajo/plan/expr/AlgebraicUtil.java
```diff
@@ -493,4 +497,190 @@ public Object visitTimeLiteral(Object ctx, Stack<Expr> stack, TimeLiteral expr)
...
+    List<Expr> list = new ArrayList<Expr>();
+    toConjunctiveNormalFormArrayRecursive(expr, list);
+    return list.toArray(new Expr[list.size()]);
+  }
+
+  private static void toConjunctiveNormalFormArrayRecursive(Expr node, List<Expr> found) {
```
jihoonson: Unnecessary change.

tajo-plan/src/main/java/org/apache/tajo/plan/expr/AlgebraicUtil.java
```diff
@@ -493,4 +497,190 @@ public Object visitTimeLiteral(Object ctx, Stack<Expr> stack, TimeLiteral expr)
...
+  /**
+   * It finds unique columns from a Expr.
+   */
+  public static LinkedHashSet<ColumnReferenceExpr> findUniqueColumnReferences(Expr expr) throws TajoException {
+    UniqueColumnReferenceFinder finder = new UniqueColumnReferenceFinder();
+    finder.visit(null, new Stack<Expr>(), expr);
+    return finder.getColumnRefs();
+  }
+
+  private static class UniqueColumnReferenceFinder extends SimpleAlgebraVisitor<Object, Expr> {
```
jihoonson: Why did you create this class instead of using ExprFinder? Is there any reason?

tajo-plan/src/main/java/org/apache/tajo/plan/expr/AlgebraicUtil.java
```diff
@@ -493,4 +497,190 @@ public Object visitTimeLiteral(Object ctx, Stack<Expr> stack, TimeLiteral expr)
...
+
+      Expr filterPerLevel = AlgebraicUtil.createSingletonExprFromCNFByExpr(
+        accumulatedFilters.toArray(new Expr[accumulatedFilters.size()]));
+      filters[i] = filterPerLevel;
+    }
+
+    return filters;
+  }
+
+  public static Expr createSingletonExprFromCNFByExpr(Collection<Expr> cnfExprs) {
```
jihoonson: What is the difference between this method and createSingletonExprFromCNF()?

tajo-plan/src/main/java/org/apache/tajo/plan/expr/AlgebraicUtil.java
```diff
@@ -493,4 +497,190 @@ public Object visitTimeLiteral(Object ctx, Stack<Expr> stack, TimeLiteral expr)
...
+   * The third Expr: col3 = 'C'
+   *
+   * 'IS NOT NULL' predicate is always true against the partition path.
+   *
+   *
+   * @param partitionColumns
+   * @param conjunctiveForms
+   * @return
+   */
+  public static Expr[] getAccumulatedFiltersByExpr(String tableName,
```
jihoonson: Method name is not intuitive. This method just rearrange the order of cnfs.

tajo-storage/tajo-storage-hdfs/src/main/java/org/apache/tajo/storage/FileTablespace.java
```diff
@@ -315,15 +315,21 @@ public boolean accept(Path path) {
...
    * Subclasses may override to, e.g., select only files matching a regular
    * expression.
    *
+   * @param check if target table is a partitioned table
+   * @param input directories
    * @return array of FileStatus objects
    * @throws IOException if zero items.
    */
-  protected List<FileStatus> listStatus(Path... dirs) throws IOException {
+  protected List<FileStatus> listStatus(boolean hasPartition, Path... dirs) throws IOException {
```
jihoonson: This method is declared as ```protected``` and not used only in ```getSplits()``` method. So, I think that we don't have to maintain two separate methods according to the ```hasPartition``` flag.

tajo-catalog/tajo-catalog-common/src/main/java/org/apache/tajo/catalog/CatalogService.java
```diff
@@ -172,7 +172,15 @@ PartitionDescProto getPartition(String databaseName, String tableName, String pa
       throws UndefinedPartitionException, UndefinedPartitionMethodException, UndefinedDatabaseException,
       UndefinedTableException;
 
-  List<PartitionDescProto> getPartitions(String databaseName, String tableName);
+  List<PartitionDescProto> getAllPartitions(String databaseName, String tableName) throws UndefinedDatabaseException,
```
jihoonson: IMO, this method must not throw PartitionNotFoundException, but return an empty list.

tajo-catalog/tajo-catalog-common/src/main/java/org/apache/tajo/catalog/CatalogService.java
```diff
@@ -172,7 +172,15 @@ PartitionDescProto getPartition(String databaseName, String tableName, String pa
...
+    UndefinedTableException, UndefinedPartitionMethodException, PartitionNotFoundException;
+
+  List<PartitionDescProto> getPartitionsByAlgebra(PartitionsByAlgebraProto request) throws
```
jihoonson: IMO, this method must not throw PartitionNotFoundException, but return an empty list.

tajo-catalog/tajo-catalog-common/src/main/java/org/apache/tajo/catalog/CatalogService.java
```diff
@@ -172,7 +172,15 @@ PartitionDescProto getPartition(String databaseName, String tableName, String pa
...
+    UndefinedDatabaseException, UndefinedTableException, UndefinedPartitionMethodException,
+    PartitionNotFoundException, UnsupportedException;
+
+  List<PartitionDescProto> getPartitionsByFilter(PartitionsByFilterProto request) throws
```
jihoonson: IMO, this method must not throw PartitionNotFoundException, but return an empty list.

tajo-catalog/tajo-catalog-server/src/main/java/org/apache/tajo/catalog/store/AbstractDBStore.java
blrunner: I'll remove it.

tajo-storage/tajo-storage-hdfs/src/main/java/org/apache/tajo/storage/FileTablespace.java
blrunner: I agree with your point. I've just removed ```getSplits()```.

tajo-catalog/tajo-catalog-server/src/main/java/org/apache/tajo/catalog/store/AbstractDBStore.java
blrunner: If we want to provide ```getPartitionsByFilter()``` by calling ```getPartitionsByAlgebra()```, we need to convert filter conditions to algebra expressions. I need more time to implement this feature and I wish to implement it at another issue.

tajo-catalog/tajo-catalog-drivers/tajo-hive/src/main/java/org/apache/tajo/catalog/store/HiveCatalogStore.java
blrunner: You're right. We need to handle above situation.

tajo-plan/src/main/java/org/apache/tajo/plan/expr/AlgebraicUtil.java
blrunner: The method is different from existing similar method because of input parameter. I added more description.

tajo-catalog/tajo-catalog-server/src/main/java/org/apache/tajo/catalog/store/AbstractDBStore.java
blrunner: I also think that self-join on partition_keys is expensive way. But it might need more time to improve the select statement and according to the need, we would change partition schema. So, I want to handle this issue at another issue. What do you think about my opinion? For your information, you can see the performance of this method at https://github.com/apache/tajo/pull/624#issuecomment-127275276 .

blrunner:
@jihoonson 

Thank you for your detailed review. I reflected your comments and left some comments.

tajo-catalog/tajo-catalog-common/src/main/java/org/apache/tajo/catalog/CatalogUtil.java
```diff
@@ -812,16 +812,34 @@ public static AlterTableDesc setProperty(String tableName, KeyValueSet params, A
...
    * @param values partition values
-   * @param location partition location
+   * @param path partition directory path
    * @param alterTableType ADD_PARTITION or DROP_PARTITION
+   * @param numBytes contents length
    * @return AlterTableDesc
    */
   public static AlterTableDesc addOrDropPartition(String tableName, String[] columns,
-                                            String[] values, String location, AlterTableType alterTableType) {
+      String[] values, @Nullable String path, AlterTableType alterTableType, long numBytes) {
```
jihoonson: Please fix indentation.

tajo-catalog/tajo-catalog-common/src/main/java/org/apache/tajo/catalog/CatalogUtil.java
```diff
@@ -812,16 +812,34 @@ public static AlterTableDesc setProperty(String tableName, KeyValueSet params, A
...
+   * @param tableName
+   * @param columns
+   * @param values
+   * @param path
+   * @param alterTableType
+   * @return
+   */
+  public static AlterTableDesc addOrDropPartition(String tableName, String[] columns,
+                                                  String[] values, @Nullable String path, AlterTableType
+                                                    alterTableType) {
```
jihoonson: Please fix indentation.

tajo-catalog/tajo-catalog-common/src/main/proto/CatalogProtos.proto
```diff
@@ -262,6 +262,18 @@ message PartitionIdentifierProto {
   optional string partitionName = 3;
 }
 
+message PartitionsByAlgebraProto {
+  required string databaseName = 1;
+  required string tableName = 2;
+  required string algebra = 3; /** json object which contains algebra expressions */
```
jihoonson: Please use ```//``` instead of ```/**/```.

tajo-catalog/tajo-catalog-common/src/main/proto/CatalogProtos.proto
```diff
@@ -262,6 +262,18 @@ message PartitionIdentifierProto {
...
+}
+
+message PartitionsByFilterProto {
+  required string databaseName = 1;
+  required string tableName = 2;
+  required string filter = 3; /** filter condition, for example col1 = 'TAJO' and col2 = 100. */
```
jihoonson: Please use ```//``` instead of ```/**/```. 
In addition, which format is used for filter serialization?

tajo-catalog/tajo-catalog-drivers/tajo-hive/src/main/java/org/apache/tajo/catalog/store/HiveCatalogStore.java
```diff
@@ -845,13 +856,174 @@ public boolean existPartitionMethod(String databaseName, String tableName) throw
...
+          if (addedFilter > 0) {
```
jihoonson: ```addedFilter``` can be replaced with sb.length().

tajo-catalog/tajo-catalog-server/src/main/java/org/apache/tajo/catalog/store/AbstractDBStore.java
```diff
@@ -2228,6 +2238,260 @@ private void setPartitionKeys(int pid, PartitionDescProto.Builder partitionDesc)
     return partitions;
   }
 
+  /**
+   * Check if list of partitions exist on catalog.
+   *
+   *
+   * @param databaseId
```
jihoonson: Wrong parameter.

tajo-catalog/tajo-catalog-server/src/main/java/org/apache/tajo/catalog/store/AbstractDBStore.java
jihoonson: I got your point. 
If so, please update the comments on both ```CatalogStore::getPartitionsByAlgebra()``` and ```CatalogStore::getPartitionsByFilter()``` to let others know each method is implemented in AbstractDBStore and HiveCatalogStore, respectively.

jihoonson: And please open a jira ticket for this issue.

tajo-catalog/tajo-catalog-server/src/main/java/org/apache/tajo/catalog/store/AbstractDBStore.java
jihoonson: Ok. Please open a jira ticket for this issue.

tajo-catalog/tajo-catalog-server/src/main/resources/schemas/mariadb/mariadb.xml
```diff
@@ -150,7 +151,8 @@
           TID INT NOT NULL,
           PARTITION_NAME VARCHAR(255) BINARY,
           PATH VARCHAR(4096) BINARY,
-          UNIQUE INDEX PARTITION_UNIQUE_IDX (TID, PARTITION_NAME),
+  				NUM_BYTES BIGINT,
```
jihoonson: Please fix the indentation.

tajo-catalog/tajo-catalog-server/src/main/resources/schemas/mysql/mysql.xml
```diff
@@ -154,7 +155,8 @@
           TID INT NOT NULL,
           PARTITION_NAME VARCHAR(255) BINARY,
           PATH VARCHAR(4096) BINARY,
-          UNIQUE INDEX PARTITION_UNIQUE_IDX (TID, PARTITION_NAME),
+  				NUM_BYTES BIGINT,
```
jihoonson: Please fix the indentation.

tajo-plan/src/main/java/org/apache/tajo/plan/expr/AlgebraicUtil.java
jihoonson: Sorry, my mistake.

tajo-storage/tajo-storage-hdfs/src/main/java/org/apache/tajo/storage/FileTablespace.java
```diff
@@ -343,7 +343,7 @@ public boolean accept(Path path) {
 
       FileStatus[] matches = fs.globStatus(p, inputFilter);
       if (matches == null) {
-        errors.add(new IOException("Input path does not exist: " + p));
+        LOG.warn("Input path does not exist: " + p);
       } else if (matches.length == 0) {
         errors.add(new IOException("Input Pattern " + p + " matches 0 files"));
```
jihoonson: This should also not be an error.

tajo-common/src/main/java/org/apache/tajo/exception/PartitionNotFoundException.java
```diff
@@ -0,0 +1,35 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.tajo.exception;
+
+import org.apache.tajo.error.Errors.ResultCode;
+import org.apache.tajo.rpc.protocolrecords.PrimitiveProtos.ReturnState;
+
+public class PartitionNotFoundException extends TajoException {
```
jihoonson: This exception is duplicated with ```UndefinedPartitionException```.

tajo-catalog/tajo-catalog-server/src/main/java/org/apache/tajo/catalog/store/AbstractDBStore.java
```diff
@@ -2228,6 +2238,260 @@ private void setPartitionKeys(int pid, PartitionDescProto.Builder partitionDesc)
...
+    List<PartitionFilterSet> filterSets = null;
+
+    int databaseId = getDatabaseId(request.getDatabaseName());
+    int tableId = getTableId(databaseId, request.getDatabaseName(), request.getTableName());
+    if (!existPartitionMethod(request.getDatabaseName(), request.getTableName())) {
+      throw new UndefinedPartitionMethodException(request.getTableName());
+    }
+
+    if (!existPartitionsOnCatalog(tableId)) {
+      throw new PartitionNotFoundException(request.getTableName());
```
jihoonson: Why does this method still throw ```PartitionNotFoundException```? Do you have any reason?

tajo-storage/tajo-storage-hdfs/src/main/java/org/apache/tajo/storage/FileTablespace.java
blrunner: If so, how about update ```LOG.warn``` to ```LOG.info```?

tajo-catalog/tajo-catalog-server/src/main/java/org/apache/tajo/catalog/store/AbstractDBStore.java
blrunner: ```CatalogStore::getPartitionsOfTable``` can returns empty list instead of ```PartitionNotFoundException```. But if ```CatalogStore::getPartitionsByAlgebra``` doesn't return ```PartitionNotFoundException```, ```PartitionedTableRewriter``` might not check in case of hdfs listing need.

tajo-common/src/main/java/org/apache/tajo/exception/PartitionNotFoundException.java
blrunner: ```UndefinedPartitionException``` means that a specified partition doesn't exist on catalog. And ```PartitionNotFoundException`` means that there are no partitions on catalog.

tajo-storage/tajo-storage-hdfs/src/main/java/org/apache/tajo/storage/FileTablespace.java
jihoonson: Would you share any reason?
I think that users can easily recognize this case even though it isn't regarded as invalid.

tajo-common/src/main/java/org/apache/tajo/exception/PartitionNotFoundException.java
jihoonson: If so, I think that this exception is not necessary.

tajo-catalog/tajo-catalog-server/src/main/java/org/apache/tajo/catalog/store/AbstractDBStore.java
jihoonson: It's difficult for me to understand why those implementations are different. 
The exception means an invalid situation. So, basically, exceptions should be handled as error cases. However, the case you explained above is not an error case. 

In addition, this exception is never exposed to users, but used only in PartitionedTableRewriter. I think this can be replaced with a better routine.

blrunner: These implementations are just for work loads. If users recover their partitions using https://github.com/apache/tajo/pull/626, this exception is never exposed to users. If we notice this point to users and decide to ignore above situation, we can remove these implementations.

tajo-storage/tajo-storage-hdfs/src/main/java/org/apache/tajo/storage/FileTablespace.java
blrunner: If users can easily recognize this case, it doesn't matter.

tajo-catalog/tajo-catalog-server/src/main/java/org/apache/tajo/catalog/store/AbstractDBStore.java
blrunner: Thank you for your understanding. Could you check the comments on CatalogStore?

blrunner:
Thanks @jihoonson 

I've reflected your comments and I'm going to create jia issues about ```CatalogStore::getPartitionsByFilter``` and performance of the select statement for getting partitions after finishing this PR.

tajo-cli/src/main/java/org/apache/tajo/cli/tools/TajoDump.java
```diff
@@ -208,6 +208,7 @@ private static void dumpDatabase(TajoClient client, String databaseName, PrintWr
         }
         writer.write("\n\n");
       } catch (Exception e) {
+        e.printStackTrace();
```
jihoonson: Please remove this line.

tajo-common/src/main/java/org/apache/tajo/exception/UndefinedPartitionsException.java
```diff
@@ -0,0 +1,35 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.tajo.exception;
+
+import org.apache.tajo.error.Errors.ResultCode;
+import org.apache.tajo.rpc.protocolrecords.PrimitiveProtos.ReturnState;
+
+public class UndefinedPartitionsException extends TajoException {
```
jihoonson: This exception is not used.

tajo-plan/src/main/java/org/apache/tajo/plan/rewrite/rules/PartitionedTableRewriter.java
```diff
@@ -118,13 +120,76 @@ public String toString() {
...
+      } else {
+        if (catalog.existPartitions(splits[0], splits[1])) {
+          PartitionsByAlgebraProto request = getPartitionsAlgebraProto(splits[0], splits[1], conjunctiveForms);
+          partitions = catalog.getPartitionsByAlgebra(request);
+          filteredPaths = findFilteredPathsByPartitionDesc(partitions);
+        } else {
+          filteredPaths = findFilteredPathsFromFileSystem(partitionColumns, conjunctiveForms, fs, tablePath);
+        }
+      }
+    } catch (UnsupportedException ue) {
```
jihoonson: IMO, this case is regarded as the full scan on partitions.
So, the same routine with the above ```if (conjunctiveForms == null)``` clause is required here.

tajo-storage/tajo-storage-hdfs/src/main/java/org/apache/tajo/storage/FileTablespace.java
```diff
@@ -342,18 +342,18 @@ public boolean accept(Path path) {
...
-        errors.add(new IOException("Input path does not exist: " + p));
-      } else if (matches.length == 0) {
-        errors.add(new IOException("Input Pattern " + p + " matches 0 files"));
-      } else {
-        for (FileStatus globStat : matches) {
-          if (globStat.isDirectory()) {
-            for (FileStatus stat : fs.listStatus(globStat.getPath(), inputFilter)) {
-              result.add(stat);
+      if (matches != null) {
+        if (matches.length == 0) {
```
jihoonson: If ```matches``` is null or ```matches.length``` is 0, you need to print a warning message.

blrunner:
Thanks @jihoonson .
I've just reflected your comments. Could you check it again? :)

blrunner:
I've updated the patch as following:

* **Remove ```CatalogStore::getPartitionsByFilter```**: AbstractDBStore doesn't support this method. Also HiveCatalogStore support the method restrictedly. You can see the grammar for hive partition api to https://github.com/apache/hive/blob/master/metastore/src/java/org/apache/hadoop/hive/metastore/parser/Filter.g. But I think that we need the method for user convenience. So I'll implement at another jira issue using tajo-sql-parser module and ```CatalogStore::getPartitionsByAlgebra```.

* **Allow invalid partition directory name**: When adding partition using alter statement, users can set partition path freely. But current PartitionedTableRewriter just allow well-defined partition directory which consists of column name and partition value. So if users use invalid partition directory name, they can scan their partitions. To resolve above situation, I added list of PartitionDescProto to PartitionedTableScanNode.

blrunner:
All unit test finished successfully on my laptop. And I tested this patch successfully as following:

* Data: TPC-H 1G
* Query: Q1, Q3, Q4, Q5, Q8
* Catalog : MySQLStore, HiveCatalogStore
* Table schema:

```
CREATE TABLE customer (c_custkey INT8, c_name TEXT, c_address TEXT, c_phone TEXT, c_acctbal FLOAT8, c_mktsegment TEXT, c_comment TEXT) USING TEXT  PARTITION BY COLUMN(c_nationkey INT8) ;

CREATE TABLE lineitem (l_orderkey INT8, l_partkey INT8, l_suppkey INT8, l_linenumber INT8, l_quantity FLOAT8, l_extendedprice FLOAT8, l_discount FLOAT8, l_tax FLOAT8, l_commitdate DATE, l_receiptdate DATE, l_shipinstruct TEXT, l_shipmode TEXT, l_comment TEXT) USING TEXT  PARTITION BY COLUMN(l_shipdate DATE, l_returnflag TEXT, l_linestatus TEXT) ;

CREATE TABLE nation (n_nationkey INT8, n_name TEXT, n_comment TEXT) USING TEXT  PARTITION BY COLUMN(n_regionkey INT8) ;

CREATE TABLE orders (o_orderkey INT8, o_custkey INT8, o_totalprice FLOAT8, o_clerk TEXT, o_shippriority INT4, o_comment TEXT) USING TEXT  PARTITION BY COLUMN(o_orderdate DATE, o_orderstatus TEXT, o_orderpriority TEXT) ;

CREATE TABLE part (p_partkey INT8, p_name TEXT, p_mfgr TEXT, p_brand TEXT, p_type TEXT, p_container TEXT, p_retailprice FLOAT8, p_comment TEXT) USING TEXT PARTITION BY COLUMN(p_size INT4) ;

CREATE TABLE partsupp (ps_partkey INT8, ps_suppkey INT8, ps_availqty INT4, ps_supplycost FLOAT8, ps_comment TEXT) USING TEXT ;

CREATE TABLE region (r_name TEXT, r_comment TEXT) USING TEXT  PARTITION BY COLUMN(r_regionkey INT8);

CREATE TABLE supplier (s_suppkey INT8, s_name TEXT, s_address TEXT, s_phone TEXT, s_acctbal FLOAT8, s_comment TEXT) USING TEXT PARTITION BY COLUMN(s_nationkey INT8);
```

jihoonson:
+1 the latest patch looks good to me.
Thank you for your effort!

blrunner:
@jihoonson 

Thank you for your detailed review. I'll ship it. :)

PR #773, opened 2015-09-21
TAJO-1872: Increase the minimum split size and add a classpath to hadoop tools

jihoonson:


jinossy:
+1 LGTM

jihoonson:
Thanks for the quick review.

PR #775, opened 2015-09-21
TAJO-1807: Not display "Finished Queries" in query.jsp

hys9958:
Hi,
I have implemented for this issue.
Thanks!


jihoonson:
Hi @hys9958, thank you for your work.
I'm wondering ```hsync()``` does not work properly. If you have already tried, please share the result.

jinossy:
@hys9958
You should set history file path, if you can't use the system default 
```
HISTORY_QUERY_DIR("tajo.history.query.dir", "file:///{your-query-history-directtory}/history")
```

PR #776, opened 2015-09-22
TAJO-1827: JSON parsing error at storage-site.json while tajo master starts up

blrunner:
Removed comments in the template file.


hyunsik:
How about removing all contents from the template file? The default template will add a tablespace as soon as a user just copies ``storage-site.json`` from ``storage-site.json.template``. It may cause some problem.

blrunner:
Thanks @hyunsik .
I've reflected your comments.

hyunsik:
+1
The patch looks good to me.

blrunner:
Thanks @hyunsik 

I added a description element to JSON template instead of removing contents. If possible, could you check it again?

blrunner:
@hyunsik 

On second thought, the template file might cause a problem when starting HBaseTableSpace. I'll shit previous version. :)

hyunsik:
+1
Ship it. the patch looks good to me.

PR #778, opened 2015-09-23
TAJO-1883: Keep tajo-client and its dependent modules to be 1.7 target.

hyunsik:


jihoonson:
Sorry for late review. This patch looks good to me. Would you rebase it?

hyunsik:
Just rebased.

jihoonson:
Thank you for the quick rebase. I have two questions.
* ```tajo-jdbc``` has a dependency on ```tajo-cluster-tests```, but ```tajo-cluster-tests```'s target is 1.8. Is it ok?
* Both ```tajo-client``` and ```tajo-jdbc``` have some nested dependencies. For example, ```tajo-client``` has a dependency on ```tajo-rpc-protobuf```, and which in turn it has a dependency on ```tajo-rpc-common```. In this case, I wonder that it is ok if ```tajo-rpc-common```'s target is still 1.8.

hyunsik:
Thank you for your quick review. Here is my response:
```tajo-cluster-tests``` is ok because 1.8 has backward compatibility.
```tajo-rpc-common``` must be 1.7 too. I'll fix it.

hyunsik:
I just fixed and rebased it.

jihoonson:
Thank you for your reply and quick fix.
+1!

PR #779, opened 2015-09-23
TAJO-1884: Add missing jetty-util dependency.

jinossy:


hyunsik:
+1
I've tested it. It works well.

PR #780, opened 2015-09-23
TAJO-1877: Adopt try-resource statement to AbstractDBStore

dkhwangbo:
See https://docs.oracle.com/javase/tutorial/essential/exceptions/tryResourceClose.html. 
JDK 7 introduced try-resource statement for better resource allocation and deallocation. 
I adopt this feature to AbstractDBStore.


jihoonson:
Sorry for late review. This patch looks good to me. Would you rebase it?

dkhwangbo:
@jihoonson Thanks for your reply. I just rebase this commit but some other problem is occured.
I'll create PR again after resolving all of the problem.

PR #781, opened 2015-09-23
TAJO-1886: Set compilerSourceVM and compilerTargetVM to web.xml.

hyunsik:


blrunner:
+1 LGTM

I've found that this patch ran as expected on testing cluster.

PR #782, opened 2015-09-24
TAJO-1885: Simple query with projection should be supported.

hyunsik:


blrunner:
+1 LGTM.
And it would be better if you commit this PR to 0.11.0 branch and master branch. :)

PR #784, opened 2015-09-24
TAJO-1889: UndefinedColumnException when a query with table subquery is executed on self-describing tables

jihoonson:


hyunsik:
+1
The patch looks good to me. The test looks failed, but it is not related to this patch.

jihoonson:
@hyunsik, thank you for your quick review, but I updated my patch to fix one more problem.
Would you please review again?

jihoonson:
I also put my patch on jira to trigger jenkins.

jihoonson:
I don't know why the diff includes the changes other than I modified.
I'll close and reopen this pr.

PR #785, opened 2015-09-24
TAJO-1887: Disable the alter table add partition statement temporarily.

blrunner:
Currently, users can use arbitary partition path using alter table statement. But they can't scan their partitions because PartitionedTableRewriter just build tuple from partition path. Fundamentally, we should allow arbitary partition path. But we need to avoid an exception by above situation until tajo allow arbitary partition path. So it seems that alter table add partition statement is disabled temporarily.


hyunsik:
Rather than commenting out for disabling the feature, I'd like to recommend to add some verification rule.

blrunner:
Thanks @hyunsik 

I reflected your comments. :)

tajo-plan/src/main/java/org/apache/tajo/plan/verifier/PreLogicalPlanVerifier.java
```diff
@@ -354,4 +354,14 @@ public Expr visitInsert(Context context, Stack<Expr> stack, Insert expr) throws
 
     return expr;
   }
+
+  // TODO: This should be removed at TAJO-1891
+  @Override
+  public Expr visitAlterTable(Context context, Stack<Expr> stack, AlterTable expr) throws TajoException {
+    if (expr.getAlterTableOpType() == AlterTableOpType.ADD_PARTITION) {
+      context.state.addVerification(new UnsupportedException("ADD PARTITION"));
```
hyunsik: It should be UnimplementedException. UnsupportedException means there is no plan to support this feature.

blrunner:
Thanks @hyunsik 
I've reflected your comments just now. :)

hyunsik:
Could you add negative tests for ``ALTER ADD PARTITION``?

blrunner:
I added negative unit test cases.

hyunsik:
Could you rebase it against the latest revision?

blrunner:
Rebased and updated the document for alter table statement.

hyunsik:
+1
@blrunner thank you for your contribution. The patch looks good to me.

PR #787, opened 2015-09-25
TAJO-1889: UndefinedColumnException when a query with table subquery is executed on self-describing tables

jihoonson:


hyunsik:
+1
I verified 'mvn clean install'. The patch looks good to me.

PR #788, opened 2015-09-25
TAJO-1892: Remove '_' as an identifier

dkhwangbo:
Use of '_' as an identifier might not be supported in releases after java se 8. 
so, we must check code in whole project.


hyunsik:
@dkhwangbo Thank you for your work. But, asm module is a thirdparty package. So, we don't need to touch this source code. Instead, we should keep asm package using 1.7. For it, I'll add the following configuration in #778.

```
      <artifactId>maven-compiler-plugin</artifactId>
      <configuration>
          <source>1.7</source>
          <target>1.7</target>
        </configuration>
```

hyunsik:
Could you close this ticket?

dkhwangbo:
@hyunsik Thanks for your reply. I act on your comment.

PR #791, opened 2015-09-25
TAJO-1894: Filter condition is ignored when a query involves multiple subqueries and aggregations

jihoonson:
The problem is that filter conditions which are not matched to the subquery schema are not restored after traversing that subquery in FFD.


hyunsik:
+1 LGTM!

PR #792, opened 2015-09-29
TAJO-1900: When a record column and its child column are retrieved together, the record column might not be inferred as record type properly

jihoonson:


jihoonson:
I verified my patch with 'mvn install'.

hyunsik:
It seems to need rebase.

jihoonson:
Rebased.

jihoonson:
I've verified with 'mvn clean install' with jdk7.

hyunsik:

jihoonson:
Thanks for your review.

PR #793, opened 2015-09-30
TAJO-1893: Remove unused import

dkhwangbo:
current version of TAJO has lots of unused import.
We can improve the performance of tajo with removing it.


jihoonson:
+1. Thank you for your contribution.

PR #794, opened 2015-09-30
TAJO-1899: Calling 'Arrays.asList()' with too few arguments cause memory extravagance

dkhwangbo:
Reports any calls to Arrays.asList() with zero arguments or only one argument.
Such calls could be replaced with either a call to Collections.singletonList() or Collections.emptyList() which will save some memory.


jihoonson:
+1 LGTM.

PR #795, opened 2015-09-30
TAJO-1902: Add line delimiter for repair partition in TajoDump.

blrunner:
I added line delimiter and verified that this patch had run as expected on testing cluster.


hyunsik:
+1 LGTM!

PR #796, opened 2015-09-30
TAJO-1901: Repair partition throws ArrayIndexOutOfBoundsException occasionally.

blrunner:
I fixed about repair partition and verified that this patch had run as expected on testing cluster.


blrunner:
@hyunsik 

I added more unit test cases for various situation. Could you check it again?

hyunsik:
I manually verified 'mvn clean install' more. But, the test failure is not related to this patch and the failure seems to be occur occasionally. So, here is my +1. The patch looks good to me and fixes the bug.

PR #799, opened 2015-10-05
TAJO-1906: Decrease the default size of hash map and array list for testing

jihoonson:


jihoonson:
I also removed unused MemSortExec.

hyunsik:
The patch looks good to me. It would be great if we use ``JOIN_HASH_TABLE_SIZE`` without ``TEST_`` prefix.

PR #800, opened 2015-10-05
TAJO-1907: Reduce the size of the INDEXES table's COLUMN_NAMES field

jihoonson:


jihoonson:
MySQL InnoDB allows the size of a single column for an index key up to 767 bytes by default.
Please refer to this document. https://dev.mysql.com/doc/refman/5.5/en/innodb-restrictions.html

jihoonson:
With out this patch, users will face an error message like ```Specified key was too long; max key length is 767.``` when starting up the tajo master. This patch must be included in the 0.11.0 release, I think.

hyunsik:
+1
This change makes sense. The patch looks good to me. I'll cancel the vote due to this reason. Thanks!

hyunsik:
I have a question. Should users backup and restore for it? If so, please share the breaking change in dev/user mailing list.

jihoonson:
Yes, I think it will be ok. Users don't have to backup and restore their catalog unless they have indexes on some columns where the length of the composite string of those column names is 256 bytes.

jihoonson:
I'll commit shortly.

PR #801, opened 2015-10-05
TAJO-1908: Add a link to the nightly build version to the download page

jihoonson:


hyunsik:
Could you add some kind of disclaimer about stability? An example is 'The nightly version comes from the current development branch. Note that this version may include unstable features and unknown bugs.'

jihoonson:
How about adding it to the wiki page (https://cwiki.apache.org/confluence/display/TAJO/Nightly+Builds)?

hyunsik:
Unless you add direct download link on the web, your suggestion sounds good to me.

jihoonson:
Thanks. I've added the disclaimer to the wiki page.

hyunsik:
+1 
Looks good to me.

jihoonson:
Thanks for your review.

PR #802, opened 2015-10-05
TAJO-1909: Eliminate remained explicit diamond expressions

dkhwangbo:
I found some remain part after working TAJO-1844.
This is another work of removing it.

### DIFF
--- a/tajo-catalog/tajo-catalog-common/src/main/java/org/apache/tajo/catalog/NestedPathUtil.java
+++ b/tajo-catalog/tajo-catalog-common/src/main/java/org/apache/tajo/catalog/NestedPathUtil.java
@@ -29,7 +29,7 @@ import java.util.*;
 public class NestedPathUtil {
   public static final String PATH_DELIMITER = "/";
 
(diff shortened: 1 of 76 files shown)


dkhwangbo:
Travis-ci gives me the error message but This commit pass the same test in my local environment.

hyunsik:
That's not your fault. I'll manually verify the patch and review it.

hyunsik:
Could you rebase the patch? The patch causes the following error:
```
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-compiler-plugin:3.3:compile (default-compile) on project tajo-common: Compilation failure
[ERROR] /home/hyunsik/Code/tajo/tajo/tajo-common/src/main/java/org/apache/tajo/util/ShutdownHookManager.java:[103,34] incompatible types: java.util.Set<java.lang.Object> cannot be converted to java.util.Set<org.apache.tajo.util.ShutdownHookManager.HookEntry>
```

#778 changes the source and target JVM of tajo-common to 1.7. It may cause this problem, I think.

dkhwangbo:
@hyunsik Thanks for your reply. I rebase it. Please review again.

PR #804, opened 2015-10-05
TAJO-1895: Add a maven profile to skip the unit tests of tajo-storage-pgsql

hyunsik:
You can disable tajo-storage-pgsql unit test by the profile ``-Ptest-storage-pgsql`` as follows:
```
mvn clean install -Ptest-storage-pgsql
```

The unit tests of tajo-storage-pgsql are disabled by default. Please refer to the discussion.
http://markmail.org/message/pdhz7u4r4rsfabz7

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -37,4 +37,4 @@ before_install: ulimit -t 514029 -u 2048 -n 3000
 install: ./dev-support/travis-install-dependencies.sh
 
 script: 
-  mvn clean install -q -ff -Dsurefire.useFile=false -Pparallel-test -DLOG_LEVEL=WARN -Dmaven.fork.count=2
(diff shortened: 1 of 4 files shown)


jihoonson:
Would you share some purpose of this ticket? Do we need to skip tests on pgsql storage sometimes?

hyunsik:
Please check out the discussion link. I proposed this workaround.

hyunsik:
Unless you add direct download link on the web, your suggestion sounds good to me.

jihoonson:
Oh, sorry. I missed the above link. I agree with your proposal. 
Here is my +1. Please remove unnecessary change at PgSQLTablespace.

hyunsik:
The change in PgSQLTablespace fixes a findbugs error.

PR #807, opened 2015-10-06
TAJO-1912: Selection from aliased schemaless tables throws an error

jihoonson:


jinossy:
+1 LGTM
Could you add to 0.11.0 ?

jihoonson:
Thanks for your review. I've committed to both master and 0.11.0.

PR #814, opened 2015-10-07
TAJO-1918: Text (de)serialization should be fixed for some types

eminency:
Fix getBytes to getTextBytes or use proper charset.


jinossy:
Thanks for your contribution
Could you add some tests?

jinossy:
+1
I verified 'mvn clean install’ and the test failure is not related to this patch.

PR #815, opened 2015-10-08
TAJO-1875: Resource leak after a query failure

jihoonson:
The main reason is that wrong resource release routine. 
When TaskManager receives a task start event ```e```, it checks that ```e``` is the first task start event of the current eb. If so, it create a context for the current eb. Here, if an error occurs while creating a eb context, the resource assigned for ```e``` must be released. 
However, in the current implementation, the resource release routine requires a task attempt id which will not exist with an eb start failure. As a result, resource manager loses the resource assigned for ```e```.


jinossy:
+1 Great!!
I verified 'mvn clean install’ and the test failure is not related to this patch.

PR #816, opened 2015-10-12
TAJO-1920: Calling 'Collection.toArray()' with zero-length array argument makes performance worse

dkhwangbo:
When passing in an array of too small size, the toArray() method has to construct a new array of the right size using reflection. This has significantly worse performance than passing in an array of at least the size of the collection itself.


hyunsik:
+1
The patch looks good to me.

PR #818, opened 2015-10-12
TAJO-1850: Using TUtil.newHash(Set/Map) should be replaced by Java's diamond operator

dkhwangbo:
See the title. We introduced Java 7. We don't need TUtil.newHash(Set|Map) utility methods anymore. We need to eliminate them.
This commit pass all the test with mvn clean install.


hyunsik:
The patch looks good to me. I manually tested it. It passed all unit tests. Could you remove TUtil::newHash method too?

dkhwangbo:
OK. I'll do that and commit again soon.

hyunsik:
I missed to check out TUtil class. Here is my +1.

PR #819, opened 2015-10-12
TAJO-1922: Allow date types with IN operator for partition pruning.

blrunner:
The following amendments were.

- Allow Date, Timestamp, Time with IN operator
- Update java.sql.Timestamp to TimestampDatum


tajo-core-tests/src/test/java/org/apache/tajo/engine/query/TestTablePartitions.java
```diff
@@ -1542,6 +1542,19 @@ public final void testDatePartitionColumn() throws Exception {
     assertEquals(expectedResult, resultSetToString(res));
     res.close();
 
+    // IN
+    res = executeString("SELECT * FROM " + tableName
+      + " WHERE key IN ( to_date('1994-02-02', 'YYYY-MM-DD'), to_date('1993-11-09', 'YYYY-MM-DD')) order by col1, " +
```
hyunsik: SQL standard representation for date would be simpler; e.g., ``DATE '1994-02-02'``.

tajo-core-tests/src/test/java/org/apache/tajo/engine/query/TestTablePartitions.java
```diff
@@ -1611,6 +1624,19 @@ public final void testTimestampPartitionColumn() throws Exception {
     assertEquals(expectedResult, resultSetToString(res));
     res.close();
 
+    // IN
+    res = executeString("SELECT * FROM " + tableName
+      + " WHERE key IN (to_timestamp('1994-02-02', 'YYYY-MM-DD'), to_timestamp('1993-11-09', 'YYYY-MM-DD')) " +
```
hyunsik: TIMESTAMP '1992-02-02 00:00:00' would be proper here.

tajo-core-tests/src/test/java/org/apache/tajo/engine/query/TestTablePartitions.java
```diff
@@ -1689,6 +1715,18 @@ public final void testTimePartitionColumn() throws Exception {
     assertEquals(expectedResult, resultSetToString(res));
     res.close();
 
+    // IN
+    res = executeString("SELECT * FROM " + tableName
+      + " WHERE key IN (cast('11:20:40' as time), cast('12:10:20' as time)) order by col1, col2, key desc");
```
hyunsik: SQL standard form like ``TIME '11:20:40'`` would be proper here.

tajo-plan/src/main/java/org/apache/tajo/plan/util/EvalNodeToExprConverter.java
```diff
@@ -143,34 +144,59 @@ protected EvalNode visitBinaryEval(Object o, Stack<EvalNode> stack, BinaryEval b
...
     TimeValue timeValue;
 
-    switch (evalNode.getValueType().getType()) {
+    switch (type) {
       case NULL_TYPE:
         value = new NullLiteral();
         break;
       case BOOLEAN:
-        value = new LiteralValue(evalNode.getValue().asChars(), LiteralValue.LiteralType.Boolean);
+        value = new LiteralValue(datum.asChars(), LiteralValue.LiteralType.Boolean);
```
hyunsik: The cases for ``BOOLEAN, ... TEXT`` can be simplified after getting a proper LiteralValue type.

tajo-plan/src/main/java/org/apache/tajo/plan/util/PartitionFilterAlgebraVisitor.java
```diff
@@ -150,8 +151,7 @@ public Expr visitTimestampLiteral(Object ctx, Stack<Expr> stack, TimestampLitera
       DateTimeUtil.toUTCTimezone(tm, tz);
 
       sb.append("?").append(" )");
-      Timestamp timestamp = new Timestamp(DateTimeUtil.julianTimeToJavaTime(DateTimeUtil.toJulianTimestamp(tm)));
-      parameters.add(new Pair(Type.TIMESTAMP, timestamp));
+      parameters.add(new Pair(Type.TIMESTAMP, new TimestampDatum(DateTimeUtil.toJulianTimestamp(tm))));
```
hyunsik: This code seem to change the timestamp to a julian time. Could you check if it is valid? We need to make the partitioned table compatible with Apache Hive. Hive does not appear to use a julian time.

hyunsik:
@blrunner I leaved some comments.

blrunner:
Thanks @hyunsik 

I reflected your comments. And currently, timestamp partition column cant't be compatible with Hive because of TAJO-1925. So, I commented unit test cases for timestamp temporarily.

tajo-plan/src/main/java/org/apache/tajo/plan/util/EvalNodeToExprConverter.java
```diff
@@ -178,65 +204,28 @@ protected EvalNode visitConst(Object o, ConstEval evalNode, Stack<EvalNode> stac
...
-          break;
-        case FLOAT4:
-        case FLOAT8:
-          value = new LiteralValue(datum.asChars(), LiteralValue.LiteralType.Unsigned_Float);
-          break;
-        default:
-          throw new RuntimeException("Unsupported type: " + datum.type().name());
-      }
-      values[i] = value;
+        throw new RuntimeException("Unsupported type: " + type.name());
```
jihoonson: It would be better to throw UnsupportedException.

blrunner: If we use UnsupportedException instead of RuntimeException, we should update more classes because ```EvalNodeToExprConverter``` has been implemented  by extending```SimpleEvalNodeVisitor```. How about TajoInternalError instead of UnsupportedException and RuntimeException?

jihoonson: TajoInternalError means an unrecoverable error is occured and at least the current running query must be stopped. It won't be what we intend here.
Instead, you can simply wrap UnsupportedException with RuntimeException.

blrunner: Thanks @jihoonson . I reflected the patch with your comments. :)

blrunner:
Separate unit test cases for IN operator to another method.

jihoonson:
The latest patch looks good to me. I'll finish my review after the CI test.

jihoonson:
+1 ship it.

PR #820, opened 2015-10-13
TAJO-1578: User configuration for pull server port is ignored.

dkhwangbo:
When TajoPullServerService is initialized, it overwrites user-configured pull server port with the default value.


jihoonson:
+1. This patch is straightforward.

PR #825, opened 2015-10-14
TAJO-1927: PartitionedTableRewriter should use a local variable to remember the volume of partitions.

blrunner:


blrunner:
I implemented ``FilteredPartitionInfo`` which includes partition directories and the total volume of partitions. And I added more unit test cases for verifying ``PartitionedTableRewriter``.

blrunner:
It would be resolved at TAJO-1952.

PR #827, opened 2015-10-15
TAJO-1891: Allow arbitrary partition path

blrunner:
This patch includes following changes:
- Add the list of PartitionDescProto to PartitionedTableScanNode
- Implement a method to build tuple using PartitionDescProto
- Restore alter table add partition statement
- Fix bugs about handling arbitrary partition path
- Add unit test cases


blrunner:
I'll reopen after resolving TAJO-1952.

PR #828, opened 2015-10-15
TAJO-1933: When a simple query executed on partitioned tables, the number of result rows is always the number of rows of the whole table

jihoonson:
I've changed to print unknown row number and unknown bytes if the total row number is larger than that printed in a page. Otherwise, the actual row number is printed.
IMHO, this is valid because the execution of a simple query is an iterable process, so it is difficult to get the total row number of the result.


jihoonson:
With my laptop, I verified this patch by running 'mvn install'.

jinossy:
+1 Looks good to me.

jihoonson:
Thanks for quick review!

PR #829, opened 2015-10-17
TAJO-1936: NPE when calling the query result REST api

charsyam:
Currently, Tajo Rest API return NPE when sessionId is null.
because of ConcurrentHashMap's put method throws when value is null.
So. I changed NPE to BadRequestExeception when session id is null.

Thanks.

### DIFF
--- a/tajo-core-tests/src/test/java/org/apache/tajo/ws/rs/resources/TestQueryResultResource.java
+++ b/tajo-core-tests/src/test/java/org/apache/tajo/ws/rs/resources/TestQueryResultResource.java
@@ -39,6 +39,7 @@ import org.junit.After;
 import org.junit.Before;
 import org.junit.Test;
 
+import javax.ws.rs.BadRequestException;
 import javax.ws.rs.client.Client;
 import javax.ws.rs.client.ClientBuilder;
 import javax.ws.rs.client.Entity;
@@ -58,7 +59,6 @@ import static org.apache.tajo.exception.ErrorUtil.isOk;
 import static org.junit.Assert.*;
 
 public class TestQueryResultResource extends QueryTestCaseBase {
-
   private URI restServiceURI;
   private URI sessionsURI;
   private URI queriesURI;
@@ -148,6 +148,17 @@ public class TestQueryResultResource extends QueryTestCaseBase {
     assertNotNull(response.getResultset());
     assertTrue(response.getResultset().getId() != 0);
     assertNotNull(response.getResultset().getLink());
+  }
+
+  @Test(expected = BadRequestException.class)
+  public void testGetQueryResultWithoutSessionId() throws Exception {
+    String sessionId = generateNewSessionAndGetId();
+    URI queryIdURI = sendNewQueryResquest(sessionId, "select * from lineitem");
+    URI queryResultURI = new URI(queryIdURI + "/result");
+
+    GetQueryResultDataResponse response = restClient.target(queryResultURI)
+            .request()
+            .get(new GenericType<>(GetQueryResultDataResponse.class));
   }
 
   @Test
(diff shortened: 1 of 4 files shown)


tajo-core-tests/src/test/java/org/apache/tajo/ws/rs/resources/TestQueryResultResource.java
```diff
@@ -58,6 +61,7 @@
 import static org.junit.Assert.*;
 
 public class TestQueryResultResource extends QueryTestCaseBase {
+  private static final Log LOG = LogFactory.getLog(TestQueryResultResource.class);
```
jihoonson: It seems not be used.

tajo-core-tests/src/test/java/org/apache/tajo/ws/rs/resources/TestQueryResultResource.java
```diff
@@ -151,6 +155,25 @@ public void testGetQueryResult() throws Exception {
...
+    URI queryIdURI = sendNewQueryResquest(sessionId, "select * from lineitem");
+    URI queryResultURI = new URI(queryIdURI + "/result");
+
+    boolean isBadRequest = false;
+    try {
+      GetQueryResultDataResponse response = restClient.target(queryResultURI)
+              .request()
+              .get(new GenericType<>(GetQueryResultDataResponse.class));
+    } catch (BadRequestException e) {
+      isBadRequest = true;
```
jihoonson: How about using expected exception of junit? It will make test codes simpler.
Please refer to test cases in TestCatalogAdminClientExceptions.

jihoonson:
Hi @charsyam, thanks for your work.
It looks nice to me. I left very trivial comments.

Anyway, as you already know, our REST api lacks methods to test failure cases. So, IMHO, it would be great if more methods to test various failure cases are added like ```testGetQueryResultWithoutSessionId()``` you already added in this patch.
What do you think?

charsyam:
@jihoonson 
Thanks for your review :)
I will apply your advice.

and I also agree with you. We need more failure testcases :)
I will add more failure testcases for REST-API in my space time :)
Thanks :)

jihoonson:
Thanks for understanding.
I will review as soon as possible after you update the patch.

charsyam:
@jihoonson maybe this tonight :) I will upload patch :) Thanks

charsyam:
@jihoonson I applied your review Thanks.
and Junit expected really make it simple :)

tajo-dist/src/main/conf/tajo-env.sh
```diff
@@ -22,10 +22,10 @@
 # remote nodes.
 
 # Hadoop home. Required
-# export HADOOP_HOME=
+export HADOOP_HOME=/Users/charsyam/hadoop
```
jihoonson: It seems your mistake.

tajo-core-tests/src/test/java/org/apache/tajo/ws/rs/resources/TestQueryResultResource.java
```diff
@@ -19,6 +19,8 @@
 package org.apache.tajo.ws.rs.resources;
 
 import org.apache.commons.codec.binary.Base64;
+import org.apache.commons.logging.Log;
+import org.apache.commons.logging.LogFactory;
```
jihoonson: Please remove unused imports.

jihoonson:
Thanks for quick update.
It seems that some unnecessary codes are included mistakenly.
Would you mind cleaning them up?

charsyam:
@jihoonson Thanks for your kind advice :)
I applied your review.

jihoonson:
+1 LGTM!

jihoonson:
@charsyam, would you please close this pr? 
Thanks.

PR #833, opened 2015-10-20
TAJO-1941: PermGen elimination in JDK 8

dkhwangbo:
The Permanent Generation (PermGen) space has completely been removed and is kind of replaced by a new space called Metaspace in JDK 8. So, I replace it to avoid WARN message and get performence improvement.

### DIFF
--- a/tajo-core-tests/pom.xml
+++ b/tajo-core-tests/pom.xml
@@ -73,7 +73,7 @@
           <systemProperties>
             <tajo.test.enabled>true</tajo.test.enabled>
           </systemProperties>
-          <argLine>-Xms512m -Xmx1024m -XX:MaxPermSize=152m -Dfile.encoding=UTF-8 -Dderby.storage.pageSize=1024 -Dderby.stream.error.file=/dev/null</argLine>
+          <argLine>-Xms512m -Xmx1024m -XX:MaxMetaspaceSize=152m -Dfile.encoding=UTF-8 -Dderby.storage.pageSize=1024 -Dderby.stream.error.file=/dev/null</argLine>
         </configuration>
       </plugin>
       <plugin>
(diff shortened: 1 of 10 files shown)


tajo-core-tests/pom.xml
```diff
@@ -375,7 +375,7 @@
               <forkCount>${maven.fork.count}</forkCount>
               <reuseForks>true</reuseForks>
               <trimStackTrace>false</trimStackTrace>
-              <argLine>-Xms512m -Xmx1024m -XX:MaxPermSize=152m -Dfile.encoding=UTF-8 -Dderby.storage.pageSize=1024 -Dderby.stream.error.file=/dev/null</argLine>
+              <argLine>-Xms512m -Xmx1024m -Dfile.encoding=UTF-8 -Dderby.storage.pageSize=1024 -Dderby.stream.error.file=/dev/null</argLine>
```
hyunsik: Any reason to omit ``XX:MaxMetaspaceSize`` here?

dkhwangbo:
With this option, Infinite GC POOL is occured during parallel test and cannot run next test. To avoid this phenomenon, I enter the parallel test without XX:MaxMetaspaceSize.
Currently, Whole test of TAJO is running without this option after version up of JAVA.

tajo-core-tests/pom.xml
dkhwangbo: With this option, Infinite GC POOL is occured during parallel test and cannot run next test. To avoid this phenomenon, I enter the parallel test without XX:MaxMetaspaceSize.
Currently, Whole test of TAJO is running without this option after version up of JAVA.

hyunsik:
@dkhwangbo Do you mean that ``XX:MaxMetaspaceSize`` causes hanging unit tests?

dkhwangbo:
@hyunsik in parallel test, yes.

hyunsik:
That may be size matter because JVM basically uses a default value for XX:MaxMetaspaceSize. 
Even though it may need more investigation for finding actual reason, It would be good to go ahead.
Here is my +1. Thank you for your work.

PR #834, opened 2015-10-20
TAJO-1643: SQLParse can't parse parentheses table clause

hyunsik:


jihoonson:
+1 looks good to me.
Anyway, as you already know, this is a workaround. 
I wonder that you have any plan to start to support explicit join order using parentheses in the ```from``` clause. I think that this feature will be helpful for users by allowing hint for join ordering as well as for us to estimate the efficiency of our join ordering algorithm.

hyunsik:
I totally agreed. We need to implement explicit join order too. I'll commit it shortly. Thank you !

PR #838, opened 2015-11-02
TAJO-1954: Fix memory leak in physical operator.

jinossy:


jihoonson:
+1 LGTM!

tajo-core/src/main/java/org/apache/tajo/engine/planner/physical/DistinctGroupbySortAggregationExec.java
```diff
@@ -172,6 +168,13 @@ public void close() throws IOException {
 
   @Override
   public void init() throws IOException {
+    super.init();
+
+    if (aggregateExecs != null) {
```
charsyam: @jinossy I have a question? Do we need check null check for aggregateExecs?

jinossy: @charsyam 
I refer to close() but It seems to the unnecessary checking. Are you working on this operator ?

charsyam: @jinossy I'm sorry.
I tested below code. and it causes NPE. 
```java
   List<String> lists = null;
    for(String a : lists) {
    }
```

I think null checking is needed. sorry to disturb you.

jinossy: PhysicalPlannerImpl.createSortAggregationDistinctGroupbyExec always set the array, so it can not occurs NPE. 
Next time, I will refactor these code.

Thanks.

PR #840, opened 2015-11-02
TAJO-1955: Add a feature to strip quotes from CSV file.

hyunsik:


jihoonson:
I'm concerned with the performance decrease. Do you have any idea of this?
If not, would you mind performing some brief evaluation?

hyunsik:
Could you share your thinking about the performance degradation point?

If you are concerned with the added if-condition branch, it does not matter in my opinion. Basically, it is very trivial than other code parts. In addition, ``hasQuoteChar`` is a final field, so JIT compiler will easily eliminate the added code if ``hasQuoteChar`` is false.

jihoonson:
Thanks for your comment. There was some my misunderstanding for this issue. It was, I understood that you try to allow delimiters between quote characters like this.
```
..., "ex, am, pl,e", ...
```

So, I concerned with that it will cost to figure out the start and end of quote characters. Sorry for my misunderstanding.

I've seen every code. After running a test manually, I'll give +1.

jihoonson:
You are so fast. Here is my +1.
This patch passed 'mvn install'.

PR #846, opened 2015-11-05
TAJO-1952: Implement PartitionFileFragment

blrunner:
This patch contains following modifications:

* Remove partition paths from PartitionedTableScanNode
* Implement PartitionedFileFragment
* Separate a method for pruning partition paths from PartitionedTableRewriter to PartitionedTableUtil
* Build a type by using partition name which contains partition keys and values

### DIFF
--- a/tajo-storage/tajo-storage-hdfs/src/main/java/org/apache/tajo/storage/parquet/TajoRecordConverter.java
+++ b/tajo-storage/tajo-storage-hdfs/src/main/java/org/apache/tajo/storage/parquet/TajoRecordConverter.java
@@ -20,6 +20,12 @@ package org.apache.tajo.storage.parquet;
 
 import com.google.protobuf.InvalidProtocolBufferException;
 import com.google.protobuf.Message;
+import org.apache.parquet.io.api.Binary;
+import org.apache.parquet.io.api.Converter;
+import org.apache.parquet.io.api.GroupConverter;
+import org.apache.parquet.io.api.PrimitiveConverter;
+import org.apache.parquet.schema.GroupType;
+import org.apache.parquet.schema.Type;
 import org.apache.tajo.catalog.Column;
 import org.apache.tajo.catalog.Schema;
 import org.apache.tajo.common.TajoDataTypes;
@@ -27,13 +33,7 @@ import org.apache.tajo.common.TajoDataTypes.DataType;
 import org.apache.tajo.datum.*;
 import org.apache.tajo.storage.Tuple;
 import org.apache.tajo.storage.VTuple;
-import org.apache.parquet.io.api.Binary;
-import org.apache.parquet.io.api.Converter;
-import org.apache.parquet.io.api.GroupConverter;
-import org.apache.parquet.io.api.PrimitiveConverter;
-import org.apache.parquet.schema.GroupType;
-import org.apache.parquet.schema.Type;
-import org.apache.tajo.util.datetime.DateTimeUtil;
+import org.apache.tajo.util.datetime.DateTimeConstants;
 
 import java.nio.ByteBuffer;
 import java.util.Arrays;
@@ -337,7 +337,7 @@ public class TajoRecordConverter extends GroupConverter {
     @Override
     final public void addInt(int value) {
       // Parquet DATE type is based on Unix Epoch(Jan 1, 1970).
-      parent.add(DatumFactory.createDate(value + DateTimeUtil.DAYS_FROM_JULIAN_TO_EPOCH));
+      parent.add(DatumFactory.createDate(value + DateTimeConstants.UNIX_EPOCH_JDATE));
     }
   }
 
--- a/tajo-storage/tajo-storage-hdfs/src/main/java/org/apache/tajo/storage/parquet/TajoWriteSupport.java
+++ b/tajo-storage/tajo-storage-hdfs/src/main/java/org/apache/tajo/storage/parquet/TajoWriteSupport.java
@@ -19,18 +19,18 @@
 package org.apache.tajo.storage.parquet;
 
 import org.apache.hadoop.conf.Configuration;
-import org.apache.tajo.catalog.Column;
-import org.apache.tajo.catalog.Schema;
-import org.apache.tajo.common.TajoDataTypes;
-import org.apache.tajo.exception.ValueTooLongForTypeCharactersException;
-import org.apache.tajo.storage.Tuple;
 import org.apache.parquet.hadoop.api.WriteSupport;
 import org.apache.parquet.io.api.Binary;
 import org.apache.parquet.io.api.RecordConsumer;
 import org.apache.parquet.schema.GroupType;
 import org.apache.parquet.schema.MessageType;
 import org.apache.parquet.schema.Type;
-import org.apache.tajo.util.datetime.DateTimeUtil;
+import org.apache.tajo.catalog.Column;
+import org.apache.tajo.catalog.Schema;
+import org.apache.tajo.common.TajoDataTypes;
+import org.apache.tajo.exception.ValueTooLongForTypeCharactersException;
+import org.apache.tajo.storage.Tuple;
+import org.apache.tajo.util.datetime.DateTimeConstants;
 
 import java.util.HashMap;
 import java.util.List;
@@ -143,7 +143,7 @@ public class TajoWriteSupport extends WriteSupport<Tuple> {
         break;
       case DATE:
         // Parquet DATE type is based on Unix Epoch(Jan 1, 1970).
-        recordConsumer.addInteger(tuple.getInt4(index) - DateTimeUtil.DAYS_FROM_JULIAN_TO_EPOCH);
+        recordConsumer.addInteger(tuple.getInt4(index) - DateTimeConstants.UNIX_EPOCH_JDATE);
         break;
       case PROTOBUF:
       case BLOB:


tajo-storage/tajo-storage-hdfs/src/main/java/org/apache/tajo/storage/FileTablespace.java
```diff
@@ -477,12 +503,17 @@ public long getMinSplitSize() {
...
+    return getSplits(tableName, meta, schema, null, inputs);
+  }
+
   /**
    * Generate the list of files and make them into FileSplits.
    *
    * @throws IOException
    */
-  public List<Fragment> getSplits(String tableName, TableMeta meta, Schema schema, Path... inputs)
+  public List<Fragment> getSplits(String tableName, TableMeta meta, Schema schema, String[] partitions, Path... inputs)
```
hyunsik: I'm going to give a suggestion. We need to keep each method for one purpose if possible. That approach would keep logic simpler.

hyunsik:
Could you rebase it against the latest revision? I'd like to try the patch on my machine.

blrunner:
@hyunsik 

Thank you for your review. I've just rebased it against the latest version.

tajo-storage/tajo-storage-hdfs/src/main/proto/StorageFragmentProtos.proto
```diff
@@ -32,3 +32,14 @@ message FileFragmentProto {
...
+
+message PartitionedFileFragmentProto {
+  required string id = 1;
+  required string path = 2;
+  required int64 start_offset = 3;
+  required int64 length = 4;
+  repeated string hosts = 5;
+  repeated int32 disk_ids = 6;
+  // Partition Name: country=KOREA/city=SEOUL
+  required string partitionName = 7;
```
hyunsik: In my opionion, partitionKeys would be more proper because the attribute includes concatenated partition keys.

tajo-storage/tajo-storage-hdfs/src/main/java/org/apache/tajo/storage/FileTablespace.java
```diff
@@ -420,9 +421,29 @@ protected FileFragment makeSplit(String fragmentId, Path file, BlockLocation blo
     return new FileFragment(fragmentId, file, blockLocation);
   }
 
+  protected PartitionedFileFragment makePartitionSplit(String fragmentId, Path file, long start, long length,
```
hyunsik: It seems to be an unused method.

tajo-storage/tajo-storage-hdfs/src/main/java/org/apache/tajo/storage/fragment/PartitionedFileFragment.java
```diff
@@ -0,0 +1,254 @@
...
+import static org.apache.tajo.catalog.proto.CatalogProtos.FragmentProto;
+
+public class PartitionedFileFragment implements Fragment, Comparable<PartitionedFileFragment>, Cloneable {
+  @Expose private String tableName; // required
+  @Expose private Path uri; // required
+  @Expose public Long startOffset; // required
+  @Expose public Long length; // required
+
+  private String[] hosts; // Datanode hostnames
+  @Expose private int[] diskIds;
```
hyunsik: Its getter and setter seem to be not used. Are there any reason to keep this information here?

blrunner:
@hyunsik 

Thank your for your review. I've just reflected your comments.

blrunner:
I updated the patch as following:

* Renamed PartitionedFileFragment to PartitionFileFragment
* Recovered codes for pruning partition in PartitionedTableRewrtier
* Recovered partition paths of PartitionedTableScanNode and added partition keys to PartitionedTableScanNode
* Implemented PartitionContent which contains partition paths, partition keys, partition volume

For the reference, I've tried to maintain existing codes as far as possible.

tajo-plan/src/main/java/org/apache/tajo/plan/partition/PartitionContent.java
```diff
@@ -0,0 +1,69 @@
...
+
+public class PartitionContent {
+  private Path[] partitionPaths;
+  private String[] partitionKeys;
+  private long totalVolume;
+
+  public PartitionContent() {
+  }
+
+  public PartitionContent(Path[] partitionPaths) {
```
hyunsik: unused ctor

tajo-plan/src/main/java/org/apache/tajo/plan/partition/PartitionContent.java
```diff
@@ -0,0 +1,69 @@
...
+    this.partitionPaths = partitionPaths;
+  }
+
+  public PartitionContent(Path[] partitionPaths, long totalVolume) {
```
hyunsik: unused constructor

tajo-plan/src/main/java/org/apache/tajo/plan/rewrite/rules/PartitionedTableRewriter.java
```diff
@@ -42,13 +45,13 @@
 import org.apache.tajo.storage.Tuple;
 import org.apache.tajo.storage.VTuple;
 import org.apache.tajo.util.StringUtils;
+import org.apache.tajo.util.TUtil;
```
hyunsik: unused

tajo-plan/src/main/java/org/apache/tajo/plan/rewrite/rules/PartitionedTableRewriter.java
```diff
@@ -216,7 +228,22 @@ public String toString() {
...
+    partitionKeys = new String[filteredPaths.length];
+    for (int i = 0; i < partitionKeys.length; i++) {
+      Path path = filteredPaths[i];
+      startIdx = path.toString().indexOf(getColumnPartitionPathPrefix(partitionColumns));
+      partitionKeys[i] = path.toString().substring(startIdx);
+      summary = fs.getContentSummary(path);
+      totalVolume += summary.getLength();
+    }
+
+    partitionContent.setPartitionPaths(filteredPaths);
```
hyunsik: It can be refactored to use just the constructor instead of setters.

blrunner:
@hyunsik 

Updated the patch as following:

* Remove unused packages
* Remove PartitionedTableScanNode of PlanProto
* Remove unnecessary PartitionedTableScanNode usages of Repartitioner and PhysicalPlannerImpl

blrunner:
Added unit test cases for PartitionedTableRewriter

tajo-storage/tajo-storage-hdfs/src/main/java/org/apache/tajo/storage/FileTablespace.java
```diff
@@ -550,6 +551,181 @@ public long getMinSplitSize() {
...
+        if (hostsBlockMap.containsKey(host)) {
+          hostsBlockMap.put(host, hostsBlockMap.get(host) + 1);
+        } else {
+          hostsBlockMap.put(host, 1);
+        }
+      }
+    }
+
+    List<Map.Entry<String, Integer>> entries = new ArrayList<>(hostsBlockMap.entrySet());
+    Collections.sort(entries, new Comparator<Map.Entry<String, Integer>>() {
```
hyunsik: It can be simplified with lambda.

tajo-storage/tajo-storage-hdfs/src/main/java/org/apache/tajo/storage/FileTablespace.java
```diff
@@ -550,6 +551,181 @@ public long getMinSplitSize() {
     return splits;
   }
 
+  ////////////////////////////////////////////////////////////////////////////////
+  // The below code is for splitting partitioned table.
+  ////////////////////////////////////////////////////////////////////////////////
+
+  public PartitionFileFragment[] partitionSplit(String tableName, Path tablePath, String partitionKeys)
```
hyunsik: It seems to be not used.

tajo-storage/tajo-storage-hdfs/src/main/java/org/apache/tajo/storage/FileTablespace.java
```diff
@@ -550,6 +551,181 @@ public long getMinSplitSize() {
...
+
+  protected PartitionFileFragment makePartitionSplit(String fragmentId, Path file, BlockLocation blockLocation
+    , String partitionKeys) throws IOException {
+    return new PartitionFileFragment(fragmentId, file, blockLocation, partitionKeys);
+  }
+
+  protected Fragment makeNonPartitionSplit(String fragmentId, Path file, long start, long length,
+                                           BlockLocation[] blkLocations, String partitionKeys) throws IOException {
+
+    Map<String, Integer> hostsBlockMap = new HashMap<>();
```
hyunsik: This code fragment is duplicated to that of ``makeNonSplit``. Could you refactor it?

tajo-storage/tajo-storage-hdfs/src/main/java/org/apache/tajo/storage/FileTablespace.java
```diff
@@ -550,6 +551,181 @@ public long getMinSplitSize() {
...
+                  blkLocations[blkIndex].getHosts(), partitionKeys[i]));
+              }
+            } else { // Non splittable
+              splits.add(makeNonPartitionSplit(tableName, path, 0, length, blkLocations, partitionKeys[i]));
+            }
+          }
+        }
+      }
+      if(LOG.isDebugEnabled()){
+        LOG.debug("# of splits per partition: " + (splits.size() - previousSplitSize));
```
hyunsik: The log message is somewhat different from the actual meaning.

``PreviousSplitSize`` is the number of files in partition directories, and split is the number of total fragments. For example, there are two files in all partitions, and each file makes 5 fragments. Then, the total number of fragments is 10. In that case, 10 - 2 is jut 8. 8 is not the number of splits per partition.  The actual answer should be 5. In addition, we can say X per Y when X is the same across all Ys. It is hard to get answer because each partition file can be vary in terms of size. So, average is more proper here, I think.

tajo-storage/tajo-storage-hdfs/src/main/java/org/apache/tajo/storage/FileTablespace.java
```diff
@@ -550,6 +551,181 @@ public long getMinSplitSize() {
...
+    int i = 0;
+    for (Path p : inputs) {
+      ArrayList<FileStatus> files = Lists.newArrayList();
+      if (fs.isFile(p)) {
+        files.addAll(Lists.newArrayList(fs.getFileStatus(p)));
+      } else {
+        files.addAll(listStatus(p));
+      }
+
+      int previousSplitSize = splits.size();
```
hyunsik: It would be better if the ```final``` marked for ``previousSplitSize``.

tajo-plan/src/main/java/org/apache/tajo/plan/partition/PartitionContent.java
```diff
@@ -0,0 +1,45 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.tajo.plan.partition;
+
+import org.apache.hadoop.fs.Path;
+
+public class PartitionContent {
+  private Path[] partitionPaths;
```
hyunsik: It would be better if they all are marked by final.

blrunner:
@hyunsik 

Thank you for your detailed review. I reflected your comments.

blrunner:
@hyunsik 

I changed ```PartitionFileFragment``` by extending ```FileFragment``` because there are some codes for only ```FileFragment```. For example, you can see cast codes as following:
```
  public FileScanner(Configuration conf, final Schema schema, final TableMeta meta, final Fragment fragment) {
    this.conf = conf;
    this.meta = meta;
    this.schema = schema;
    this.fragment = (FileFragment)fragment;
    this.tableStats = new TableStats();
    this.columnNum = this.schema.size();
  }
```

blrunner:
Evaluated patch testing before and after ``PartitionFileFragment`` implementation as following:

* Dataset : TPCH-100G
* Tajo Cluster : 1 Master, 6 Workers
* Queries: Q1, Q3, Q5, Q6, Q7, Q8, Q9, Q10
* Tajo version for before  ``PartitionFileFragment`` implementation 
 -  0.11.1-SNAPSHOT
* Tajo version for after  ``PartitionFileFragment`` implementation:
 - 0.12.0-SNAPSHOT (partitions exist on catalog)
 - 0.12.0-SNAPSHOT (partitions doesn't exist on catalog)

The results were same as following:

* The number and order of execution blocks
* The number of stags in a each execution block
* The number of rows in a result
* All tuples in a result (excluded a few floating point value)

blrunner:
The Travis CI Build seems like a fail by another reason. Here is the pre-commit report by lasted patch as follows. https://builds.apache.org/job/PreCommit-TAJO-Build/900//console

blrunner:
@hyunsik 

Is there a problem that needs to be fixed?

blrunner:
@hyunsik 

I updated this PR as following:

* Remove the list of partition paths in PartitionedTableScanNode. It just contains the path of table root.
* PartitionedTableRewriter just rewrite ScanNode to PartitionedTableScanNode in BaseLogicalPlanRewriteRuleProvider.
* Actual partition pruning will be executed in Repartitioner by PartitionedTableRewriter.

PR #847, opened 2015-11-05
TAJO-1964: TestKillQuery::testKillQueryFromInitState causes NPE.

jinossy:


jihoonson:
+1. I've manually tested with 'mvn install'.

PR #844, opened 2015-11-05
TAJO-1963: Add more configuration descriptions to document

jihoonson:
I also fixed a wrong configuration name.

PR #848, opened 2015-11-05
TAJO-1962: Add description for session variables

jihoonson:


tajo-docs/src/main/sphinx/tsql/variables.rst
```diff
@@ -28,35 +30,456 @@ Each client connection to TajoMaster creates a unique session, and the client an
...
+  * Default value: false
+  * Example
+
+.. code-block:: sh
+
+  \set DEBUG_ENABLED false
+
+.. describe:: FETCH_ROWNUM
+
+Sets the number of rows at a time from Master.
```
hyunsik: How about them?
 * the number of rows in a ResultSet for each fetch
 * the number of rows to be fetched from Master

tajo-common/src/main/java/org/apache/tajo/SessionVars.java
```diff
@@ -131,7 +131,7 @@
   NULL_CHAR(ConfVars.$TEXT_NULL, "null char of text file output", DEFAULT),
   CODEGEN(ConfVars.$CODEGEN, "Runtime code generation enabled (experiment)", DEFAULT),
   AGG_HASH_TABLE_SIZE(ConfVars.$AGG_HASH_TABLE_SIZE, "Aggregation hash table size", DEFAULT),
-  SORT_HASH_TABLE_SIZE(ConfVars.$SORT_HASH_TABLE_SIZE, "Sort hash table size", DEFAULT),
+  SORT_LIST_SIZE(ConfVars.$SORT_LIST_SIZE, "Sort hash table size", DEFAULT),
```
eminency: Does the description need to be changed?

tajo-docs/src/main/sphinx/tsql/variables.rst
```diff
@@ -28,35 +30,456 @@ Each client connection to TajoMaster creates a unique session, and the client an
...
+  * Default value: org.apache.tajo.cli.tsql.DefaultTajoCliOutputFormatter
+  * Example
+
+.. code-block:: sh
+
+  \set CLI_FORMATTER_CLASS org.apache.tajo.cli.tsql.DefaultTajoCliOutputFormatter
+
+.. describe:: ON_ERROR_STOP
+
+tsql will exist if an error occurs.
```
eminency: exist -> exit

tajo-docs/src/main/sphinx/tsql/variables.rst
```diff
@@ -28,35 +30,456 @@ Each client connection to TajoMaster creates a unique session, and the client an
...
+  * Default value: false
+  * Example
+
+.. code-block:: sh
+
+  \set ON_ERROR_STOP false
+
+.. describe:: NULL_CHAR
+
+Null char of text file output.
```
hyunsik: It would be better to add 'used when the table property ``text.null`` is not specified' after the this description.

tajo-docs/src/main/sphinx/tsql/variables.rst
```diff
@@ -28,35 +30,456 @@ Each client connection to TajoMaster creates a unique session, and the client an
...
+  * Property value: Integer
+  * Unit: MB
+  * Default value: 64
+  * Example
+
+.. code-block:: sh
+
+  \set JOIN_TASK_INPUT_SIZE 64
+
+.. describe:: JOIN_PER_SHUFFLE_SIZE
```
eminency: I think this name looks weird. It lets me think it is like 'number of join task for each shuffle (size)'.
But I have no idea for good alternative.

tajo-docs/src/main/sphinx/tsql/variables.rst
```diff
@@ -28,35 +30,456 @@ Each client connection to TajoMaster creates a unique session, and the client an
...
+  * Property value: Integer
+  * Unit: MB
+  * Default value: 128
+  * Example
+
+.. code-block:: sh
+
+  \set JOIN_PER_SHUFFLE_SIZE 128
+
+.. describe:: HASH_JOIN_SIZE_LIMIT
```
eminency: Corresponding config property name uses 'threshold', but this one is 'limit'.
I think it needs some kind of consistency.

tajo-docs/src/main/sphinx/tsql/variables.rst
```diff
@@ -28,35 +30,456 @@ Each client connection to TajoMaster creates a unique session, and the client an
...
+  \set HASH_GROUPBY_SIZE_LIMIT 64
+
+.. warning::
+  This value is the size of the input stored on file systems. So, when the input data is loaded into JVM heap,
+  its actual size is usually much larger than the configured value, which means that too large threshold can cause unexpected OutOfMemory errors.
+  This value should be tuned carefully.
+
+.. describe:: AGG_HASH_TABLE_SIZE
+
+The initial size of list for in-memory sort.
```
eminency: Description is for 'list', but name is 'hash table'.

tajo-docs/src/main/sphinx/tsql/variables.rst
hyunsik: Actually, join is processed in the second stage. The first stage is for shuffle. This parameters only determines the input size of the second stages. So, this name is proper in my opinion.

In addition, this parameter has a similar context to map (reduce) task size. So, it would be familiar to those who have Hadoop experiences.

tajo-docs/src/main/sphinx/tsql/variables.rst
hyunsik: I agreed. ``Threshold`` may be too long for human to put on cli. Actually, there mainly are two kinds of suffixes: SIZE and LIMIT. SIZE is literally used to represent size matter. ``LIMIT`` has been used to represent some number to determine whether doing or not doing according to the number. It would be better to make them consistent. 

But, it needs to be addressed in an another jira.

tajo-docs/src/main/sphinx/tsql/variables.rst
```diff
@@ -28,35 +30,456 @@ Each client connection to TajoMaster creates a unique session, and the client an
...
+  * Default value: '\\N'
+  * Example
+
+.. code-block:: sh
+
+  \set NULL_CHAR '\\N'
+
+.. describe:: DEBUG_ENABLED
+
+Debug mode enabled.
```
eminency: 'To enable debug mode' might be more appropriate.

eminency:
I leave some comments.
Actually many items are related and duplicated with 'tajo-site.xml' doc.

I think you'd better consider to omit explanations for those items and to put corresponding references to tajo-site.xml.

jihoonson:
Nice idea. I'll add references to corresponding configurations in tajo-site.xml after #844 is committed.

tajo-docs/src/main/sphinx/tsql/variables.rst
jihoonson: How about ```The number of rows to be fetched from Master at once```?

tajo-docs/src/main/sphinx/tsql/variables.rst
jihoonson: My mistake. Thanks.

tajo-docs/src/main/sphinx/tsql/variables.rst
jihoonson: Thanks for the good comment.

tajo-docs/src/main/sphinx/tsql/variables.rst
jihoonson: I opened a new jira for this issue. https://issues.apache.org/jira/browse/TAJO-1968

tajo-docs/src/main/sphinx/tsql/variables.rst
jihoonson: My mistake. Thanks.

tajo-docs/src/main/sphinx/tsql/variables.rst
jihoonson: Thanks for comment. I fixed.

tajo-docs/src/main/sphinx/tsql/variables.rst
jihoonson: Yes, this name is valid. Maybe, there are more intuitive names, but I also don't have any idea now. We can change when a good idea comes to mind.

eminency: My target was 'JOIN_PER_SHUFFLE_SIZE', not 'JOIN_TASK_INPUT_SIZE'.

hyunsik:
@jihoonson Is it ready to be committed? If so, I'll finish the review soon.

jihoonson:
Thanks @hyunsik, but not yet. I'll ping after update soon.

jihoonson:
@eminency, @hyunsik thanks for your review. I've updated my patch.
Please review it.

eminency:
I think it's very good.

+1

hyunsik:
+1 ship it!

jihoonson:
@eminency, @hyunsik thank you for your review. I've committed my patch.

PR #851, opened 2015-11-05
TAJO-1896: Fix the travis test failure.

jinossy:


jihoonson:
Thanks for your great work. This is the really necessary and urgent issue to be resolved as soon as possible. 
Your patch looks good to me overall. I have only one question. Would you share your intention why you removed the restriction on ```MaxMetaspaceSize```?

jinossy:
Garbage collection of the dead classes and classloaders is triggered once the class metadata usage reaches the “MaxMetaspaceSize”
So we don't need limitation. 

https://dzone.com/articles/java-8-permgen-metaspace

jihoonson:
Thanks for your reply. I also found the following phrase from the link.
```
By default class metadata allocation is limited by the amount of available native memory.
```
So, we can expect that our test can use every available native memory with GC. If our test is failed due to lack of available memory, it means that our test exhaust too much memory, I think.
Here is my +1. Appreciate for your work.

PR #855, opened 2015-11-09
TAJO-1971: Replace 'for' loop with 'foreach'

dkhwangbo:
Reports for loops which iterate over collections or arrays, and can be replaced with the foreach iteration syntax, available in Java 5 and newer, to get the improvement of readability and so on.


eminency:
Could you merge latest master branch please?

dkhwangbo:
I'll re-make same PR with latest master. It's more easy than resolving conflict.

PR #856, opened 2015-11-09
TAJO-1970: Change the null first syntax

jihoonson:


blrunner:
+1 

LGTM! I found that the patch ran successfully on my local cluster. :)

jihoonson:
@blrunner, thank you for quick review!

PR #858, opened 2015-11-09
TAJO-1973: Replace 'while' loop with 'foreach'

dkhwangbo:
'While' loops which iterate over collections can be replaced with the 'foreach' iteration syntax, which is available in Java 5 and newer.

### DIFF
--- a/tajo-catalog/tajo-catalog-server/src/test/java/org/apache/tajo/catalog/TestCatalog.java
+++ b/tajo-catalog/tajo-catalog-server/src/test/java/org/apache/tajo/catalog/TestCatalog.java
@@ -274,10 +274,8 @@ public class TestCatalog {
     Map<String, List<String>> createdTablesMap = createBaseDatabaseAndTables();
 
     // Each time we drop one database, check all databases and their tables.
-    Iterator<String> it = new ArrayList<>(createdTablesMap.keySet()).iterator();
-    while(it.hasNext()) {
+    for (String databaseName : new ArrayList<>(createdTablesMap.keySet())) {
       // drop one database
-      String databaseName = it.next();
       assertTrue(catalog.existDatabase(databaseName));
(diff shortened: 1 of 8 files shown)


eminency:
+1

tajo-catalog/tajo-catalog-server/src/test/java/org/apache/tajo/catalog/TestCatalog.java
```diff
@@ -274,10 +274,8 @@ public void testDropDatabaseWithAllTables() throws Exception {
     Map<String, List<String>> createdTablesMap = createBaseDatabaseAndTables();
 
     // Each time we drop one database, check all databases and their tables.
-    Iterator<String> it = new ArrayList<>(createdTablesMap.keySet()).iterator();
-    while(it.hasNext()) {
+    for (String databaseName : new ArrayList<>(createdTablesMap.keySet())) {
```
jihoonson: Creation of this array list looks redundant.

dkhwangbo: Thanks for your reply. 
As my understanding, That line is written to convert Set to List because Set has no ordering. I have no idea with another way to get same result. Can I get some more details about your comment?

dkhwangbo: Oh, I found that my above comment has no relation about this issue.
Creation of this array is needed because of 284 line in same file.

jihoonson: Right. My misunderstanding.

jihoonson:
+1 LGTM

PR #859, opened 2015-11-12
TAJO-1975: Gathering fine-grained column statistics for range shuffle

jihoonson:


tajo-storage/tajo-storage-hbase/src/main/java/org/apache/tajo/storage/hbase/AbstractHBaseAppender.java
```diff
@@ -53,7 +51,8 @@
 
   protected ColumnMapping columnMapping;
   protected TableStatistics stats;
-  protected boolean enabledStats;
+  protected boolean tableStatsEnabled;
+  protected BitSet columnStatsEnabled;
```
jinossy: BitSet needs computation cost. How about change to the array?

jihoonson: Thanks for your comment. I've changed it.

jinossy:
+1 Looks great to me

PR #861, opened 2015-11-16
TAJO-1978: Printout message before terminating TSQL

dkhwangbo:
In ExitCommand::invoke, the code line is written that printout "bye!" before terminating. But, There is no sout.flush() before exit(0) so that message is not shown.

### DIFF
--- a/tajo-cli/src/main/java/org/apache/tajo/cli/tsql/commands/ExitCommand.java
+++ b/tajo-cli/src/main/java/org/apache/tajo/cli/tsql/commands/ExitCommand.java
@@ -37,7 +37,7 @@ public class ExitCommand extends TajoShellCommand {
   @Override
   public void invoke(String[] cmd) throws Exception {
     context.getOutput().println("bye!");
-    context.getOutput().flush();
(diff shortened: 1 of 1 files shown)


tajo-cli/src/main/java/org/apache/tajo/cli/tsql/commands/ExitCommand.java
```diff
@@ -37,6 +37,7 @@ public String getCommand() {
   @Override
   public void invoke(String[] cmd) throws Exception {
     context.getOutput().println("bye!");
+    context.getOutput().flush();
```
blrunner: ```flush()``` flushes content of buffer to destination. And you can write something again into the stream. ```close()``` flushes content to destination and closes the stream. After execute the command for terminating TSql, we will never write anything anymore. Thus, we need to use ```close()``` instead of  ```flush()```.

dkhwangbo: @blrunner  Thanks for your reply. I apply your comment.

PR #862, opened 2015-11-16
TAJO-1979: Usage of some TajoShellCommand is omitted

dkhwangbo:
Write omitted usage of TajoShellCommand.

### DIFF
--- a/tajo-cli/src/main/java/org/apache/tajo/cli/tsql/commands/ConnectDatabaseCommand.java
+++ b/tajo-cli/src/main/java/org/apache/tajo/cli/tsql/commands/ConnectDatabaseCommand.java
@@ -72,7 +72,7 @@ public class ConnectDatabaseCommand extends TajoShellCommand {
 
   @Override
   public String getUsage() {
-    return "";
+    return "[database_name]";
   }
 
   @Override
--- a/tajo-cli/src/main/java/org/apache/tajo/cli/tsql/commands/SetCommand.java
+++ b/tajo-cli/src/main/java/org/apache/tajo/cli/tsql/commands/SetCommand.java
@@ -87,7 +87,7 @@ public class SetCommand extends TajoShellCommand {
 
   @Override
   public String getUsage() {
-    return "";
+    return "[[NAME] VALUE]";
   }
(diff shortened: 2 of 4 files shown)


tajo-cli/src/main/java/org/apache/tajo/cli/tsql/commands/ConnectDatabaseCommand.java
```diff
@@ -63,7 +63,7 @@ public void invoke(String[] cmd) throws Exception {
 
   @Override
   public String getUsage() {
-    return "";
+    return "[database_name]";
```
blrunner: How about use ```DBNAME``` instead of ```database_name```?  TSQL already uses ```DBNAME``` to print help descriptions with ```\help``` parameter.

blrunner:
If you add some unit test cases for this patch, it would be more better.

tajo-cli/src/main/java/org/apache/tajo/cli/tsql/commands/ConnectDatabaseCommand.java
dkhwangbo: I write ```database_name``` refer to ```DescTableCommand::getUsage()```. In ```DescTableCommand::getUsage()```, return ```[table_name]``` in spite of ```[TBNAME]``` with ```\help``` . You mean that editting other ```getUsage()``` is needed?

dkhwangbo:
@blrunner Thanks for your review. I found another bug that ```getUsage()``` has no usage in whole TAJO, so I have no idea about making unit test cases. ```TajoShellCommand::printHelp()``` needs ```getUsage()``` but It's omitted. After resolve this issue in another ticket, I add unit test cases relate of this PR using shell command like ```TestTajoCli::testTimeZoneSessionVars1()```.

blrunner:
I know you've finished the issue which is involved with this PR. Could you add an unit test case?

dkhwangbo:
@blrunner Oh, thanks for your remind. I'll add unit soon.

blrunner:
+1

LGTM!

PR #865, opened 2015-11-18
TAJO-1981: Invalid nulls sort order in VectorizedSorter

jihoonson:
I'm sorry for a large patch. Here are the highlights of this patch.
* Fix the problem of invalid sort order of nulls in VectorizedSorter
* Change the name of ```nullFirst``` variable to ```nullsFirst```
* Change the string representation of ```SortSpec``` to print out the nulls order

Most changes of this patch is due to the changed string representation of ```SortSpec```, and there are only very minor changes.
Please review this patch.


jinossy:
+1 LGTM!!

PR #867, opened 2015-11-19
TAJO-1862: TSQL gives user wrong URL of documentation

dkhwangbo:
When user build master branch and run TSQL, TSQL gives user wrong url like below.

http://tajo.apache.org/docs/0.12.0/tsql.html

To help beginner or intermediate user, It should be fix.


hyunsik:
+1
The patch looks good to me.

PR #869, opened 2015-11-19
TAJO-1983: Improve memory usage of ExternalSortExec.

jinossy:
I've add UnSafeTupleList. it uses off-heap memory. 
A last chunk are not stored in local file and the chunks are merged when the next method is called.
In addition, I’ve change the insignificant log level to debug.


jinossy:
This PR is ready to review.
Thanks

tajo-pullserver/src/main/java/org/apache/tajo/pullserver/TajoPullServerService.java
```diff
@@ -403,12 +400,20 @@ public void decrementRemainFiles(FileRegion filePart, long fileStartTime) {
         minTime = fileSendTime;
       }
 
+      if (fileSendTime > 20 * 1000) {
+        LOG.warn("PullServer send too long time: filePos:" + filePart.position()
```
hyunsik: I know you just moved this code. But, the original logging message needs to be corrected. Could you update the log message?

tajo-core/src/main/java/org/apache/tajo/engine/planner/physical/ExternalSortExec.java
```diff
@@ -103,7 +117,7 @@
   /** the final result */
```
hyunsik: The member variable ``memoryResident`` seems to be not necessary anymore.

hyunsik:
The patch looks great to me. Here is my +1. I leaved some trivial comments.You can commit the patch if you reflect my comments.

jinossy:
I've update the patch that reflects your comments and I will commit soon
Thanks for your review!

PR #874, opened 2015-11-24
TAJO-1987: Add stream API to Projectable and refactor its usage codes.

hyunsik:
This patch is a trivial patch to add a stream API to Projectable and refactor its usage codes.


blrunner:
I'm reviewing this patch.

tajo-plan/src/main/java/org/apache/tajo/plan/logical/ScanNode.java
```diff
@@ -32,7 +32,10 @@
 import org.apache.tajo.util.TUtil;
 
 import java.util.ArrayList;
+import java.util.Collections;
 import java.util.List;
+import java.util.stream.Stream;
+import java.util.stream.StreamSupport;
```
blrunner: It seems to be an unused package.

tajo-plan/src/main/java/org/apache/tajo/plan/verifier/LogicalPlanVerifier.java
```diff
@@ -90,9 +90,7 @@ public LogicalNode visitProjection(Context state, LogicalPlan plan, LogicalPlan.
                                      ProjectionNode node, Stack<LogicalNode> stack) throws TajoException {
     super.visitProjection(state, plan, block, node, stack);
 
-    for (Target target : node.getTargets()) {
```
blrunner: Could you remove ```Target``` in package import codes?

blrunner:
The patch looks good to me. I leaved a few trivial comments.

tajo-plan/src/main/java/org/apache/tajo/plan/logical/ProjectionNode.java
```diff
@@ -64,8 +65,13 @@ public void setTargets(List<Target> targets) {
   public List<Target> getTargets() {
     return this.targets;
   }
-	
-	public void setChild(LogicalNode subNode) {
+
+  @Override
+  public Stream<Target> targets() {
```
jihoonson: Would you share why you keep the old method ```getTargets()```? For the compatibility with 1.7?

hyunsik: This is because we still need List allowing access elements by index in some codes.

hyunsik:
I reflected the comments and rebased.

jihoonson:
This patch looks good to me. Would you rebase against master?

hyunsik:
Close this issue because most of improvements are already included in TAJO-2108.

PR #875, opened 2015-11-24
TAJO-1988: Remove some duplicated codes in toString() of Projectable.

hyunsik:
This is an trivial patch to remove some duplicated codes.


jinossy:
+1

dkhwangbo:
+1

PR #876, opened 2015-11-24
TAJO-1944: Support text resultset for REST

charsyam:
This is just for review. and I just add csv only now :)

I worried about 3 points. 

first of all, memory usages for other type.
I think most users want to download all data as csv one time.

second, add new method for NonForwardQueryResultScanner to get tuple list.

third, removing base64 checking for consistency with other type(csv, binary)


hyunsik:
The patch looks good to me. Soon, I'll finish the review.

jihoonson:
I'm reviewing too.

tajo-core/src/main/java/org/apache/tajo/ws/rs/resources/outputs/AbstractStreamingOutput.java
```diff
@@ -0,0 +1,46 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.tajo.ws.rs.resources.outputs;
+
+import org.apache.tajo.master.exec.NonForwardQueryResultScanner;
+
+import javax.ws.rs.core.HttpHeaders;
```
jihoonson: Please remove unused import.

tajo-core/src/main/java/org/apache/tajo/ws/rs/resources/outputs/CSVStreamingOutput.java
```diff
@@ -0,0 +1,134 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.tajo.ws.rs.resources.outputs;
+
+import com.google.protobuf.ByteString;
```
jihoonson: Please remove unused imports.

tajo-core/src/main/java/org/apache/tajo/ws/rs/resources/outputs/CSVStreamingOutput.java
```diff
@@ -0,0 +1,134 @@
...
+
+        outpuTupletList = scanner.getNextTupleRows(count);
+
+        StringBuilder sb = new StringBuilder();
+        if (startOffset == 0) {
+            Schema schema = this.scanner.getLogicalSchema();
+            List<Column> columns = schema.getAllColumns();
+            boolean first = true;
+            for (Column column : columns) {
+                if (first == false) {
```
jihoonson: This condition can be simplified to ```!first```.

tajo-core/src/main/java/org/apache/tajo/ws/rs/resources/outputs/CSVStreamingOutput.java
```diff
@@ -0,0 +1,134 @@
...
+import java.io.IOException;
+import java.io.OutputStream;
+import java.util.List;
+
+@RestReturnType(
+        description = "default binary streaming output",
+        headerType = "application/csv"
+)
+public class CSVStreamingOutput extends AbstractStreamingOutput {
+    private List<Tuple> outpuTupletList;
```
jihoonson: The list of tuples doesn't have to be kept. Removing this list will reduce the memory usage.

tajo-core/src/main/java/org/apache/tajo/ws/rs/resources/outputs/BinaryStreamingOutput.java
```diff
@@ -0,0 +1,79 @@
...
+public class BinaryStreamingOutput extends AbstractStreamingOutput {
+    private List<ByteString> outputList = null;
+
+    public BinaryStreamingOutput(NonForwardQueryResultScanner scanner, Integer count, Integer startOffset) throws IOException {
+        super(scanner, count, startOffset);
+    }
+
+    @Override
+    public boolean hasLength() {
+        return false;
```
jihoonson: Binary output length can also be calculated.

tajo-core/src/main/java/org/apache/tajo/ws/rs/resources/outputs/RestOutputFactory.java
```diff
@@ -0,0 +1,91 @@
...
+ * limitations under the License.
+ */
+
+package org.apache.tajo.ws.rs.resources.outputs;
+
+import com.facebook.presto.hive.shaded.com.google.common.collect.Maps;
+import com.facebook.presto.hive.shaded.org.apache.commons.lang.StringUtils;
+import org.apache.commons.logging.Log;
+import org.apache.commons.logging.LogFactory;
+import org.apache.tajo.catalog.FunctionDesc;
```
jihoonson: Please remove unused imports.

tajo-core/src/main/java/org/apache/tajo/ws/rs/resources/outputs/RestOutputFactory.java
```diff
@@ -0,0 +1,91 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.tajo.ws.rs.resources.outputs;
+
+import com.facebook.presto.hive.shaded.com.google.common.collect.Maps;
```
jihoonson: Please use google's library instead of facebook's one.

tajo-core/src/main/java/org/apache/tajo/ws/rs/annotation/RestReturnType.java
```diff
@@ -0,0 +1,37 @@
...
+
+/**
+ * Description.
+ *
+ */
+@Retention(RetentionPolicy.RUNTIME)
+@Target(ElementType.TYPE)
+public @interface RestReturnType {
+
+    String description() default "";
```
jihoonson: ```Description``` seems to be not used. Maybe you have a plan to use it later.
Would you share your intention?

tajo-core/src/main/java/org/apache/tajo/ws/rs/annotation/RestReturnType.java
```diff
@@ -0,0 +1,37 @@
...
+
+package org.apache.tajo.ws.rs.annotation;
+
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+/**
+ * Description.
```
jihoonson: Please add a description.

tajo-core/src/main/java/org/apache/tajo/ws/rs/resources/outputs/CSVStreamingOutput.java
```diff
@@ -0,0 +1,134 @@
...
+    private String output;
+    private boolean alreadyCalculated = false;
```
jihoonson: Please remove unused variables.

tajo-core/src/main/java/org/apache/tajo/ws/rs/resources/outputs/RestOutputFactory.java
```diff
@@ -0,0 +1,91 @@
...
+        Set<Class> restOutputClasses = ClassUtil.findClasses(AbstractStreamingOutput.class, "org.apache.tajo.ws.rs.resources.outputs");
+
+        for (Class eachClass : restOutputClasses) {
+            if (eachClass.isInterface() || Modifier.isAbstract(eachClass.getModifiers())) {
+                continue;
+            }
+
+            AbstractStreamingOutput streamingOutput = null;
+            try {
+                streamingOutput = (AbstractStreamingOutput) eachClass.getDeclaredConstructor(
```
jihoonson: ```streamingOutput``` instance is created unnecessarily.

jihoonson:
Hi @charsyam, I left some comments.

tajo-core/src/main/java/org/apache/tajo/ws/rs/resources/outputs/CSVStreamingOutput.java
charsyam: @jihoonson you mean just replace first == false -> !first
is it coding convention of tajo?
Actually, I prefer using this explict expression

but it is coding convention of tajo. I will replace :)

tajo-core/src/main/java/org/apache/tajo/ws/rs/resources/outputs/RestOutputFactory.java
charsyam: Thanks. Actually, I missed it. :)

tajo-core/src/main/java/org/apache/tajo/ws/rs/annotation/RestReturnType.java
charsyam: Actually, I think someone that want to give description need.
so, I think currently removing description is better :)
Thanks. I will remove it soon.

tajo-core/src/main/java/org/apache/tajo/ws/rs/resources/outputs/BinaryStreamingOutput.java
charsyam: @jihoonson Yes, we can.
But, I think BinaryStramingOutput class send bytestring with size.
so I think calculating length is not needed.
What do you think?

jihoonson: When BinaryStreamingOutput is used, Content-Length attribute of the http header can be filled by calculating the total length of the binary data. Content-Length attribute needs to be filled in most cases. Please see http://www.w3.org/Protocols/rfc2616/rfc2616-sec14.html

charsyam:
@jihoonson I added length for BinaryStreamingOutput.
but I can't change output style of BinaryStreamingOutput for Compatibility :) Thanks for your review :)

tajo-core/src/main/java/org/apache/tajo/ws/rs/annotation/RestReturnType.java
```diff
@@ -0,0 +1,34 @@
...
+package org.apache.tajo.ws.rs.annotation;
+
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+/**
+ * Definition of RestApi result Return Type
+ * According to headerType, StramingOutput class will be choose.
```
jihoonson: How about to change like below?
```
Definition of the return type of RestAPI.
According to the output type, RestAPIs return their results in text or binary.
```

tajo-core/src/main/java/org/apache/tajo/ws/rs/annotation/RestReturnType.java
```diff
@@ -0,0 +1,34 @@
...
+ */
+@Retention(RetentionPolicy.RUNTIME)
+@Target(ElementType.TYPE)
+public @interface RestReturnType {
+    String headerType();
```
jihoonson: ```header type``` sounds less intuitive to me. How about ```output type```?

tajo-core/src/main/java/org/apache/tajo/ws/rs/resources/outputs/RestOutputFactory.java
```diff
@@ -0,0 +1,85 @@
...
+    try {
+      if (restOutputClasses.containsKey(headerType)) {
+        String className = (String) restOutputClasses.get(headerType);
+        Class<?> clazz = Class.forName(className);
+        output = (AbstractStreamingOutput) clazz.getDeclaredConstructor(
+                  new Class[]{NonForwardQueryResultScanner.class,
+                  Integer.class, Integer.class})
+                  .newInstance(scanner, count, startOffset);
+      } else {
+        output = new BinaryStreamingOutput(scanner, count, startOffset);
```
jihoonson: I think most users are familiar with TextStreamingOutput. How about to use it as the default encoding format?

jihoonson:
@charsyam thank you for updating patch. I left some trivial comments.
In addition, please fix the test failure.
Thanks.

tajo-core/src/main/java/org/apache/tajo/ws/rs/annotation/RestReturnType.java
charsyam: @jihoonson how about mimeType? event though, It is type for output.
but, It has mime type like "application/csv" or "application/octet-stream"

tajo-core/src/main/java/org/apache/tajo/ws/rs/resources/outputs/RestOutputFactory.java
charsyam: @jihoonson I think it is better. but I also afraid of changing default output type because some users are already using it with binary type. What do you think?

tajo-core/src/main/java/org/apache/tajo/ws/rs/annotation/RestReturnType.java
charsyam: Thanks :)

tajo-core/src/main/java/org/apache/tajo/ws/rs/annotation/RestReturnType.java
jihoonson: Well, mimeType also looks good. If you prefer to mimeType, ```application/csv``` needs to be changed to ```text/csv```. Please refer to http://www.freeformatter.com/mime-types-list.html#mime-types-list.

tajo-core/src/main/java/org/apache/tajo/ws/rs/resources/outputs/RestOutputFactory.java
jihoonson: From that point, I think it will be ok because BinaryStreamingOutput is designed to be used for clients of other languages.

charsyam:
@jihoonson I applied your reivews Thanks :)
1. changing headerType -> mimeType
2. changing default output type to "CSV"
3. fix test and Description.

Thanks

jihoonson:
Thanks, but some tests are still failed. Please check it.

charsyam:
@jihoonson I think fail is because of travis ci's OOM.
Thanks. :)

tajo-core-tests/src/test/java/org/apache/tajo/ws/rs/resources/TestQueryResultResource.java
```diff
@@ -26,6 +26,7 @@
 import org.apache.tajo.exception.ErrorUtil;
 import org.apache.tajo.storage.RowStoreUtil;
 import org.apache.tajo.storage.Tuple;
+import org.apache.tajo.storage.thirdparty.orc.IntegerColumnStatistics;
```
jihoonson: Unnecessary import.

jihoonson:
Please remove the above unused import before commit.
Thanks!

PR #879, opened 2015-11-25
TAJO-1985: Supporting Metric report via JMX.

keybod:
I added a JMX metric reporter.

### DIFF
--- a/tajo-core/src/main/java/org/apache/tajo/util/metrics/TajoSystemMetrics.java
+++ b/tajo-core/src/main/java/org/apache/tajo/util/metrics/TajoSystemMetrics.java
@@ -18,6 +18,7 @@
 
 package org.apache.tajo.util.metrics;
 
+import com.codahale.metrics.JmxReporter;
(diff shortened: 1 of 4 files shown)


jinossy:
Travis failure is not related this PR
Could you trigger the travis CI?

tajo-core/src/main/java/org/apache/tajo/util/metrics/TajoSystemMetrics.java
```diff
@@ -113,6 +117,9 @@ public void start() {
 
     final String jvmMetricsName = metricsGroupName + "-JVM";
     setMetricsReporter(jvmMetricsName);
+    jmxReporter = JmxReporter.forRegistry(metricRegistry).inDomain("Tajo")
```
jinossy: This line should move into the ’if(!inited) {}’

keybod:
Thank you for reviewing!
I have applied it.

jinossy:
+1 LGTM!
Thanks for your contribution!

PR #883, opened 2015-11-27
TAJO-1997: Registering UDF, it needs to check duplication

eminency:


eminency:
Ready for review

tajo-catalog/tajo-catalog-common/src/main/java/org/apache/tajo/catalog/FunctionDesc.java
```diff
@@ -167,6 +171,13 @@ public boolean equals(Object obj) {
     }
     return false;
   }
+
+  public boolean equalsSignature(Object obj) {
```
jihoonson: Please remove the unused method.

tajo-core/src/main/java/org/apache/tajo/engine/function/FunctionLoader.java
```diff
@@ -299,4 +284,38 @@ private static StaticMethodInvocationDesc extractStaticMethodInvocation(Method m
...
+    // Build a map with a first list
+    for (FunctionDesc desc: baseFuncList) {
+      funcMap.put(desc.hashCodeWithoutType(), desc);
+    }
+
+    // Check duplicates for other function lists(should be UDFs practically)
+    for (int i=1; i<functionLists.length; i++) {
+      for (FunctionDesc desc: functionLists[i]) {
+        if (funcMap.containsKey(desc.hashCodeWithoutType())) {
+          throw new AmbiguousFunctionException(String.format("UDF %s", desc.toString()));
```
jihoonson: It would be better to print which function is ambiguous with found one.

tajo-core/src/main/java/org/apache/tajo/engine/function/FunctionLoader.java
```diff
@@ -299,4 +284,38 @@ private static StaticMethodInvocationDesc extractStaticMethodInvocation(Method m
...
+
+    // Build a map with a first list
+    for (FunctionDesc desc: baseFuncList) {
+      funcMap.put(desc.hashCodeWithoutType(), desc);
+    }
+
+    // Check duplicates for other function lists(should be UDFs practically)
+    for (int i=1; i<functionLists.length; i++) {
+      for (FunctionDesc desc: functionLists[i]) {
+        if (funcMap.containsKey(desc.hashCodeWithoutType())) {
```
jihoonson: As you know, the equality check with hash code involves a possibility of false positive. I know it is very rare, but we need to check the real equality for valid operation.

tajo-core/src/main/java/org/apache/tajo/engine/function/FunctionLoader.java
```diff
@@ -299,4 +284,38 @@ private static StaticMethodInvocationDesc extractStaticMethodInvocation(Method m
...
+    Map<Integer, FunctionDesc> funcMap = new HashMap<>();
+    List<FunctionDesc> baseFuncList = functionLists[0];
+
+    // Build a map with a first list
+    for (FunctionDesc desc: baseFuncList) {
+      funcMap.put(desc.hashCodeWithoutType(), desc);
+    }
+
+    // Check duplicates for other function lists(should be UDFs practically)
+    for (int i=1; i<functionLists.length; i++) {
```
jihoonson: You seem to assume that there is no conflict between built-in functions. But, to check our mistake, how about checking built-in functions too?

tajo-catalog/tajo-catalog-common/src/main/java/org/apache/tajo/catalog/FunctionDesc.java
jihoonson: Maybe this method is used to check the real equality between functions. If so, please fix the below code. In the below, function signature is compared with function desc.

eminency: It's used now.

tajo-core/src/main/java/org/apache/tajo/engine/function/FunctionLoader.java
eminency: Good suggestion! I fix it.

tajo-core/src/main/java/org/apache/tajo/engine/function/FunctionLoader.java
eminency: It is based on the error message template that has existed.

https://github.com/apache/tajo/blob/ef94bb38d225deaff2d1eeb3d916fbe765412c3a/tajo-common/src/main/java/org/apache/tajo/exception/ErrorMessages.java#L87

So to fix as you advised, some other positions need to be fixed.

If you need, I think you'd better create another issue.

tajo-core/src/main/java/org/apache/tajo/engine/function/FunctionLoader.java
eminency: I considered about that.
But, there are two reasons why I didn't decide to do it.

First is built-in functions exist statically, that is, they are not changed frequently. So I thought it could be useless burden to check each startup (number of built-in functions are more than 200, so number of checking will be more than 20K times).

Secondly, the check routine is not considering function type as you already know. But there are already duplicate functions in built-in functions except function type. For example, there is sum() with or without 'distinct' feature.
Since the check logic should be different, it should be done separately before current code part. 
Thus, I thought the task was not the part of this issue if it should be done.

eminency:
@jihoonson 
Thanks for the review.
I applied what you advised and left some answers.
Please check them.

tajo-core/src/main/java/org/apache/tajo/engine/function/FunctionLoader.java
jihoonson: I understand. Would you please leave some comments about this discussion?

jihoonson:
Thanks for update. I left a trivial comment.
The latest patch looks good to me. Would you rebase it?

eminency:
Hi, @jihoonson .
I applied what you suggested and rebased.
Could you verify them?

jihoonson:
I'll commit shortly.

PR #885, opened 2015-11-27
TAJO-1998: Waiting for gracefully stopping in tajo-deamon.sh is too short

jinossy:


blrunner:
+1 

I found that the patch ran as expected on testing cluster as following:
### Before TAJO-1998
```
$ ./bin/stop-tajo.sh 
Stopping single TajoMaster
stopping master
worker04: stopping worker
worker02: stopping worker
worker05: stopping worker
worker03: stopping worker
worker04: worker did not stop gracefully after 2 seconds: killing with kill -9
worker02: worker did not stop gracefully after 2 seconds: killing with kill -9
worker05: worker did not stop gracefully after 2 seconds: killing with kill -9
worker03: worker did not stop gracefully after 2 seconds: killing with kill -9
```

### After TAJO-1998
```
$ ./bin/stop-tajo.sh 
worker03: stopping worker
worker04: stopping worker
worker05: stopping worker
worker02: stopping worker
stopping master
```

PR #886, opened 2015-11-27
TAJO-1992 \set timezone in cli doesn't work because of casesensitive

charsyam:
\set timezone doesn't work.
default> \set timezone GMT-3

but TIMEZONE works.
default> \set TIMEZONE GMT-3

so, this patch change timeZone -> TIMEZONE.name();


hyunsik:
+1
The patch looks good to me.

PR #887, opened 2015-11-28
TAJO-1705: Update document for index support

jihoonson:


tajo-docs/src/main/sphinx/index/how_to_use.rst
```diff
@@ -1,18 +1,19 @@
...
 
--------------------------------------
+---------------
 1. Create index
--------------------------------------
+---------------
 
-The first step for utilizing index is index creation. You can create index using SQL (:doc:`/sql_language/ddl`) or Tajo API (:doc:`/tajo_client_api`). For example, you can create a BST index on the lineitem table by submitting the following SQL to Tajo.
+The first step for utilizing index is to create an index. You can create an index using SQL (:doc:`/sql_language/ddl`) or Tajo API (:doc:`/tajo_client_api`).
+For example, the following SQL statement can be used to create a BST index on the lineitem table.
```
hyunsik: How about "statement will create"?

tajo-docs/src/main/sphinx/index/how_to_use.rst
```diff
@@ -48,17 +49,17 @@ If the index is created successfully, you can see the information about that ind
...
 For more information about index creation, please refer to the above links.
 
--------------------------------------
+-----------------------------
 2. Enable/disable index scans
--------------------------------------
+-----------------------------
 
-When an index is successfully created, you must enable the index scan feature as follows:
+When an index is successfully created, you must enable index scan as follows:
```
hyunsik: I suggest:
``Even though an index is successfully created, you need a further step, enabling 'index scan' as following. It is necessary because 'index scan' is disabled by default.``

tajo-docs/src/main/sphinx/index/how_to_use.rst
```diff
@@ -66,4 +67,10 @@ If you don't want to use the index scan feature anymore, you can simply disable
 
 .. note::
 
-     Once the index scan feature is enabled, Tajo currently always performs the index scan regardless of its efficiency. You should set this option when the expected number of retrieved tuples is sufficiently small.
+     Once index scan is enabled, Tajo currently always performs index scan regardless of its efficiency. You should set this option when the expected number of retrieved tuples is sufficiently small.
```
hyunsik: I suggest:
``Once index scan is enabled, Tajo will perform 'index scan' if possible. In some cases, it may cause performance degradation. If you always want to get better performance, you should either enable or disable 'index scan' according to selectivity. Usually, the performance gain of index will increase when the selectivity is low.``

tajo-docs/src/main/sphinx/index/how_to_use.rst
jihoonson: This section is related to the previous section ```Create index```. So, starting with ```even though``` looks not good I think. 
How about ```Reading data using index is disabled by default.
So, exploiting the created index, you need a further step, enabling 'index scan' as following:```?

tajo-docs/src/main/sphinx/index/how_to_use.rst
jihoonson: Done.

tajo-docs/src/main/sphinx/index/how_to_use.rst
jihoonson: Thanks for the good comment.

jihoonson:
@hyunsik thanks for your comment. I've updated my patch.

hyunsik:
+1 LGTM

PR #888, opened 2015-11-28
TAJO-1993 Table Timezone doesn't work when Timezone is not exactly same.

charsyam:
Auto Converted

ASIA/Seoul -> Asia/Seoul
Asia/seoul -> Asia/Seoul

### DIFF
--- a/tajo-core/src/main/java/org/apache/tajo/util/TimeZoneUtil.java
+++ b/tajo-core/src/main/java/org/apache/tajo/util/TimeZoneUtil.java
@@ -18,27 +18,26 @@
 
 package org.apache.tajo.util;
 
-import com.facebook.presto.hive.shaded.com.google.common.collect.Maps;
-import org.apache.tajo.SessionVars;
+import com.google.common.collect.ImmutableMap;
+import com.google.common.collect.Maps;
 import org.apache.tajo.storage.StorageConstants;
-
 import java.util.Map;
 
 public class TimeZoneUtil {
-  static Map<String, String> timeZoneIdMap;
+  static ImmutableMap<String, String> timeZoneIdMap;
 
   static {
     timeZoneIdMap = load();
   }
 
-  private static Map<String, String> load() {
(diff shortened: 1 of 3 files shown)


tajo-core/src/main/java/org/apache/tajo/util/TimeZoneUtil.java
```diff
@@ -0,0 +1,64 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.tajo.util;
+
+import com.facebook.presto.hive.shaded.com.google.common.collect.Maps;
```
hyunsik: It should be ``com.google.common.collect.Maps`` instead of presto one.

tajo-core/src/main/java/org/apache/tajo/util/TimeZoneUtil.java
```diff
@@ -0,0 +1,64 @@
...
+import org.apache.tajo.SessionVars;
+import org.apache.tajo.storage.StorageConstants;
+
+import java.util.Map;
+
+public class TimeZoneUtil {
+  static Map<String, String> timeZoneIdMap;
```
hyunsik: It can be an ImmutableHashMap. Also, we can mark ``final`` for this member variable.

hyunsik:
The patch looks good to me. I leave some trivial comments.

charsyam:
@hyunsik I applied your review :) Thanks.

charsyam:
@hyunsik Thanks for your review :)
I applied it :)

hyunsik:
+1

One test is failed, but it is not related to your change. The patch looks good to me.

PR #893, opened 2015-12-02
TAJO-2002: Change log4j log level for aws

blrunner:
I found that the patch ran as expected on emr cluster.


dkhwangbo:
I have a question.
This patch has no mention about s3, s3n. just s3a is written. Can I get a explanation about it?

blrunner:
@dkhwangbo 

Honestly, I borrowed line 1,2,3 from hadoop default configuration.
For the reference, when I testing the patch with ```s://`` prefix on EMR, I couldn't see any log messages for S3Filesystem.

PR #897, opened 2015-12-03
TAJO-1948: Change GroupbyNode::setAggFunctions and getAggFunctions to set and get List

dkhwangbo:
See TAJO-1938. It would simplify lots of codes for 'for-loop' on Targets if we Change GroupbyNode::setAggFunctions and getAggFunctions to set and get List.


hyunsik:
+1 LGTM

PR #898, opened 2015-12-03
TAJO-2004: S3TableSpace needs to extend FileTableSpace.

blrunner:
Currently, ```S3TableSpace``` implemented by coping ```FileTableSpace```. But ```S3TableSpace``` need to extend ```FileTableSpace``` instead of source copy because there are many codes for using ```FileTableSpace``` in the master branch. In addition, this issue will remove duplicated codes in ```S3TableSpace```.


hyunsik:
+1 LGTM!

blrunner:
@hyunsik 

Thank you for quick review. :)

PR #900, opened 2015-12-03
TAJO-2007: By default, Optimizer should use the table volume in TableStat.

hyunsik:
This patch depends on TAJO-2005.


hyunsik:
I rebased and fixed all tests. It is ready to review.

blrunner:
I'm reviewing this PR. :)

tajo-core-tests/src/test/java/org/apache/tajo/engine/query/TestInnerJoinWithSubQuery.java
```diff
@@ -71,24 +71,6 @@ public final void testJoinWithMultipleJoinQual4() throws Exception {
   }
 
   @Test
-  public final void testJoinWithJson2() throws Exception {
```
blrunner: Could you explain why you remove this methods?

blrunner:
I leaved a trivial comment.

And could you explain why you added ``order by clause`` to partial queries?

tajo-core-tests/src/test/java/org/apache/tajo/engine/query/TestInnerJoinWithSubQuery.java
hyunsik: That is unnecessary and hard to change.

hyunsik:
@blrunner They were not deterministic. I added 'order by clause' to make some unit tests deterministic. 'deterministic' means that their results sometimes can be different.

blrunner:
@hyunsik Thank you for comments. I agree with your opinion.

tajo-core-tests/src/test/resources/results/TestCrossJoin/testCrossJoinOfOneSmallTableAndJoin.1.Hash.plan
```diff
@@ -37,21 +37,90 @@ Execution Block Graph (TERMINAL - eb_0000000000000_0000_000008)
 -------------------------------------------------------------------------------
 |-eb_0000000000000_0000_000008
    |-eb_0000000000000_0000_000007
+      |-eb_0000000000000_0000_000004
```
jihoonson: This plan should not be changed because it tests broadcast join.

tajo-core-tests/src/test/resources/results/TestCrossJoin/testCrossJoinOfOneSmallTableAndJoin.1.Sort.plan
```diff
@@ -37,21 +37,90 @@ Execution Block Graph (TERMINAL - eb_0000000000000_0000_000008)
 -------------------------------------------------------------------------------
 |-eb_0000000000000_0000_000008
    |-eb_0000000000000_0000_000007
+      |-eb_0000000000000_0000_000004
```
jihoonson: This test should not be changed because it tests broadcast join.

jihoonson:
Hi @hyunsik, I found some query plans are changed, especially related to broadcast join. I suppose this change is due to the table volumes stored in catalog, but it is difficult to find where the code is. Would you let me know some points where need to be reviewed?

jihoonson:
@hyunsik thanks for your work. Even though this patch contains lots of changes, the code change is very minor. These changes look good.
In addition, changes in tests also look good. Honestly, I think we can do more realistic query plan tests on TPC-H tables. 
The only thing I'm concerned with is that some tests are intended to test broadcast join, but are not actually executed with broadcast join. I think we need to change the test query. Do you have any good idea?

hyunsik:
@jihoonson Thank you for your comments. As we discussed in an offline, it may be hard to handle the problem in this issue. Instead, I just created the another jira [1] and add ``FIXME`` tags in the unit tests. We can resolve this problem after we introduce hints.

[1] https://issues.apache.org/jira/browse/TAJO-2026

jihoonson:
Thanks. It looks good to me. In that issue, we need to investigate the proper goal of each join tests and make them work properly.
Here is my +1. Thanks for your work!

hyunsik:
Thank you! I'll commit it soon.

PR #901, opened 2015-12-03
TAJO-1991: Tablespace::getVolume should take filter predication.

hyunsik:
See https://issues.apache.org/jira/browse/TAJO-1991.

This patch depends on TAJO-2007.


tajo-plan/src/main/java/org/apache/tajo/plan/rewrite/TableStatUpdateRewriter.java
```diff
@@ -0,0 +1,124 @@
...
+      try {
+        if (table.getStats() != null) {
+          return storage.getTableVolumn(table.getUri(), filter);
+        }
+      } catch (UnsupportedException t) {
+        LOG.warn(table.getName() + " does not support Tablespace::getTableVolume()");
+        // By default, return 1GB to avoid a single task
+      }
+
+      return StorageUnit.GB;
```
jinossy: How about the configurable value ?

hyunsik:
I recover the default 1GB to the original value -1. Actually, it was related to many other code parts. It should be resolved in another jira.

hyunsik:
I rebased and fixed all tests. It is ready to review.

hyunsik:
Rebased.

jihoonson:
This patch looks good to me. Test failure seems not related to the changes in this PR.

hyunsik:
Rebased. Thanks!

jihoonson:
+1 ship it!

PR #903, opened 2015-12-03
TAJO-2010: Parquet can not read null value

jinossy:


tajo-storage/tajo-storage-hdfs/src/main/java/org/apache/tajo/storage/parquet/TajoRecordConverter.java
```diff
@@ -154,14 +157,6 @@ public void start() {
    */
   @Override
   public void end() {
-    for (int i = 0; i < projectionMap.length; ++i) {
```
blrunner: Could you explain why you remove existing codes?

jinossy: You can find it in start()

blrunner:
+1

Ship it. :)

jinossy:
Thanks for your quick review!

PR #904, opened 2015-12-08
TAJO-2013: FilterPushDownRule fails due to the unsupported index.

jinossy:


blrunner:
+1 

I found that ```UnsupportedOperationException``` doesn't appear after using this patch on HiveCatalogStore.

PR #907, opened 2015-12-11
TAJO-2017: Replace manual array copy with Collection

dkhwangbo:
Copying of array contents to a collection where each element is added individually using a for loop. Such constructs may be replaced by a call to Collection.addAll(Arrays.asList()) or Collections.addAll().


jinossy:
Would you rebase your patch?

dkhwangbo:
@jinossy Hi! I rebase this patch. Thanks for your notice.

jinossy:
+1 LGTM

PR #908, opened 2015-12-11
TAJO-2018: Remove redundant 'String.toString()'

dkhwangbo:
Remove any to call toString() on a String object.


jinossy:
+1 LGTM!
Thanks for your contribution!

PR #909, opened 2015-12-11
TAJO-2019: Replace manual array copy with System.arraycopy()

dkhwangbo:
Manual copying of array contents can be replaced by calls to System.arraycopy(). It makes code line lighten.


jihoonson:
+1 LGTM.

PR #910, opened 2015-12-14
TAJO-1940: Implement HBaseTablespace::getTableVolume() method.

hyunsik:
This patch depends on TAJO-1990, TAJO-1991 and TAJO-2007. It is almost done. It only needs some more unit tests.


hyunsik:
rebased.

tajo-storage/tajo-storage-hbase/src/main/java/org/apache/tajo/storage/hbase/HBaseTablespace.java
```diff
@@ -100,8 +102,19 @@ public Configuration getHbaseConf() {
   }
 
   @Override
-  public long getTableVolume(URI uri, Optional<EvalNode> filter) throws UnsupportedException {
-    throw new UnsupportedException();
+  public long getTableVolume(TableDesc table, Optional<EvalNode> filter) throws UnsupportedException {
```
jihoonson: Please remove ```UnsupportedException```.

jihoonson:
Thanks @hyunsik. This patch looks good to me overall. I have one suggestion.
It looks that ```getTableVolume()``` works well, but we need to verify its behaviour. How about adding any test?

hyunsik:
Added unit tests and rebased the patch.

tajo-storage/tajo-storage-hbase/src/main/java/org/apache/tajo/storage/hbase/RegionSizeCalculator.java
```diff
@@ -0,0 +1,147 @@
...
+    //iterate all cluster regions, filter regions from our table and compute their size
+    for (ServerName serverName: servers) {
+      ServerLoad serverLoad = clusterStatus.getLoad(serverName);
+
+      for (RegionLoad regionLoad: serverLoad.getRegionsLoad().values()) {
+        byte[] regionId = regionLoad.getName();
+
+        if (tableRegions.contains(regionId)) {
+
+          long regionSizeBytes = (regionLoad.getStorefileSizeMB() + regionLoad.getMemStoreSizeMB()) * megaByte;
```
jihoonson: This line looks to be improved to consider the mem store size for region size calculation. Would you leave some comments?

tajo-storage/tajo-storage-hbase/src/main/java/org/apache/tajo/storage/hbase/HBaseTablespace.java
```diff
@@ -101,8 +104,19 @@ public Configuration getHbaseConf() {
...
   @Override
-  public long getTableVolume(URI uri, Optional<EvalNode> filter) throws UnsupportedException {
-    throw new UnsupportedException();
+  public long getTableVolume(TableDesc table, Optional<EvalNode> filter) {
+    long totalVolume;
+    try {
+      totalVolume = getRawSplits("", table, filter.orElse(null)).stream()
+          .map(f -> f.getLength())
+          .filter(size -> size > 0) // eliminate unknown sizes (-1)
+          .reduce(0L, (sum, size) -> sum + size);
```
jihoonson: This line can be more simplified into ```.reduce(0L, Long::sum)```.

tajo-core-tests/src/test/java/org/apache/tajo/engine/query/TestHBaseTable.java
```diff
@@ -210,7 +210,62 @@ public void testCreateExternalHBaseTable() throws Exception {
...
+      put.add("col1".getBytes(), "b".getBytes(), ("b-" + i).getBytes());
+      put.add("col2".getBytes(), "k1".getBytes(), ("k1-" + i).getBytes());
+      put.add("col2".getBytes(), "k2".getBytes(), ("k2-" + i).getBytes());
+      put.add("col3".getBytes(), "b".getBytes(), ("b-" + i).getBytes());
+      htable.put(put);
+    }
+  }
+
+  @Test
+  public void testGetTableVolume() throws Exception {
```
jihoonson: This test looks good, but I'm concerned with its location.
IMO, more proper class which should contain this test is ```TestHBaseTablespace```, but, as you already know, that class doesn't involve running a test hbase cluster. 
So, if we move this test to ```TestHBaseTablespace```, we also need to launch another hbase cluster which can increase our test time. 
What do you think?

hyunsik:
Yes, that's why I placed the test in TestHbaseTable. As you know, Hbase cluster requires lots of additional memory, and we are suffering from memory problem of unit tests.

tajo-storage/tajo-storage-hbase/src/main/java/org/apache/tajo/storage/hbase/RegionSizeCalculator.java
hyunsik: The original RegionSizeCalculator does not concern the size of memory store. So, I forked it and modified the class to use region files as well as memory store.

tajo-storage/tajo-storage-hbase/src/main/java/org/apache/tajo/storage/hbase/HBaseTablespace.java
hyunsik: Thanks. I'll fix it.

hyunsik:
Rebased and reflected a comment.

jihoonson:
I've tested the latest patch manually.
Before you commit, please add some comments about the difference between the original RegionSizeCalculator and the copied one on https://github.com/apache/tajo/pull/910#discussion-diff-50492684.
Here is my +1.

hyunsik:
Thank you. I've added some note for that.

PR #917, opened 2015-12-18
TAJO-2029: Showing query progress percentage in TAJO JDBC

combineads:
It would be nice to check out query progress percentage in TAJO JDBC like TAJO CLI already did. The purpose is simple. I want to verify whether query is in block or not by monitoring query progress percentage.


jihoonson:
Hi @combineads, thanks for your patch.
However, I'm sorry to say but this patch looks a sort of workaround to see the query progress. I think we need to devise a better solution.

combineads:
I just need to check the execution progress of query.

The issue would be close?

Thanks for response @jihoonson .

jihoonson:
I appreciate if you do. Thanks.

PR #919, opened 2015-12-22
TAJO-1858: Aligning error message in execute query page of web UI is needed.

combineads:
Fixed using <pre> tag.
screenshot from 2015-12-22 13 18 03


jihoonson:
Hi @combineads, thanks for your patch. This patch contains the changes of other issue. Would you remove it?

combineads:
Thank you for response.

jihoonson:
@combineads, it seems that there are some misunderstanding between us. I mean, this patch includes some changes of TAJO-1858, so you need to remove those changes. 
This issue makes sense, so I'd like to help you to finish your work. If you still want to contribute this patch to Tajo, would you open this PR again or make a new one?

PR #920, opened 2015-12-23
TAJO-2031: Provides a simple contribution script for some beginner.

combineads:
It is a simple shell script for some beginner developers should be to contribute the Tajo project in the github.


blrunner:
Hi @combineads 

Thank you for your contribution. But I don't sense that ``sync`` is right name for contribution. If you create a document file instead of shell script file, contributors will know that it is a guide for contribution by file name and readability of github would be better in comparison with shell script. I recommend the file name is ``CONTRIBUTING.md``.

blrunner:
Sorry @combineads, I thought about it again. Tajo already provides some documents for contributing as follows. https://cwiki.apache.org/confluence/display/TAJO/How+to+Contribute+Codes+to+Tajo.
If possible, could you update existing wiki page? If you don't have a wiki account, you can request it to dev@tajo.apache.org.

PR #921, opened 2015-12-24
TAJO-1858: Aligning error message in execute query page of web UI is needed.

combineads:
Fixed using pre tag.
screenshot from 2015-12-22 13 18 03


combineads:
Hi @jihoonson , sorry for my misunderstanding.
I've clean up the files.

jihoonson:
+1. Even though test is failed, but it is not related to this patch.
I'll commit shortly.
Thanks!

PR #922, opened 2015-12-29
TAJO-2032: Change tajo.rest.service.port to be the general style in TAJO.

combineads:
Change the variable name to tajo.rest.service-http.address from tajo.rest.service.port with the default value 0.0.0.0:26880. And then it was modified to refer to several classes.

### DIFF
--- a/tajo-common/src/main/java/org/apache/tajo/conf/TajoConf.java
+++ b/tajo-common/src/main/java/org/apache/tajo/conf/TajoConf.java
@@ -48,7 +48,7 @@ import java.util.concurrent.TimeUnit;
 public class TajoConf extends Configuration {
   private static TimeZone SYSTEM_TIMEZONE;
   private static int DATE_ORDER = -1;
-  
+
   private static final Map<String, ConfVars> vars = new HashMap<>();
 
   static {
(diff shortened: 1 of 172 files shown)


tajo-common/src/main/java/org/apache/tajo/conf/TajoConf.java
```diff
@@ -140,7 +140,7 @@ public static int setDateOrder(int dateOrder) {
     TAJO_MASTER_INFO_ADDRESS("tajo.master.info-http.address", "0.0.0.0:26080", Validators.networkAddr()),
 
     // Tajo Rest Service
-    REST_SERVICE_PORT("tajo.rest.service.port", 26880),
+    REST_SERVICE_ADDRESS("tajo.rest.service-http.address", "0.0.0.0:26880", Validators.networkAddr()),
```
jihoonson: IMO, ```tajo.rest.service-http.address``` seems to have a duplicate meaning because rest service is mostly based on http. How about ```tajo.rest.service.address```?

combineads:
Hi @jihoonson .
I was to change a property name to tajo.rest.service.address from tajo.rest.service-http.address.
Thank you.

combineads:
@jihoonson Anything else to commit this PR?

jihoonson:
Sorry for late response. 
+1, This patch looks good to me.

PR #923, opened 2016-01-04
[finding travis-ci failure] find MiniHDFSCluster hangs

jinossy:


jinossy:
first and last failure is not related this issue after jetty upgraded

PR #924, opened 2016-01-05
TAJO-2033: Printing out query status with progress bar in TSQL

dkhwangbo:
Currently, TSQL print out query running progress and query running time line-by-line. To improve this UI, I adopt simple progress bar.


jihoonson:
Hi @dkhwangbo, your patch looks good to me. However, I'm concerned with the test failure at https://travis-ci.org/apache/tajo/builds/101996748. I've never seen before, so would you investigate what caused this failure?

dkhwangbo:
@jihoonson Hi! Thanks for your reply. To investigate the test failure you mentioned, I ran that test code on my local machine. There's no failure. Also, I pushed several travis CI trigger and same failure was not appeared.
(On the other hand, Another failure was appeared in https://travis-ci.org/apache/tajo/builds/103794425. )
This PR has no modification about these failured test's code. Condition difference in build-by-build has no relation about this PR, I guess. It means, I find no evidence yet about this problem. Can I get some advice or tip to resolve it?

dkhwangbo:
@jihoonson @hyunsik Hi! Thanks for your helpful comment. I make using ```println()``` instead of ```print()``` in ```SetCommand::set``` you mentioned. To make it, I rewrite some part in ```TajoCli::waitforQueryCompleted``` . Can I get some review message about updated commit?

tajo-cli/src/main/java/org/apache/tajo/cli/tsql/DefaultTajoCliOutputFormatter.java
```diff
@@ -206,4 +278,30 @@ public static String parseErrorMessage(String message) {
 
     return message;
   }
+
+  private static boolean detectRealTerminal()
```
jihoonson: You seem to borrow this code from Presto. If so, you need to add a comment explicitly.
And why did you exclude comments?

tajo-cli/src/main/java/org/apache/tajo/cli/tsql/TajoCli.java
```diff
@@ -607,8 +608,11 @@ private void waitForQueryCompleted(QueryId queryId) {
           continue;
         }
 
-        if (TajoClientUtil.isQueryRunning(status.getState()) || status.getState() == QueryState.QUERY_SUCCEEDED) {
+        if (TajoClientUtil.isQueryRunning(status.getState())) {
           displayFormatter.printProgress(sout, status);
+          if (isProgressPrinting == false) {
```
jihoonson: ```[val] == [boolean]``` is a bad idiom.

tajo-cli/src/main/java/org/apache/tajo/cli/tsql/TajoCli.java
```diff
@@ -619,6 +623,10 @@ private void waitForQueryCompleted(QueryId queryId) {
         }
       }
 
+      if (isProgressPrinting == true) {
```
jihoonson: ```[val] == [boolean]``` is a bad idiom.

tajo-cli/src/main/java/org/apache/tajo/cli/tsql/TajoCli.java
```diff
@@ -619,6 +623,10 @@ private void waitForQueryCompleted(QueryId queryId) {
...
+        displayFormatter.printProgress(sout, status);
+        sout.println();
```
jihoonson: Here, you added a new line. What does this new line mean? Please add some comments on that.

tajo-cli/src/main/java/org/apache/tajo/cli/tsql/DefaultTajoCliOutputFormatter.java
dkhwangbo: Yes. You're right. I've not known the rule about code redistribution. I'll add the comment explicitly.

tajo-cli/src/main/java/org/apache/tajo/cli/tsql/TajoCli.java
dkhwangbo: OK. I'll fix it.

tajo-cli/src/main/java/org/apache/tajo/cli/tsql/TajoCli.java
dkhwangbo: OK. I'll fix it.

tajo-cli/src/main/java/org/apache/tajo/cli/tsql/TajoCli.java
dkhwangbo: ```printProgress()`` only express the progress of query. To print query result in next line, add a new line is needed.

dkhwangbo:
@jihoonson Hi. Thanks for your comment. I reflect your mention. Please review it again.

tajo-cli/src/main/java/org/apache/tajo/cli/tsql/DefaultTajoCliOutputFormatter.java
jihoonson: This is not a rule, but is required for code maintenance.

tajo-cli/src/main/java/org/apache/tajo/cli/tsql/DefaultTajoCliOutputFormatter.java
```diff
@@ -206,4 +278,33 @@ public static String parseErrorMessage(String message) {
 
     return message;
   }
+
+  /**
+   * borrowed from Presto
+   */
+  private static boolean detectRealTerminal()
```
jihoonson: My previous question was too short, so you may miss my point. 
I mean, why are the comments in Presto's detectRealTerminal() excluded when you borrow?
Please see https://github.com/facebook/presto/blob/master/presto-cli/src/main/java/com/facebook/presto/cli/ConsolePrinter.java. It will be helpful for us as well.

dkhwangbo: Oh, I see. When I borrowed this method, I wondered about fitness of this comments to tajo. Currently, This comments are written for Presto, not Tajo. So I just confused about this comments to include this PR. But, After Testing this PR in many individual environment, I take conviction about this method, but I missed to include its comments. This is my mistake. I'll add comments.

jihoonson:
+1 LGTM.

jihoonson:
I found some codes which don't follow our coding convention, and fixed them before I commit.

PR #925, opened 2016-01-05
TAJO-2034: Files required for executing python functions are not copi…

jihoonson:
…ed in testEval().


eminency:
That's clear. My +1 here is.

hyunsik:
The patch looks straightforward. +1

PR #928, opened 2016-01-06
TAJO-2036: Prevent out of memory in the master server, if the query result is large

combineads:
This problem will occur frequently in our production server.
Please confirm.

### DIFF
--- a/tajo-core/src/main/java/org/apache/tajo/webapp/QueryExecutorServlet.java
+++ b/tajo-core/src/main/java/org/apache/tajo/webapp/QueryExecutorServlet.java
@@ -162,6 +162,11 @@ public class QueryExecutorServlet extends HttpServlet {
         } catch (java.lang.NumberFormatException nfe) {
           queryRunner.sizeLimit = 1048576;
         }
+        try {
+          queryRunner.rowLimit = Integer.parseInt(request.getParameter("limitRow"));
+        } catch (java.lang.NumberFormatException nfe) {
+          queryRunner.rowLimit = 3000000;
+        }
         synchronized(queryRunners) {
           queryRunners.put(queryRunnerId, queryRunner);
         }
@@ -286,6 +291,7 @@ public class QueryExecutorServlet extends HttpServlet {
     String database;
     long resultRows;
     int sizeLimit;
+    long rowLimit;
     Exception error;
 
     AtomicInteger progress = new AtomicInteger(0);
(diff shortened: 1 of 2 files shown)


tajo-core/src/main/java/org/apache/tajo/webapp/QueryExecutorServlet.java
```diff
@@ -507,14 +508,20 @@ private void MakeResultText(ResultSet res, TableDesc desc) throws SQLException {
         numOfRows = resultRows;
       }
 
+      int currentResultSize = 0;
       int rowCount = 0;
       while (res.next()) {
-        if(rowCount > numOfRows) {
+        if(rowCount > numOfRows || currentResultSize > sizeLimit) {
```
jihoonson: ```numOfRows = (long)((float)(resultRows) * ((float)sizeLimit / (float) resultRows));``` on the 506 line looks not necessary. Would you remove it?

combineads:
@jihoonson Anything else to commit this PR?

jihoonson:
@combineads, sorry for late response. 
Actually, my intention was to keep both factors, but clean some unnecessary codes.
I think we need to keep only the max number of rows rather than max byte size in the future. However, for now, it would be ok if we support both factors. How about adding the max number of rows again?

combineads:
Hi @jihoonson .
I did to add the row limit field.
Thank you for review.

tajo-core/src/main/java/org/apache/tajo/webapp/QueryExecutorServlet.java
```diff
@@ -197,7 +203,6 @@ public void service(HttpServletRequest request,
             errorResponse(response, queryRunner.error);
             return;
           }
-          returnValue.put("numOfRows", queryRunner.numOfRows);
```
jihoonson: limitRow also needs to be added to returnValue.

tajo-core/src/main/resources/webapps/admin/query_executor.jsp
```diff
@@ -87,6 +88,11 @@ function runQuery() {
   } else if(Math.ceil(Number($("#sizeLimit").val())) > 0) {
     SIZE_LIMIT = Number($("#sizeLimit").val()) * 1024 * 1024;
   }
+  if(Math.ceil(Number($("#rowLimit").val())) >= 10) {
```
jihoonson: IMO, we don't have to set the upper bound. It would be better to allow a very large value for rowLimit if users want.

tajo-core/src/main/java/org/apache/tajo/webapp/QueryExecutorServlet.java
jihoonson: Oh, my mistake. Never mind.

combineads:
Hi @jihoonson .
I did to remove the max row limit.
Thank you for review.

jihoonson:
+1 LGTM. This patch does not relate to unit tests.
I'll commit shortly.

PR #929, opened 2016-01-07
TAJO-1686: Allow Tajo to use Hive UDF

eminency:
This patch consists of parts below.

#### Core Logic 
1. **TajoMaster.java / HiveFunctionLoader.java** : in charge of finding Hive UDFs and registering them.
2. **HiveFunctionInvoke.java** : to call Hive UDF

#### Utility
1. **WritableTypeConverter.java** : utility class, to convert from Writable type to Datum and vice versa.
2. **FunctionDescBuilder.java** : helper/builder to build FunctionDesc more conveniently.

and others.


tajo-plan/src/main/java/org/apache/tajo/plan/function/HiveFunctionInvoke.java
```diff
@@ -0,0 +1,95 @@
...
+
+    try {
+      instance = constructor.newInstance();
+    } catch (InstantiationException|IllegalAccessException|InvocationTargetException e) {
+      throw new TajoInternalError(e);
+    }
+
+    for (Method m: clazz.getMethods()) {
+      if (m.getName().equals("evaluate")) {
+        return m;
```
jihoonson: If there are several functions of the same name which accept different arguments, this looks to return a wrong function.

tajo-plan/src/main/java/org/apache/tajo/plan/function/HiveFunctionInvoke.java
```diff
@@ -0,0 +1,95 @@
...
+      }
+    }
+
+    return null;
+  }
+
+  @Override
+  public Datum eval(Tuple tuple) {
+    Datum resultDatum;
+    Writable [] params = new Writable[tuple.size()];
```
jihoonson: How about making this ```params``` as a member variable? Its contents can be overwritten per ```eval()``` call instead of creating new one.

tajo-core/src/main/java/org/apache/tajo/engine/function/hiveudf/HiveFunctionLoader.java
```diff
@@ -0,0 +1,154 @@
...
+import java.net.URLClassLoader;
+import java.util.*;
+
+public class HiveFunctionLoader {
+  public static Collection<FunctionDesc> loadHiveUDFs(TajoConf conf) {
+    ArrayList<FunctionDesc> funcList = new ArrayList<>();
+    String udfdir = conf.getVar(TajoConf.ConfVars.HIVE_UDF_DIR);
+
+    try {
+      FileSystem localFS = FileSystem.getLocal(conf);
```
jihoonson: Hive basically assumes that the Jar files are stored on HDFS. We also need to support it.

tajo-common/src/main/java/org/apache/tajo/util/WritableTypeConverter.java
```diff
@@ -0,0 +1,103 @@
...
+      case INT1: return new ByteWritable(value.asByte());
+      case INT2: return new ShortWritable(value.asInt2());
+      case INT4: return new IntWritable(value.asInt4());
+      case INT8: return new LongWritable(value.asInt8());
+
+      case FLOAT4: return new FloatWritable(value.asFloat4());
+      case FLOAT8: return new DoubleWritable(value.asFloat8());
+
+      case TEXT: return new Text(value.asChars());
+      case VARBINARY: return new BytesWritable(value.asByteArray());
```
jihoonson: We need to support date, time, timestamp, and boolean types as well.

tajo-core/src/main/java/org/apache/tajo/worker/TajoWorker.java
```diff
@@ -200,6 +200,7 @@ public void serviceInit(Configuration conf) throws Exception {
     historyReader = new HistoryReader(workerContext.getWorkerName(), this.systemConf);
 
     FunctionLoader.loadUserDefinedFunctions(systemConf);
+    HiveFunctionLoader.loadHiveUDFs(systemConf);
```
jihoonson: It seems that TajoWorker also needs to check duplicated functions as TajoMaster. In addition, it will be better if the routine to load functions is consistent no matter where it is called from.

jihoonson:
Hi @eminency, thanks for your work. It seems good overall.
I left some comments. In addition, would you briefly evaluate the performance of Hive UDF call? I wonder how large its overhead is.

eminency:
I missed many things... I will make the code handle them soon.
UDF call test also will be done ASAP.
Thanks @jihoonson .

tajo-plan/src/main/java/org/apache/tajo/plan/function/HiveFunctionInvoke.java
eminency: Good idea!

tajo-core/src/main/java/org/apache/tajo/worker/TajoWorker.java
eminency: If any duplicate exists, it is already designed to raise AmbiguousFunctionException.
So this point might be skipped.

eminency:
@jihoonson 
I applied most of your suggestions.
Could you check them again?

tajo-common/src/main/java/org/apache/tajo/util/WritableTypeConverter.java
```diff
@@ -0,0 +1,127 @@
...
+
+      case DATE: return new DateDatum(((DateWritable)value).getDays() + DateTimeConstants.UNIX_EPOCH_JDATE);
+      case TIMESTAMP: return new TimestampDatum(DateTimeUtil.javaTimeToJulianTime(
+          ((TimestampWritable)value).getTimestamp().getTime()));
+
+      case TEXT: return new TextDatum(value.toString());
+      case VARBINARY: return new BlobDatum(((BytesWritable)value).getBytes());
+    }
+
+    return NullDatum.get();
```
jihoonson: I think throwing UnsupportedDataTypeException looks more proper.

tajo-common/src/test/java/org/apache/tajo/util/TestWritableTypeConverter.java
```diff
@@ -0,0 +1,65 @@
...
+import org.apache.tajo.datum.DateDatum;
+import org.apache.tajo.datum.Datum;
+import org.apache.tajo.datum.TimestampDatum;
+import org.apache.tajo.util.datetime.DateTimeUtil;
+import org.apache.tajo.util.datetime.TimeMeta;
+import org.junit.Test;
+
+import static org.junit.Assert.*;
+
+public class TestWritableTypeConverter {
```
jihoonson: It would be better if you add some negative tests.

tajo-core/src/main/java/org/apache/tajo/engine/function/hiveudf/HiveFunctionLoader.java
```diff
@@ -0,0 +1,161 @@
...
+  public static Collection<FunctionDesc> loadHiveUDFs(TajoConf conf) {
+    ArrayList<FunctionDesc> funcList = new ArrayList<>();
+    String udfdir = conf.getVar(TajoConf.ConfVars.HIVE_UDF_DIR);
+
+    try {
+      Path udfPath = new Path(udfdir);
+      FileSystem fs = udfPath.getFileSystem(conf);
+
+      if (!fs.isDirectory(udfPath)) {
+        return null;
```
jihoonson: How about using Optional?

tajo-plan/src/main/java/org/apache/tajo/plan/function/HiveFunctionInvoke.java
```diff
@@ -0,0 +1,112 @@
...
+    UDFInvocationDesc udfDesc = functionDesc.getInvocation().getUDF();
+
+    URL [] urls = new URL [] { new URL(udfDesc.getPath()) };
+    URLClassLoader loader = new URLClassLoader(urls);
+
+    try {
+      Class<?> udfclass = loader.loadClass(udfDesc.getName());
+      evalMethod = getEvaluateMethod(functionDesc.getParamTypes(), udfclass);
+    } catch (ClassNotFoundException e) {
+      e.printStackTrace();
```
jihoonson: Throwing exception is required.

jihoonson:
Sorry for late review. I left some comments.

tajo-catalog/tajo-catalog-common/src/main/java/org/apache/tajo/catalog/FunctionDescBuilder.java
```diff
@@ -0,0 +1,105 @@
...
+ * limitations under the License.
+ */
+
+package org.apache.tajo.catalog;
+
+import org.apache.tajo.catalog.proto.CatalogProtos.FunctionType;
+import org.apache.tajo.common.TajoDataTypes.DataType;
+import org.apache.tajo.function.*;
+
+public class FunctionDescBuilder {
```
jihoonson: It is seemingly similar to FunctionDescProto.Builder. Would you have any reason for adding a new class instead of using PB's builder?

tajo-common/pom.xml
```diff
@@ -245,6 +249,12 @@
       <artifactId>powermock-api-mockito</artifactId>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>org.apache.hive</groupId>
```
jihoonson: This can increase the size of the tajo client. How about moving it to other modules like tajo-core and tajo-plan?

tajo-core/src/main/java/org/apache/tajo/engine/function/hiveudf/HiveFunctionLoader.java
```diff
@@ -0,0 +1,161 @@
...
+public class HiveFunctionLoader {
+  public static Optional<List<FunctionDesc>> loadHiveUDFs(TajoConf conf) {
+    ArrayList<FunctionDesc> funcList = new ArrayList<>();
+    String udfdir = conf.getVar(TajoConf.ConfVars.HIVE_UDF_DIR);
+
+    try {
+      Path udfPath = new Path(udfdir);
+      FileSystem fs = udfPath.getFileSystem(conf);
+
+      if (!fs.isDirectory(udfPath)) {
```
jihoonson: It would be good if we write some logs to let users know what's the problem.

tajo-common/src/main/java/org/apache/tajo/util/WritableTypeConverter.java
```diff
@@ -0,0 +1,128 @@
...
+import org.apache.tajo.util.datetime.DateTimeConstants;
+import org.apache.tajo.util.datetime.DateTimeUtil;
+import org.reflections.ReflectionUtils;
+
+import java.sql.Timestamp;
+import java.util.Set;
+
+public class WritableTypeConverter {
+
+  public static DataType convertWritableToTajoType(Class<? extends Writable> writableClass) {
```
jihoonson: Char type is missed.

tajo-common/src/main/java/org/apache/tajo/util/WritableTypeConverter.java
```diff
@@ -0,0 +1,128 @@
...
+      return builder.setType(Type.TIMESTAMP).build();
+    }
+    if (writableClass == BytesWritable.class || parents.contains(BytesWritable.class)) {
+      return builder.setType(Type.VARBINARY).build();
+    }
+
+    throw new TajoRuntimeException(new NotImplementedException(writableClass.getSimpleName()));
+  }
+
+  public static Writable convertDatum2Writable(Datum value) {
```
jihoonson: Char type is missed.

tajo-common/src/main/java/org/apache/tajo/conf/TajoConf.java
```diff
@@ -309,6 +309,9 @@ public static int setDateOrder(int dateOrder) {
     PYTHON_CODE_DIR("tajo.function.python.code-dir", ""),
     PYTHON_CONTROLLER_LOG_DIR("tajo.function.python.controller.log-dir", ""),
 
+    // HIVE UDF
+    HIVE_UDF_DIR("hive.udf.dir", "./lib/hiveudf"),
```
jihoonson: How about changing to tajo.function.hive.code-dir?

tajo-core-tests/src/test/java/org/apache/tajo/engine/function/hiveudf/TestHiveFunctionLoader.java
```diff
@@ -0,0 +1,149 @@
...
+import org.junit.Before;
+import org.junit.Test;
+
+import static org.junit.Assert.assertEquals;
+
+import java.net.URL;
+import java.sql.ResultSet;
+import java.util.*;
+
+public class TestHiveFunctionLoader {
```
jihoonson: IMO, it would be better to write this test class using ExprTestBase. Please refer to TestPythonFunctions.
In addition, we need to separate tests for HiveFunctionLoader and hive udf execution.

tajo-core/src/main/java/org/apache/tajo/engine/function/hiveudf/HiveFunctionLoader.java
```diff
@@ -0,0 +1,161 @@
...
+        return Optional.empty();
+      }
+
+      // loop each jar file
+      for (FileStatus fstatus : fs.listStatus(udfPath, (Path path) -> path.getName().endsWith(".jar"))) {
+
+        URL[] urls = new URL[]{new URL("jar:" + fstatus.getPath().toUri().toURL() + "!/")};
+
+        // extract and register UDF's decendants (legacy Hive UDF form)
+        Set<Class<? extends UDF>> udfClasses = getSubclassesFromJarEntry(urls, UDF.class);
```
jihoonson: Since getSubclassesFromJarEntry() can take the array of urls, it would be better to pull the getSubclassesFromJarEntry() and buildFunctionsFromUDF() calls out from the for loop.

tajo-core/src/main/java/org/apache/tajo/engine/function/hiveudf/HiveFunctionLoader.java
```diff
@@ -0,0 +1,161 @@
...
+        names = desc.name().split(",");
+        for (int i=0; i<names.length; i++) {
+          names[i] = names[i].trim();
+        }
+
+        value = desc.value();
+        extended = desc.extended();
+      }
+      else {
+        names = new String [] {clazz.getName().replace('.','_')};
```
jihoonson: class.getName() returns canonical name for udf classes, so this line will make a very long function name. How about transforming the simple name of the class from camel case to snake case?

tajo-core/src/main/java/org/apache/tajo/master/TajoMaster.java
```diff
@@ -215,12 +216,48 @@ public void serviceInit(Configuration conf) throws Exception {
...
+
+  /**
+   * Checks duplicates between pre-loaded functions and UDFs. And they are meged to funcList.
+   *
+   * @param udfs UDF list
+   * @param funcSet set for pre-loaded functions to match signature
+   * @param funcList list to be merged
+   * @throws AmbiguousFunctionException
+   */
+  private void checkUDFduplicateAndMerge(Optional<List<FunctionDesc>> udfs, HashMap<Integer, FunctionDesc> funcSet, List<FunctionDesc> funcList)
```
jihoonson: It would be better if you add a test for this method.

tajo-plan/src/main/java/org/apache/tajo/plan/function/HiveFunctionInvoke.java
```diff
@@ -0,0 +1,113 @@
...
+    Constructor constructor = clazz.getConstructors()[0];
+
+    try {
+      instance = constructor.newInstance();
+    } catch (InstantiationException|IllegalAccessException|InvocationTargetException e) {
+      throw new TajoInternalError(e);
+    }
+
+    for (Method m: clazz.getMethods()) {
+      if (m.getName().equals("evaluate")) {
```
jihoonson: Class.getMethod(String name, Class<?>... parameterTypes) can simplify this for loop.

tajo-plan/src/main/java/org/apache/tajo/plan/function/HiveFunctionInvoke.java
```diff
@@ -0,0 +1,113 @@
...
+        Class [] methodParamTypes = m.getParameterTypes();
+        if (checkParamTypes(methodParamTypes, paramTypes)) {
+          return m;
+        }
+      }
+    }
+
+    return null;
```
jihoonson: If 'evaluate' method is not found, an exception should be thrown. How about adding a new exception?

jihoonson:
Hi @eminency, I left some comments.
In addition, I found the below error when I start a tajo cluster. I put the hive udf jar file on hdfs and specified it in tajo-site.xml.

```
2016-03-09 14:34:06,613 FATAL org.apache.tajo.master.TajoMaster: Error starting TajoMaster
org.apache.tajo.exception.TajoInternalError: internal error: unknown protocol: hdfs
        at org.apache.tajo.engine.function.hiveudf.HiveFunctionLoader.loadHiveUDFs(HiveFunctionLoader.java:70)
        at org.apache.tajo.master.TajoMaster.loadFunctions(TajoMaster.java:229)
        at org.apache.tajo.master.TajoMaster.serviceInit(TajoMaster.java:181)
        at org.apache.hadoop.service.AbstractService.init(AbstractService.java:163)
        at org.apache.tajo.master.TajoMaster.main(TajoMaster.java:628)
```

tajo-catalog/tajo-catalog-common/src/main/java/org/apache/tajo/catalog/FunctionDescBuilder.java
eminency: I think it was just inconvenient and I didn't consider about Proto.Builder.
It'd be better to use it.

eminency: Actually, I found out a main difference. PB's builder has a nested structure. It must build children's PB at first and build its own PB. So it looks complex to use.

The builder I made receives all informations needed at same level, then builds FunctionDesc including children objects at once. So it doesn't have to consider objects' hierarchy.

Which one can be better do you think?

tajo-common/pom.xml
eminency: I will move it into tajo-plan.

tajo-core/src/main/java/org/apache/tajo/engine/function/hiveudf/HiveFunctionLoader.java
eminency: This part is from Hive code and I guess it is to avoid name conflict.
Keeping it is not bad in my opinion because it can force to use the annotation.

tajo-plan/src/main/java/org/apache/tajo/plan/function/HiveFunctionInvoke.java
eminency: What a good point

eminency: To do that, Tajo type should be converted to Writable type, but Writable type cannot be fixed. For example, it can be inherited.

So comparing between Tajo types can be more robust.

eminency:
@jihoonson 
I fixed some of what you suggested. Read my comments.
In addition, loading from HDFS might be separated to another issue. It needs too large working to fix it.

tajo-core-tests/pom.xml
```diff
@@ -323,6 +323,10 @@
         </exclusion>
       </exclusions>
     </dependency>
+    <dependency>
+      <groupId>org.apache.hive</groupId>
+      <artifactId>hive-exec</artifactId>
```
jihoonson: The dependency scope should be test.

tajo-core/pom.xml
```diff
@@ -252,6 +252,10 @@
       <artifactId>hadoop-yarn-server-common</artifactId>
       <scope>provided</scope>
     </dependency>
+    <dependency>
+      <groupId>org.apache.hive</groupId>
+      <artifactId>hive-exec</artifactId>
```
jihoonson: The dependency scope should be provided.

tajo-plan/pom.xml
```diff
@@ -173,6 +173,10 @@
       <artifactId>hadoop-common</artifactId>
     </dependency>
     <dependency>
+      <groupId>org.apache.hive</groupId>
+      <artifactId>hive-exec</artifactId>
```
jihoonson: The dependency scope should be provided.

tajo-project/pom.xml
```diff
@@ -1007,6 +1006,12 @@
           </exclusion>
         </exclusions>
       </dependency>
+      <dependency>
+        <groupId>org.apache.hive</groupId>
+        <artifactId>hive-exec</artifactId>
+        <version>${hive.version}</version>
+        <scope>provided</scope>
```
jihoonson: I missed this. Please ignore the above comments.

jihoonson:
Hi @eminency, I met this warning message while starting up Tajo. This is weird because I configured the path to hive udf like below. Would you check it?

```
# in tajo-site.xml
 <property>
  <name>tajo.function.hive.code-dir</name>
  <value>${TAJO_HOME}/tajo-core-tests/src/test/resources/hiveudf/</value>
 </property>

# in TajoCli
default> select my_upper('abcd');
ERROR: function does not exist: my_upper(text)

# in the master log
...
2016-04-18 15:15:19,987 WARN org.apache.tajo.engine.function.hiveudf.HiveFunctionLoader: Hive UDF directory doesn't exist
...
```

eminency:
@jihoonson 
The problem you mentioned and null handling problem are fixed.

jihoonson:
+1 the latest patch looks good to me.
@eminency thank you for your work!

PR #930, opened 2016-01-07
TAJO-2038: NPE in FileScanner#getProgress.

jinossy:


CHANGES
```diff
@@ -1,5 +1,6 @@
 Tajo Change Log 
 
+
```
blrunner: It seems to be an unnecessary modification.

blrunner:
+1

LGTM! I just leaved a trivial comment. Before committing this PR, check it again. :)

PR #934, opened 2016-01-07
TAJO-2043: Implement new data type and schema.

hyunsik:


jihoonson:
Thanks. This patch looks good overall. I'll leave some comments soon.

jihoonson:
Hi @hyunsik, this patch contains a new design and interface of schema. So, to merge this patch with other parts completely, there will be quite a lot of issues. 
Would you share some plan briefly if you have?

hyunsik:
Hi @jihoonson,

It requires a bunch of works. First of all, I'll implement some proxy to keep existing implementation while working in new approach too. Then, I'll continue to change existing implementation incrementally.

jihoonson:
Thanks. It would be great if we can do together after you implement the proxy.
Here is my +1. Ship it!

PR #936, opened 2016-01-11
TAJO-2048: QueryMaster and TajoWorker should support the exception propagation.

jinossy:


tajo-core/src/main/java/org/apache/tajo/querymaster/Stage.java
```diff
@@ -1415,8 +1434,9 @@ public StageState transition(Stage stage, StageEvent stageEvent) {
             stage.abort(StageState.KILLED);
             return StageState.KILLED;
           } else {
-            LOG.error("Invalid State " + stage.getSynchronizedState() + " State");
-            stage.abort(StageState.ERROR);
+            TajoInternalError error = new TajoInternalError("Invalid State " + stage.getSynchronizedState() + " State");
```
jihoonson: I wonder when this else clause can be executed. Would you explain some details?

tajo-core/src/main/java/org/apache/tajo/master/TajoMasterClientService.java
```diff
@@ -505,6 +503,11 @@ public GetQueryStatusResponse getQueryStatus(RpcController controller, GetQueryS
               builder.setFinishTime(queryInfo.getFinishTime());
             } else {
               builder.setFinishTime(System.currentTimeMillis());
+
+              if(!StringUtils.isEmpty(queryInfo.getLastMessage())) {
+                builder.setErrorMessage(queryInfo.getLastMessage());
+              }
+              builder.setState(OK);
```
jihoonson: It seems duplicate code to set the query state.

jihoonson:
Hi @jinossy, thanks for your work. It looks good overall.
I left a couple of comments. Please consider them.

tajo-core/src/main/java/org/apache/tajo/querymaster/Stage.java
jinossy: @jihoonson 
this else clause seems to never be executed. I will remove this code block for clarify

jihoonson:
+1 ship it!

PR #940, opened 2016-01-13
TAJO-2056: Revise TPC-H benchmark test set

dkhwangbo:
Update data type of some column.


dkhwangbo:
This PR needs more modification deeply. I'll reopen this PR when I resolve all the problem about it.

PR #942, opened 2016-01-15
TAJO-1969: Replace Anonymous class with Lambda expression

dkhwangbo:
I replace All anonymous classes with lambda expressions.
Lambda syntax is not supported under Java 1.7 or earlier JVMs.


hyunsik:
could you rebase it?

dkhwangbo:
@hyunsik Thanks for your review. I rebase it.

PR #944, opened 2016-01-15
TAJO-2058: foreach loop can be collapsed with stream api

dkhwangbo:
I replace many foreach loop with stream api.

### DIFF
--- a/tajo-algebra/src/main/java/org/apache/tajo/algebra/CreateTable.java
+++ b/tajo-algebra/src/main/java/org/apache/tajo/algebra/CreateTable.java
@@ -30,6 +30,7 @@ import java.util.HashMap;
 import java.lang.reflect.Type;
 import java.util.List;
 import java.util.Map;
+import java.util.stream.Collectors;
 
 public class CreateTable extends Expr {
   @Expose @SerializedName("IsExternal")
@@ -419,9 +420,7 @@ public class CreateTable extends Expr {
       hash.quantity = quantity;
       if (specifiers != null) {
         hash.specifiers = new ArrayList<>();
-        for (PartitionSpecifier specifier : specifiers) {
-          hash.specifiers.add(specifier);
-        }
+        hash.specifiers.addAll(specifiers.stream().collect(Collectors.toList()));
       }
       return hash;
     }
@@ -472,9 +471,7 @@ public class CreateTable extends Expr {
       }
       if (specifiers != null) {
         listPartition.specifiers = new ArrayList<>();
-        for (ListPartitionSpecifier specifier : specifiers) {
-          listPartition.specifiers.add(specifier);
-        }
+        listPartition.specifiers.addAll(specifiers.stream().collect(Collectors.toList()));
       }
       return listPartition;
     }
--- a/tajo-catalog/tajo-catalog-server/src/main/java/org/apache/tajo/catalog/store/XMLCatalogSchemaManager.java
+++ b/tajo-catalog/tajo-catalog-server/src/main/java/org/apache/tajo/catalog/store/XMLCatalogSchemaManager.java
@@ -43,6 +43,7 @@ import java.sql.*;
 import java.util.*;
 import java.util.jar.JarEntry;
 import java.util.jar.JarFile;
+import java.util.stream.Collectors;
 
 public class XMLCatalogSchemaManager {
   protected final Log LOG = LogFactory.getLog(getClass());
@@ -304,12 +305,8 @@ public class XMLCatalogSchemaManager {
     
     final List<SchemaPatch> candidatePatches = new ArrayList<>();
     Statement stmt;
-    
-    for (SchemaPatch patch: this.catalogStore.getPatches()) {
-      if (currentVersion >= patch.getPriorVersion()) {
-        candidatePatches.add(patch);
-      }
-    }
+
+    candidatePatches.addAll(this.catalogStore.getPatches().stream().filter(patch -> currentVersion >= patch.getPriorVersion()).collect(Collectors.toList()));
     
     Collections.sort(candidatePatches);
     try {
--- a/tajo-core-tests/src/test/java/org/apache/tajo/client/TestTajoClient.java
+++ b/tajo-core-tests/src/test/java/org/apache/tajo/client/TestTajoClient.java
@@ -57,6 +57,7 @@ import java.io.InputStream;
 import java.sql.ResultSet;
 import java.sql.SQLException;
 import java.util.*;
+import java.util.stream.Collectors;
 
 import static org.junit.Assert.*;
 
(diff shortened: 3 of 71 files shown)


tajo-algebra/src/main/java/org/apache/tajo/algebra/CreateTable.java
```diff
@@ -419,9 +420,7 @@ public Object clone() throws CloneNotSupportedException {
       hash.quantity = quantity;
       if (specifiers != null) {
         hash.specifiers = new ArrayList<>();
-        for (PartitionSpecifier specifier : specifiers) {
-          hash.specifiers.add(specifier);
-        }
+        hash.specifiers.addAll(specifiers.stream().collect(Collectors.toList()));
```
eminency: Actually in a case like this, I think it's sufficient and more clear.

```java
hash.specifiers.addAll(specifiers);
```

or more simply,

```java
hash.specifiers = new ArrayList<>(specifiers);
```

Anyway, I think stream doesn't matter.
Because it is simple shallow copy, there are many ways.

dkhwangbo: @eminency Thanks for your comment. I agree with your opinion.
I'll rewrite it overall correlated with a simple shallow copy.

eminency:
Time to merge.

dkhwangbo:
@eminency Thanks for your comment. Unfortunately, I face with some unsuspected error while travis test. I'm in investigation about it's cause.

tajo-algebra/src/main/java/org/apache/tajo/algebra/CreateTable.java
```diff
@@ -418,10 +419,7 @@ public Object clone() throws CloneNotSupportedException {
       }
       hash.quantity = quantity;
       if (specifiers != null) {
-        hash.specifiers = new ArrayList<>();
-        for (PartitionSpecifier specifier : specifiers) {
-          hash.specifiers.add(specifier);
-        }
+        hash.specifiers = specifiers.stream().collect(Collectors.toList());
```
eminency: It is sufficient with: 

```java
hash.specifiers = new ArrayList<>(specifiers);
```

tajo-algebra/src/main/java/org/apache/tajo/algebra/CreateTable.java
```diff
@@ -471,10 +469,7 @@ public Object clone() throws CloneNotSupportedException {
         }
       }
       if (specifiers != null) {
-        listPartition.specifiers = new ArrayList<>();
-        for (ListPartitionSpecifier specifier : specifiers) {
-          listPartition.specifiers.add(specifier);
-        }
+        listPartition.specifiers = specifiers.stream().collect(Collectors.toList());
```
eminency: Same as above.

tajo-catalog/tajo-catalog-server/src/main/java/org/apache/tajo/catalog/store/XMLCatalogSchemaManager.java
```diff
@@ -602,18 +597,10 @@ protected void copySchemaInfo(StoreObject sourceStore) {
...
-      
-      for (DatabaseObject object: unorderedObjects) {
-        if (object != null) {
-          mergedObjects.add(object);
-        }
-      }
+
+      orderedObjects.stream().filter(object -> object != null).forEach(mergedObjects::add);
+
+      unorderedObjects.stream().filter(object -> object != null).forEach(mergedObjects::add);
```
eminency: You can merge two statements into one like this:

```java
Stream.concat(orderedObjects.stream(), unorderedObjects.stream())
   .filter(o -> o != null)
   .forEach(mergedObjects::add);
```

tajo-catalog/tajo-catalog-server/src/main/java/org/apache/tajo/catalog/store/XMLCatalogSchemaManager.java
```diff
@@ -644,30 +631,22 @@ protected void validatePatch(List<SchemaPatch> patches, SchemaPatch testPatch) {
     
     protected void mergePatches(List<SchemaPatch> patches) {
       final List<DatabaseObject> objects = new ArrayList<>();
```
eminency: I can't understand why this exists.
Please check if it can be removed.

tajo-catalog/tajo-catalog-server/src/main/java/org/apache/tajo/catalog/store/XMLCatalogSchemaManager.java
```diff
@@ -675,40 +654,38 @@ protected void validateSQLObject(List<SQLObject> queries, SQLObject testQuery) {
...
     protected void mergeDropStatements(List<SQLObject> queries) {
-      for (SQLObject query: queries) {
+      queries.forEach(query -> {
         validateSQLObject(queries, query);
-        
+
         targetStore.addDropStatement(query);
-      }
+      });
     }
```
eminency: I think these two methods can be removed and merged as lambda itself unless it is called from many places.

tajo-core-tests/src/test/java/org/apache/tajo/client/TestTajoClient.java
```diff
@@ -175,10 +176,8 @@ public final void testSessionVariables() throws IOException, TajoException, Inte
     String prefixName = "key_";
     String prefixValue = "val_";
 
-    List<String> unsetList = new ArrayList<>();
-    for(Map.Entry<String, String> entry: client.getAllSessionVariables().entrySet()) {
-      unsetList.add(entry.getKey());
-    }
+    List<String> unsetList = client.getAllSessionVariables().entrySet().stream()
+      .map(Map.Entry<String, String>::getKey).collect(Collectors.toList());
```
eminency: I think next one is more readable:

```java
 List<String> unsetList = new ArrayList<>(client.getAllSessionVariables().keySet());
```

tajo-core/src/main/java/org/apache/tajo/engine/query/TaskRequestImpl.java
```diff
@@ -216,49 +217,42 @@ public Enforcer getEnforcer() {
...
+    if (this.fetches != null) {
       return;
     }
     TaskRequestProtoOrBuilder p = viaProto ? proto : builder;
-    this.fetches = new ArrayList<>();
-    for(FetchProto fetch : p.getFetchesList()) {
-      fetches.add(fetch);
-    }
-	}
+    this.fetches = p.getFetchesList().stream().collect(Collectors.toList());
```
eminency: Same as above early.

tajo-core/src/main/java/org/apache/tajo/querymaster/Query.java
```diff
@@ -303,12 +304,7 @@ public QueryHistory getQueryHistory() {
   }
 
   private List<StageHistory> makeStageHistories() {
-    List<StageHistory> stageHistories = new ArrayList<>();
-    for(Stage eachStage : getStages()) {
-      stageHistories.add(eachStage.getStageHistory());
-    }
-
-    return stageHistories;
+    return getStages().stream().map(Stage::getStageHistory).collect(Collectors.toList());
```
eminency: It can be put into caller directly.

eminency:
Even though they may not exact, I left some comments.
Please check them out.

dkhwangbo:
@eminency Hi! Thanks for your comment. I apply your comment.

PR #945, opened 2016-01-15
TAJO-2059: Binary search in BST reader does compare too frequently

eminency:
Some simple test based on unit test was done with 1M tuples and searching 1M times.
Two columns, which are long and double, are used as sort key.
Count above is a number of invoking compare().

##### Current

find : 44 sec
count : 42001207

##### Patch

find : 43 sec
count : 20141495

### DIFF
--- a/tajo-storage/tajo-storage-hdfs/src/main/java/org/apache/tajo/storage/index/bst/BSTIndex.java
+++ b/tajo-storage/tajo-storage-hdfs/src/main/java/org/apache/tajo/storage/index/bst/BSTIndex.java
@@ -794,6 +794,7 @@ public class BSTIndex implements IndexMethod {
       int offset = -1;
       int start = startPos;
       int end = endPos;
+      int prevCenter = -1;
 
       //http://bugs.sun.com/bugdatabase/view_bug.do?bug_id=6412541
       int centerPos = (start + end) >>> 1;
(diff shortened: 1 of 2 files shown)


jihoonson:
Overall, this patch looks good to me. Would you check the test failures?

eminency:
@jihoonson 
Thank you for the comment.
I'm working on it. I will keep you posted.

tajo-storage/tajo-storage-hdfs/src/main/java/org/apache/tajo/storage/index/bst/BSTIndex.java
```diff
@@ -794,42 +794,79 @@ private int binarySearch(Tuple[] arr, Tuple key, int startPos, int endPos) {
...
+        } else {
+          prevCenter = centerPos;
+
+          if (compareResult > 0) {
+            end = centerPos;
+          } else {
+            start = centerPos;
+          }
+
+          centerPos = (start + end) / 2;
```
jihoonson: This line also looks to be changed to ```centerPos = (start + end) >>> 1;```

tajo-storage/tajo-storage-hdfs/src/main/java/org/apache/tajo/storage/index/bst/BSTIndex.java
```diff
@@ -794,42 +794,79 @@ private int binarySearch(Tuple[] arr, Tuple key, int startPos, int endPos) {
...
-          }
-        } else if (comparator.compare(arr[centerPos], key) < 0) {
-          if (centerPos == arr.length - 1) {
-            correctable = false;
-            offset = centerPos;
-            break;
-          } else if (comparator.compare(arr[centerPos + 1], key) > 0) {
-            correctable = false;
-            offset = centerPos;
+            if (end == arr.length) {
```
jihoonson: In this method ```binarySearch()```, ```endPos``` is intended as an exclusive end position. It looks that codes in this block need to consider this intention.

PR #946, opened 2016-01-19
TAJO-2060: Upgrade geoip-api-java library

combineads:
Tajo use the dspace-geoip library, version is 1.2.3.
Now geoip data doesn't work at 1.2.3 version.
We need change 1.2.14 from 1.2.3 version.


tajo-core-tests/src/test/java/org/apache/tajo/util/TestGeoIPUtil.java
```diff
@@ -0,0 +1,61 @@
...
+    File tmpFile = File.createTempFile("GeoIP", ".dat");
+
+    try (InputStream is = new URL("http://www.maxmind.com/download/geoip/database/GeoLiteCountry/GeoIP.dat.gz")
+        .openConnection().getInputStream();
+        OutputStream out = new FileOutputStream(tmpFile)) {
+      IOUtils.copy(new GZIPInputStream(is), out);
+    }
+    return tmpFile;
+  }
+}
```
jinossy: In my opinion, we don’t need to tests of third-party module.

tajo-core/pom.xml
```diff
@@ -262,9 +262,9 @@
       <artifactId>commons-codec</artifactId>
     </dependency>
     <dependency>
-      <groupId>org.dspace.dependencies</groupId>
-      <artifactId>dspace-geoip</artifactId>
-      <version>1.2.3</version>
+      <groupId>com.maxmind.geoip</groupId>
+      <artifactId>geoip-api</artifactId>
+      <version>1.2.14</version>
```
jinossy: Could you change to latest version? and this dependency scope should be changed to “provided”.

jinossy:
@combineads 
Thanks for your contribution.
I left some comments

combineads:
I’ve update the patch that reflects your suggestions. @jinossy 
Thank you for comment.

jinossy:
+1 LGTM!
Test failure is not related this PR

PR #950, opened 2016-01-28
TAJO-2065: Increase default heap size.

jinossy:
I've change the default value to following:
`Default Xmx: 1G -> 5G`
`tajo.task.resource.min.memory-mb: 500 -> 1000`
`tajo.worker.resource.disks: 1 -> 2`
`Default GC: -XX:+UseParallelOldGC`
This configuration can execute 4 task simultaneously.


jihoonson:
+1. This patch doesn't relate to the unit test.
Ship it!

PR #954, opened 2016-02-01
TAJO-2070: Remove unnecessarily version property in maven pom.

jinossy:


blrunner:
+1

LGTM. Ship it! :)

PR #955, opened 2016-02-01
TAJO-2064: Supporting auto-completion in Tsql

eminency:
Use tab key in Tsql


blrunner:
Sorry for late review. I'll finish to review this PR in a few days.

eminency:
Never mind.
It is not a critical patch.

blrunner:
@eminency 

If you implement ``Completer`` to ``DescFunctionCommand`` that shows built-in function description, this patch will be more useful.

eminency:
@blrunner 

Good, I've done.

tajo-cli/src/main/java/org/apache/tajo/cli/tsql/TajoCli.java
```diff
@@ -374,11 +387,43 @@ private void initCommands() {
...
+            char c = buf.charAt(pos);
+            return Character.isWhitespace(c) || !(Character.isLetterOrDigit(c)) && c != '_';
+          }
+        },
+        new StringsCompleter(getKeywords())
+    );
+  }
+
+  private Collection<String> getKeywords() {
+    List<String> klist = new ArrayList<>();
```
blrunner: It looks like a ambiguous name. How about rename it to ``keywords`` or ``keywordList``?

tajo-cli/src/main/java/org/apache/tajo/cli/tsql/commands/TajoShellCommand.java
```diff
@@ -126,4 +136,56 @@ protected void print(char c, int count) {
     println();
     return columnWidths;
   }
+
+  public ArgumentCompleter getArgumentComplementer() {
```
blrunner: It seems to be a typo.

eminency: Good catch

tajo-cli/src/main/java/org/apache/tajo/cli/tsql/commands/TajoGetConfCommand.java
```diff
@@ -55,4 +62,52 @@ public String getUsage() {
...
+    }
+
+    return new ArgumentCompleter(
+        new StringsCompleter(getCommand()),
+        new ConfCompleter(confNames.toArray(new String[confNames.size()])),
+        NullCompleter.INSTANCE);
+  }
+}
+
+class ConfCompleter extends StringsCompleter {
```
blrunner: Currently, many tajo inner-classes have been declared inside their top-class. How about put ``ConfCompleter`` inside ``TajoGetConfCommand``?

eminency: I think it is a good suggestion.

eminency:
@blrunner
Ok, please check what you commented.

blrunner:
@eminency 

Thanks for your reflection.

Well, I have one more suggestion. Currently, Tajo provides a SQL grammar file as following:

https://github.com/apache/incubator-tajo/blob/master/tajo-core/tajo-core-backend/src/main/antlr4/org/apache/tajo/engine/parser/SQLLexer.g4

But if we have existing grammar file and ``SQLKeywords``, when other contributors add new grammar, they must add it to above two files. I think that it is a little inefficient for maintaing codes. Even though it may be trouble, if you use ``SQLLexer.g4``, this patch will be better. For reference, there are some unnecessary keywords to ``SQLLexter.g4`` on tsql. So you need to filter theme at ``TajoCli::getKeywords``.

eminency:
@blrunner 
SQL keywords are fetched from SQLLexer now.

blrunner:
+1

Sorry for late review.
I found that the auto-completion ran as expected and your reflection about my opinion.

PR #961, opened 2016-02-12
TAJO-2074: Upgrade hadoop and netty.

jinossy:


jihoonson:
@jinossy, thanks for your work. This patch looks nice to me, but how about separating it into two small patches for CI testing and upgrading Hadoop version?

jinossy:
OK I’ll separate the travis.yml

jihoonson:
+1 LGTM!

PR #962, opened 2016-02-15
TAJO-2076: Add backward compatibility testing with hadoop on travis.

jinossy:


jihoonson:
+1. It looks good to disable HA test before we fix it.

PR #963, opened 2016-02-17
TAJO-2077: Join condition causes incorrect result, when a table has an empty row file.

jinossy:


jinossy:
Test failure is not related this PR.
refer to https://issues.apache.org/jira/browse/TAJO-2078

jihoonson:
+1 LGTM!

PR #965, opened 2016-02-18
TAJO-1793: result row count unmatched for UNION ALL

blrunner:


blrunner:
Removed codes for checking ``LogicalRootNode`` and add a flag for separating union all query to ``ExecutionBlock``.

blrunner:
Remove unit test in ``TajoCli`` and add unit tests by ``TableStats``.

jihoonson:
In newly added tests, the same queries of our existing tests on union are executed, which means the queries of our union tests are executed twice for query result validation and result stat validation. I think it will be better to verify both query result and result stat together.

It is simple if you get the query id of executed test query. You can get the query id from ResultSet which is actually the TajoResultSetBase. Once you get the query id, you can get TableDesc of the query result which includes the stat information. Here is an example.

```
ResultSet res = executeQuery();
QueryId qid = ((TajoResultSetBase)res).getQueryId();
QueryInfo queryInfo = testingCluster.getMaster().getContext().getQueryJobManager().getFinishedQuery(qid);
TableDesc desc = queryInfo.getResultDesc();
System.out.println(desc.getStats().getNumRows());
```

blrunner:
Thanks @jihoonson ,

That's a good point. I reflected your comments. :-)

tajo-core-tests/src/test/java/org/apache/tajo/engine/query/TestUnionQuery.java
```diff
@@ -697,4 +721,14 @@ public void testComplexUnion2() throws Exception {
   public void testUnionAndFilter() throws Exception {
     runSimpleTests();
```
jihoonson: Simple test is also possible to verify result stat. See  #971.

blrunner: I applied ``QueryTestCaseBase::runSimpleTests`` with your opinion.

blrunner:
I got to know that ``Query`` must calculates result stats of child stages. So I removed existing related codes and implemented codes for computing stats to ``Query``. And I applied verification for result stats in ``TestUnionQuery``.

tajo-core/src/main/java/org/apache/tajo/querymaster/Query.java
```diff
@@ -402,6 +403,39 @@ public QueryState getSynchronizedState() {
     }
   }
 
+  public boolean hasUnionPlan() throws Exception {
+    boolean result = false;
+
+    // During executing UNION statement or UNION ALL statement, the terminal block doesn't have join plan and has
+    // two more child blocks. In above case, Tajo should summarize result TableStats from all child stages.
+    ExecutionBlock terminalBlock = plan.getTerminalBlock();
+    terminalBlock.setPlan(terminalBlock.getPlan());
```
jihoonson: The terminal block doesn't have any plan, so this line doesn't affect to anything. I wonder what is your intention.

tajo-core/src/main/java/org/apache/tajo/querymaster/Query.java
```diff
@@ -402,6 +403,39 @@ public QueryState getSynchronizedState() {
...
+    if (plan.getChilds(terminalBlock.getId()).size() >= 2) {
+      result = true;
+    }
+    return result;
+  }
+
+  public TableStats computeChildResultStats() throws Exception {
```
jihoonson: A better implementation exists at Stage.computeStatFromUnionBlock(). It would be better to move this method to StatisticsUtil.

tajo-core/src/main/java/org/apache/tajo/querymaster/Query.java
blrunner: I agree with you and reflected your comments.

tajo-core/src/main/java/org/apache/tajo/querymaster/Query.java
blrunner: Moving this method to ``StatisticsUtil`` throws compile error as following:
```
[ERROR] The projects in the reactor contain a cyclic reference: Edge between 'Vertex{label='org.apache.tajo:tajo-catalog-common:0.12.0-SNAPSHOT'}' and 'Vertex{label='org.apache.tajo:tajo-core:0.12.0-SNAPSHOT'}' introduces to cycle in the graph org.apache.tajo:tajo-core:0.12.0-SNAPSHOT --> org.apache.tajo:tajo-catalog-common:0.12.0-SNAPSHOT --> org.apache.tajo:tajo-core:0.12.0-SNAPSHOT -> [Help 1]
```

It seems that because there is a cyclic reference between tajo-catalog-common module and tajo-core module. So I improved this method using ``StatisticsUtil.aggregateTableStat`` instead of moving to ``StatisticsUtil``.

tajo-core/src/main/java/org/apache/tajo/querymaster/Query.java
```diff
@@ -653,12 +679,19 @@ public void execute(QueryMaster.QueryMasterContext context, QueryContext queryCo
...
+              lastStage.getOutSchema(),
                 meta,
                 finalOutputDir.toUri());
         resultTableDesc.setExternal(true);
 
-        stats.setNumBytes(getTableVolume(query.systemConf, finalOutputDir));
+        if (query.hasUnionPlan()) {
+          TableStats aggregated = query.aggregateTableStatsOfTerminalBlock();
+          stats.setNumRows(aggregated.getNumRows());
+          stats.setNumBytes(aggregated.getNumBytes());
```
jihoonson: Copying numRows and numBytes is insufficient. ```aggretated``` should be set as the stats of ```resultTableDesc```.

tajo-core-tests/src/test/java/org/apache/tajo/engine/query/TestUnionQuery.java
```diff
@@ -43,658 +49,760 @@ public TestUnionQuery() {
...
-    cleanupQuery(res);
+    Optional<TajoResultSetBase[]> existing = runSimpleTests();
+    assertTrue(existing.isPresent());
+    TajoResultSetBase[] resultSet = existing.get();
+    QueryId qid = resultSet[0].getQueryId();
+    QueryInfo queryInfo = testingCluster.getMaster().getContext().getQueryJobManager().getFinishedQuery(qid);
+    TableDesc desc = queryInfo.getResultDesc();
+    TableStats stats = desc.getStats();
+    assertEquals(8L, stats.getNumRows().longValue());
+    assertEquals(96L, stats.getNumBytes().longValue());
```
jihoonson: resultSets should be closed when a test is finished. Use QueryTestCaseBase.closeResultSets().

tajo-core-tests/src/test/java/org/apache/tajo/engine/query/TestUnionQuery.java
```diff
@@ -43,658 +49,760 @@ public TestUnionQuery() {
...
   /**
    * S (SA U SA) O
    */
+  @Test
+  @SimpleTest
   public final void testUnionAll1() throws Exception {
-    ResultSet res = executeQuery();
-    assertResultSet(res);
-    cleanupQuery(res);
+    Optional<TajoResultSetBase[]> existing = runSimpleTests();
```
jihoonson: How about making the below testing code as a separate method to reuse in every test of this class?

blrunner:
@jihoonson 

Thank your for your detailed review. I reflected your comments.

jihoonson:
+1. Please don't forget to change Java's Optional to Google's one when you commit to 0.11.2 branch.

PR #969, opened 2016-03-01
TAJO-2082: Aggregation on a derived table which includes union can cause incorrect result

jihoonson:
I didn't add unit test because it is difficult to reproduce this bug with unit test. 
IMO, it is better to add a verifier to test global plan. I created a jira ticket for global plan verifier. (https://issues.apache.org/jira/browse/TAJO-2084)

### DIFF
--- a/tajo-core/src/main/java/org/apache/tajo/engine/planner/global/MasterPlan.java
+++ b/tajo-core/src/main/java/org/apache/tajo/engine/planner/global/MasterPlan.java
@@ -51,6 +51,9 @@ public class MasterPlan {
 
   private Map<ExecutionBlockId, ShuffleContext> shuffleInfo = new HashMap<>();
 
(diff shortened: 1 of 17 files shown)


tajo-core/src/main/java/org/apache/tajo/engine/planner/global/MasterPlan.java
```diff
@@ -23,8 +23,10 @@
 
 import org.apache.tajo.ExecutionBlockId;
 import org.apache.tajo.QueryId;
+import org.apache.tajo.annotation.NotNull;
```
hyunsik: unused import

hyunsik:
+1 LGTM! The change looks straightforward.

jihoonson:
Thank you for the review.

PR #970, opened 2016-03-02
[TAJO-2085] Add parseDateTime method with TimeZone in DateTimeFormat class

combineads:
I added the UDF for my application like below.

TimeMeta tm = DateTimeFormat.parseDateTime(pastDateText, "YYYYMMDD");
DateTimeUtil.toUserTimezone(tm, timezone);

I expected to return the date to apply a time zone but It wasn't working.
So I hope to add parseDateTime method with TimeZone.
Thanks.


combineads:
It was to resolve useing DateTimeUtil.toUTCTimezone(tm, tz);
Thanks a lot.

PR #972, opened 2016-03-07
TAJO-2089: Improve null handling in UDF.

jinossy:


blrunner:
+1

LGTM! I found that ResultSet of TajoClient using some test queries had retrieved null values.

PR #973, opened 2016-03-08
TAJO-2090: Bump up orc version to 0.141.

hyunsik:


eminency:
+1.
I think it is absolutely clear :)

jihoonson:
+1

PR #974, opened 2016-03-09
TAJO-1929: Can't read ORC file on hive-meta

eminency:
Creating table and reading ORC files are tested in Tajo with Hive catalog.
And, ORC files created from Tajo are read well on Hive.


hyunsik:
+1 Ship it.

hyunsik:
I tested this patch. This patch works when we create a table. But, it does not work when an user get a table. Could you check it?

tajo-dist/pom.xml
```diff
@@ -154,9 +154,6 @@
                       run cp -r ${project.basedir}/src/main/conf .
                       run rm -rf lib/tajo-*-${project.version}.jar
 
-                      run mkdir hive
```
hyunsik: It should be kept because org.apache.hadoop.hive.ql.metadata.Table::getSd() included in hive-apache-0.9 causes NoSuchMethod exception.

jinossy:
Hi guys,
This issue is solved by https://github.com/apache/tajo/pull/985

PR #978, opened 2016-03-16
TAJO-2096: Apply offheap to the join operator.

jinossy:


jinossy:
This patch is not ready for review yet. I need to improve the projector

PR #979, opened 2016-03-16
TAJO-2087: Implement DirectOutputCommitter

blrunner:
Here is prototype codes for ``DirectOutputCommitter``. This PR is not ready to review, it shows my approach to implement ``DirectOutputCommitter``. Current version works as following:

- Register commit history to catalog (TODO).
- Each tasks will write the output data directly to the final location.
- In a commit phase, delete existing files with query type as follows. First, backup existing files or directories to staging directory. And then delete backup files or directories. 
- Update the status of commit history to catalog (TODO).
- If query fails, QueryMaster will delete committed files and update the status of query history to catalog (TODO).
- When ``TajoMaster`` starting, it will check the status of query histories to catalog. If it find running query, it will delete committed files and update the status of query history (TODO).
- Add unit test cases for failed query (TODO).


blrunner:
I designed the table of direct output commit history as following:

Column Name | Column Type | Null | Desc
-------------------|----------------------|--------------------------|-------------------
QUERY_ID | VARCHAR(128) | NOT NULL | the id of ``Query``, PRIMARY KEY
PATH | VARCHAR(4096) | NOT NULL  | the output path of table
START_TIME | BIGINT | NOT NULL | query start time
END_TIME |  BIGINT | | query finish time
QUERY_STATE | VARCHAR(50) | NOT NULL | the state of ``Query``, ``TajoProtos.QueryState`` will be used

Implemented necessary codes to ``Query`` and ``TajoMaster`` as following:
* When query starting, add the history to catalog.
* If query fails, ``QueryMaster`` will delete committed files and update the status of output commit history to catalog
* When ``TajoMaster`` starting, it will check the status of output commit histories to catalog. If it find running query, it will delete committed files and update the status of output commit history.

Not yet implemented unit test cases for failed queries.

blrunner:
I implemented unit test cases for verifying following cases.
* ``DirectOutputCommitter`` can recover existing files successfully in query failure case. 
*  ``DirectOutputCommitter`` can remove output files successfully in query failure case.
* When executing ``INSERT INTO`` query, ``DirectOutputCommitter`` can maintain existing files.

For the reference, I found that outputs of ``TestInsertQuery`` and ``TestTablePartitions`` with ``DirectOutputCommitter`` were equals to outputs of them without ``DirectOutputCommitter``.

blrunner:
Added a document for DirectOutputCommitter as follows.
https://github.com/blrunner/tajo/blob/direct-output-committer/tajo-docs/src/main/sphinx/table_management/directoutputcommitter.rst

blrunner:
Testing with S3 finished successfully as following:

* Table type: partitioned table and non-partitioned table
* Insert overwrite and CTAS queries
* Insert into a table without DirectOutputCommitter, and then insert into the table with DirectOutputCommitter, and then check output files for above two query.
* While inserting data, kill the query, and then check two factors: delete temporary data and rollback previous data.
* While inserting data, restart tajo cluster, and then check two factors: delete temporary data and rollback previous data.

PR #980, opened 2016-03-16
TAJO-2097: TajoStatement::getMoreResults throws SQLFeatureNotSupportedException

blrunner:


blrunner:
Current approach looks like a temporary solution.

PR #984, opened 2016-03-18
TAJO-1891: Allow arbitrary partition path

blrunner:
This includes features as following:

* Allow to use `` ALTER TABLE ADD PARTITION`` statement.
* Allow to use `` ALTER TABLE DROP PARTITION`` statement.
* When creating dump file to restore Catalog, use ``ALTER TABLE ADD PARTITION`` statement instead of ``ALTER TABLE REPAIR PARTITION`` statement.
* Add unit test cases.

It depends on #846 .


blrunner:
I'll reopen this PR after finishing https://github.com/apache/tajo/pull/1020.

PR #987, opened 2016-03-28
TAJO-2013: JdbcMetadataProviderBase need to differentiate between database name in DBMS and database name in Tajo.

blrunner:
review_button


blrunner:
Mismatched issue number.

PR #988, opened 2016-03-28
TAJO-2104: Implement Identifier which supports quotation information.

hyunsik:


tajo-common/src/main/java/org/apache/tajo/schema/QualifiedIdentifier.java
```diff
@@ -0,0 +1,58 @@
...
+
+public class QualifiedIdentifier {
+  private List<Identifier> names;
+
+  private QualifiedIdentifier(ImmutableList<Identifier> names) {
+    this.names = names;
+  }
+
+  public String displayString(final IdentifierPolicy policy) {
+    return StringUtils.join(names, ".", new Function<Identifier, String>() {
```
jihoonson: Looks to be IdentifierPolicy.getIdentifierSeperator() instead of ```"."```.

jihoonson:
+1 LGTM! 
Thanks for quick update.

hyunsik:
Thank you for the quick review. I'll commit it soon.

PR #990, opened 2016-03-30
TAJO-2103: JdbcMetadataProviderBase need to differentiate between database name in DBMS and database name in Tajo.

blrunner:


hyunsik:
Could you explain the purpose of this patch? 

See setDatabase() in the below. The variable 'database' already contains 'mapped_dbname'.
```
public JdbcTablespace(String name, URI uri, JSONObject config) {
    super(name, uri, config);
    setDatabase();
    setJdbcProperties();
  }

  private void setDatabase() {
    if (config.containsKey(CONFIG_KEY_MAPPED_DATABASE)) {
      database = this.config.getAsString(CONFIG_KEY_MAPPED_DATABASE);
    } else {
      database = ConnectionInfo.fromURI(uri).database();
    }
  }
```

But, your patch changes as the below. It's a duplicate.
```
public MetadataProvider getMetadataProvider() {
   return new PgSQLMetadataProvider(this, database, ConnectionInfo.fromURI(uri).database());
}
```

PR #995, opened 2016-04-06
TAJO-2112: Improve disk load, when queries run simultaneously.

jinossy:
I've remove the ``tajo.worker.resource.disks`` and ``tajo.worker.resource.dfs-dir-aware``


jihoonson:
+1 LGTM!

PR #996, opened 2016-04-08
TAJO-2115: Change the size of the VALUE_ column in OPTIONS table

blrunner:
This patch just updates the ``VALUE_`` column size of ``OPTIONS`` table. Existing codes of catalog module have remained static. Here are corrections as following:

DBMS | Existing Type | Updated Type 
-------------------|----------------------|--------------------------
Derby | VARCHAR(4000) | LONG VARCHAR  
MySQL | VARCHAR(4000) | TEXT
MariaDB | VARCHAR(4000) | TEXT 
PostgreSQL | VARCHAR(4000) | TEXT
Oracle | VARCHAR2(4000) | LONG
 
I referenced following sites.
* https://db.apache.org/derby/docs/10.1/ref/rrefsqlj15147.html
* https://docs.oracle.com/cd/B19306_01/java.102/b14188/datamap.htm
* https://en.wikibooks.org/wiki/Converting_MySQL_to_PostgreSQL


hyunsik:
Could you rebase it?

jihoonson:
This is a temporal solution. We need to investigate a better way.

blrunner:
Rebased.

hyunsik:
@jihoonson In this issue or in another issue?

@blrunner It seems to need rebase again.

jihoonson:
I mean, this issue is not addressing the fundamental problem. I believe that we need to devise a better way to map Tajo columns and HBase columns.

blrunner:
This patch is a temporary solution. We need to solve the fundamental problem about this issue.

PR #998, opened 2016-04-15
TAJO-2119: Invalid sort result when sort key columns contain non-ascii values

jihoonson:


jinossy:
+1 LGTM!

jihoonson:
Thanks for quick review!

PR #1000, opened 2016-04-15
TAJO-2108: Refactor Schema to be immutable.

hyunsik:


jinossy:
Thanks for your effort!

PR #1002, opened 2016-04-19
TAJO-2125:  Detect JAVA_HOME on OSX , if JAVA_HOME is not set.

jinossy:


jihoonson:
+1 LGTM.

hyunsik:
There is some potential problem. The original implement forces users to specific JAVA_HOME. As a result, users must set JAVA_HOME.

This change allows users to not specific JAVA_HOME. If users do not set JAVA_HOME, Tajo can run on different JVMs across multiple nodes. It may cause critical problem because we are using unsafe codes in some parts.

jinossy:
@hyunsik 
Now that master branch is based on JDK 1.8. if someone try to run the tajo on other JDK (1.7), tajo are  throw ‘Unsupported major.minor version 52.0’
If you’ve still worried about that, I will remove the finding java binary.

hyunsik:
JDK can be easily varied in many environments. For example, some machines in a cluster can use SUN JDK, and other machines can use open JDK or others. Also, JDK 9 will be released in this year. This problem is about the trade-off relationship between convenient and less mistake. If I should catch only one of them, I'd like to choose less mistake.

jihoonson:
Hi @hyunsik, thank you for your comment.
I'm wondering what kinds of unpredictable results you are concerned with. AFAIK, open JDK also supports unsafe APIs. Also, JDK 9 seems to support unsafe APIs as well (http://openjdk.java.net/jeps/260).

jinossy:
@hyunsik 
I agree

@jihoonson 
I don’t know clearly but less mistake are better

jinossy:
I've update the patch that borrow from hadoop

hyunsik:
+1 LGTM!

PR #1003, opened 2016-04-19
TAJO-2126: Allow parallel execution of non-leaf sibling ExecutionBlocks

jihoonson:


hyunsik:
The change looks straightforward and reasonable. But, I'm concerned with potential side effects. Did you do enough tests in cluster environments?

jihoonson:
I did basic tests to verify query results, but there is one thing I'm concerned with. That is the shuffle cost. This patch enables the parallel execution of non-leaf EBs, thereby more shuffle operations can be executed simultaneously. It can increase the shuffle cost especially for range shuffle due to heavy index search.
So, we need to improve hash shuffle by reusing http connections and range shuffle by redesigning the index structure.

However, I think the heavy shuffle costs should be addressed in other Jiras. What do you think?

hyunsik:
Do you mean more shuffle cost caused by heavy load?

jihoonson:
It is possible in single query execution. If multiple queries are executed, the shuffle cost of the current implementation will also be huge.

hyunsik:
Yep, I agree. I think that this patch will enhance the machine resource utilization instead of causing more shuffle cost.

+1 LGTM

jihoonson:
Thanks for your review!

PR #1005, opened 2016-04-21
TAJO-2127: Use Type instead of DataType for EvalNode.

hyunsik:


tajo-plan/src/main/java/org/apache/tajo/plan/LogicalPlanner.java
```diff
@@ -577,7 +577,7 @@ public static void verifyProjectedFields(QueryBlock block, Projectable projectab
   public static void prohibitNestedRecordProjection(Projectable projectable)
       throws TajoException {
     for (Target t : projectable.getTargets()) {
-      if (t.getEvalTree().getValueType().getType() == TajoDataTypes.Type.RECORD) {
+      if (t.getEvalTree().getValueType().isStruct()) {
```
jihoonson: Struct or record?

tajo-core-tests/src/test/java/org/apache/tajo/engine/eval/TestSQLExpression.java
```diff
@@ -565,7 +565,7 @@ public void testImplicitCastForInt8() throws TajoException {
   public void testImplicitCastForFloat4() throws TajoException {
     Schema schema = TestSchema1;
 
-    testEval(schema, "table1", "0,1,2,3,4.1,5.1,6,7", "select col4 + col0 from table1;", new String [] {"4.1"});
+    /*testEval(schema, "table1", "0,1,2,3,4.1,5.1,6,7", "select col4 + col0 from table1;", new String [] {"4.1"});
```
jihoonson: Why did you disable these tests?

hyunsik: I forget to remove the comments after some bug fixes.

tajo-plan/src/main/java/org/apache/tajo/plan/LogicalPlanner.java
hyunsik: Currently, record is right. I need to make them consistent. According to my future plan, struct would be better. That's why they are mixed.

jihoonson: Ok. Looks good.

hyunsik:
Fixed unit tests. Thank you for your quick review.

jihoonson:
+1 LGTM!

PR #1007, opened 2016-04-25
TAJO-2130: Remove inet4 data type

eminency:


tajo-core/src/main/java/org/apache/tajo/engine/function/geoip/GeoIPCountryInet4.java
```diff
@@ -1,54 +0,0 @@
...
-
-@Description(
-    functionName = "geoip_country_code",
-    description = "Convert an ipv4 address to a geoip country code.",
-    example = "> SELECT geoip_country_code(8.8.8.8);\n"
-        + "US",
-    returnType = TajoDataTypes.Type.TEXT,
-    paramTypes = {@ParamTypes(paramTypes = {Type.INET4})}
-)
-public class GeoIPCountryInet4 extends GeneralFunction {
```
jihoonson: I think these GeoIP function series are useful functions. It would be better to keep them by taking an INT4 type instead of INET4 type.

eminency: Good suggestion, however IMO, it is a separated issue from removing the type.
What do you think about that?

jihoonson: I agree. It looks good to proceed this as an umbrella issue.
Would you create an umbrella ticket and some sub tasks on Jira?

hyunsik:
It seems to need to rebase.

eminency:
@hyunsik , master branch is merged.

@jihoonson , I created the issued you mentioned, TAJO-2155 and TAJO-2156.

jihoonson:
+1 LGTM.

PR #1009, opened 2016-05-01
TAJO-2135: Invalid join result when join key columns contain nulls

jihoonson:
First of all, apologize for a large patch.
I added null values to every TPC-H tables in our test, and it makes a lot of changes in tests. However, fortunately, I could find several critical bugs with null values.
Here are the changes of this patch.

* Fixed invalid join result problem with null values.
 * When executing hash join, null values of the right table are filtered out while constructing a hash table.
* Fixed invalid cast exception in outer joins with null values.
* Fixed the problem of that sort-merge algorithm for full outer join is not executed in tests.
* Fixed wrong result of full outer hash join with null values.
* Fixed wrong result of RightOuterMergeJoinExec.
* Fixed wrong query result on partitioned tables when partition keys contain nulls.
* Changed the default null character of text file as '\N'.
* Added null values to every TPC-H tables.

I also found two additional bugs in anti join and window function executions with nulls. I think we need to fix these problems in other jiras.


jihoonson:
I also added some changes to fix window function test failures because the bug fix is really simple.

tajo-storage/tajo-storage-common/src/main/java/org/apache/tajo/storage/fragment/FragmentConvertor.java
```diff
@@ -47,8 +47,7 @@
    */
   private static final Class<?>[] DEFAULT_FRAGMENT_PARAMS = { ByteString.class };
 
-  public static Class<? extends Fragment> getFragmentClass(Configuration conf, String dataFormat)
-  throws IOException {
+  public static Class<? extends Fragment> getFragmentClass(Configuration conf, String dataFormat) {
```
blrunner: Could you explain why you remove throwing IOException and use TajoInternalError instead of IOException?

tajo-core/src/main/java/org/apache/tajo/engine/function/window/Rank.java
```diff
@@ -48,7 +48,8 @@ public Rank() {
 
   public static boolean checkIfDistinctValue(RankContext context, Tuple params) {
     for (int i = 0; i < context.latest.length; i++) {
-      if (!context.latest[i].equalsTo(params.asDatum(i)).isTrue()) {
+      if ((context.latest[i].isNotNull() || params.asDatum(i).isNotNull())
```
blrunner: Could you explain why you use ``or`` condition instead of ``and `` condition?

tajo-core/src/main/java/org/apache/tajo/engine/planner/physical/CommonJoinExec.java
```diff
@@ -145,7 +145,7 @@ public JoinNode getPlan() {
    * @return True if an input tuple is matched to the left join filter
    */
   protected boolean leftFiltered(Tuple left) {
-    return leftJoinFilter != null && !leftJoinFilter.eval(left).asBool();
+    return leftJoinFilter != null && !leftJoinFilter.eval(left).isTrue();
```
blrunner: Is the usage of ``isTrue`` method  related to modification of ``EmptyTuple::getBool``?

tajo-core/src/main/java/org/apache/tajo/engine/planner/physical/DistinctGroupbySecondAggregationExec.java
```diff
@@ -212,6 +212,7 @@ public Tuple next() throws IOException {
 
         prevKeyTuple = getKeyTuple(prevKeyTupleMap, keyTuple.getValues());
         prevTuple.put(tuple.getValues());
+        prevTuple.put(tuple.getValues());
```
blrunner: Why you use same codes again?

tajo-core/src/main/java/org/apache/tajo/engine/planner/physical/HashJoinExec.java
```diff
@@ -27,9 +28,12 @@
 
 public class HashJoinExec extends CommonHashJoinExec<TupleList> {
 
+  private final boolean isCrossJoin;
```
blrunner: The variable for checking cross join of ``CommonHashJoinExec`` have been deleted in this PR. How about add it to ``CommonHashJoinExec`` again?

tajo-storage/tajo-storage-common/src/main/java/org/apache/tajo/storage/fragment/FragmentConvertor.java
jihoonson: This function just finds and returns a proper implementation of the fragment interface. Why do we need to throw IOException?

tajo-core/src/main/java/org/apache/tajo/engine/function/window/Rank.java
jihoonson: Because, in rank(), null values are regarded as the same unlike our other implementations. Please refer to https://en.wikipedia.org/wiki/Three-valued_logic.

tajo-core/src/main/java/org/apache/tajo/engine/planner/physical/CommonJoinExec.java
jihoonson: No, asBool() incurs an invalid cast exception. Please refer to https://en.wikipedia.org/wiki/Three-valued_logic.

tajo-core/src/main/java/org/apache/tajo/engine/planner/physical/DistinctGroupbySecondAggregationExec.java
jihoonson: My fault.

tajo-core/src/main/java/org/apache/tajo/engine/planner/physical/HashJoinExec.java
jihoonson: This value is used only in HashJoinExec. We don't have to move it.

tajo-storage/tajo-storage-common/src/main/java/org/apache/tajo/storage/fragment/FragmentConvertor.java
blrunner: You're right. We don't need to throw an IOException in this method.

tajo-core-tests/src/test/java/org/apache/tajo/engine/query/TestAlterTable.java
```diff
@@ -159,12 +160,15 @@ public final void testAlterTableRepairPartition() throws Exception {
       "36.0,N,1,1\n" +
       "38.0,N,2,2\n" +
       "45.0,R,3,2\n" +
-      "49.0,R,3,3\n";
+      "49.0,R,3,3\n" +
+      "null,,null,null\n" +
+        "null,,null,null\n" +
```
blrunner: If you use the same indentation of previous codes, it will be more better. Other added codes in this class need to be updated.

blrunner:
@jihoonson 

Thanks for your contribution, this PR looks good overall.
After finishing testing on my laptop, I'll leave comments.

tajo-core-tests/src/test/java/org/apache/tajo/engine/query/TestAlterTable.java
jihoonson: Thanks. Fixed.

tajo-cluster-tests/src/test/java/org/apache/tajo/TpchTestBase.java
```diff
@@ -92,6 +92,7 @@ private void setUp() throws Exception {
     util = new LocalTajoTestingUtility();
     KeyValueSet opt = new KeyValueSet();
     opt.set(StorageConstants.TEXT_DELIMITER, StorageConstants.DEFAULT_FIELD_DELIMITER);
+    opt.set(StorageConstants.TEXT_NULL, "\\\\N");
```
jinossy: Looks like you are not apply default null character in meta.

jihoonson: Fixed. Thanks.

jinossy:
+1
The change looks great to me.  Thanks for you effort!

jihoonson:
Thanks for your review!

PR #1014, opened 2016-05-10
TAJO-2145: Error codes based on errno.h need prefix

eminency:


hyunsik:
How about the prefix 'SYS'? Actually, those error codes were designed to indicate system call errors.

hyunsik:
+1 LGTM!

PR #1016, opened 2016-05-11
TAJO-2129: Apply new type implementation to Schema and Catalog.

hyunsik:
See https://issues.apache.org/jira/browse/TAJO-2129.


tajo-algebra/src/main/java/org/apache/tajo/algebra/ColumnDefinition.java
```diff
@@ -51,6 +46,11 @@ public ColumnDefinition(String columnName, DataTypeExpr dataType) {
     if (dataType.isMapType()) {
       this.mapType = dataType.mapType;
     }
+
+    // map type
```
jihoonson: Wrong comment.

tajo-algebra/src/main/java/org/apache/tajo/algebra/DataTypeExpr.java
```diff
@@ -59,15 +68,34 @@ public String getTypeName() {
...
+    return !isArrayType()&& !isRecordType() && !isMapType();
+  }
+
+  public boolean isArrayType() {
+    return arrayType != null;
   }
 
   public boolean isRecordType() {
-    return this.typeName.equals(Type.RECORD.name());
+    return recordType != null;
```
jihoonson: Will this method be frequently called? It seems not, and I think it would be better to check the type name because we can figure out when the type is record but ```recordType``` is null.

tajo-catalog/tajo-catalog-common/src/main/java/org/apache/tajo/catalog/statistics/TableStats.java
```diff
@@ -85,13 +88,17 @@ public TableStats(CatalogProtos.TableStatsProto proto) {
...
-      if (colProto.getColumn().getDataType().getType() == TajoDataTypes.Type.PROTOBUF) {
+      if (peekType(colProto.getColumn().getType()) == PROTOBUF) {
         continue;
       }
       columnStatses.add(new ColumnStats(colProto));
     }
   }
 
+  private static TajoDataTypes.Type peekType(TypeProto proto) {
+    return proto.getElements(proto.getElementsCount() - 1).getKind();
```
jihoonson: Would you share what the idea is behind getting the last element?

tajo-core-tests/src/test/java/org/apache/tajo/engine/eval/TestEvalTree.java
```diff
@@ -385,4 +384,34 @@ private void assertCloneEqual(EvalNode eval) throws CloneNotSupportedException {
...
+        new FieldEval("default.n2.n_regionkey", TypeFactory.create(INT4))
+    );
+
+    EvalNode e4 = new BinaryEval(EvalType.EQUAL,
+        new FieldEval("default.n1.n_comment", TypeFactory.create(TEXT)),
+        new FieldEval("default.n2.n_comment", TypeFactory.create(TEXT))
+    );
+
+    FieldEval f1 = new FieldEval("default.n1.n_nationkey", TypeFactory.create(INT4));
+    System.out.println(f1.hashCode());
```
jihoonson: Should be hash code comparison instead of printing them.

jihoonson:
@hyunsik, nice work! I really like some clean simplified implementations related field, schema, and identifier. I left some trivial comments.
In addition, how about making ```SchemaLegacy``` and ```Column``` classes deprecated?

tajo-algebra/src/main/java/org/apache/tajo/algebra/ColumnDefinition.java
hyunsik: Thank you for good finding. I removed the comments because the code itself is very intuitive.

tajo-catalog/tajo-catalog-common/src/main/java/org/apache/tajo/catalog/statistics/TableStats.java
hyunsik: A TypeProto consists of a number of TypeElements, sorted in a depth-first order of nested types. The last element means its actual type instead of nested or children types. For example, a map type ```map<text, int>``` will be stored in the list of TypeElement as follows: ```[TEXT, INT, MAP]```. ```peekType()``` will just return ```MAP```.

tajo-core-tests/src/test/java/org/apache/tajo/engine/eval/TestEvalTree.java
hyunsik: Actually, I made this test for debugging. I removed the test and inserted the comparison of hash code to TestEvalTree::assertCloneEqual, which is widely used in TestEvalTree.

hyunsik:
Sure, I jut added ```@deprecated``` to SchemaLegacy and Column classes.

tajo-algebra/src/main/java/org/apache/tajo/algebra/DataTypeExpr.java
hyunsik: Thank you for the comment. I fixed it.

hyunsik:
I reflected your comments. Thank you for the comments.

jihoonson:
Thanks for quick update.
+1, ship it!

jihoonson:
Would you close this pr?

PR #1017, opened 2016-05-14
TAJO-2154: Refactor Datum to use new Type implementation.

hyunsik:


jihoonson:
Would you rebase please?

hyunsik:
rebased. Thank you.

tajo-common/src/main/java/org/apache/tajo/type/Type.java
```diff
@@ -36,6 +36,7 @@
 
   // No paramter types
   public static final Any Any = new Any();
+  public static final Bit Bit = new Bit();
```
jihoonson: How about making the modifier of the constructors of these singleton instances to be ```default```?
This is not the changes made in this patch, but it seems better.

jihoonson:
Hi @hyunsik, thank you for your patch. It looks good to me.
I left a comment, but it is not directly related to the changed made in this patch.
Also, I believe you are going to do remaining works to support complex data types. So, if you want to consider my comment in another issue, please do that.

tajo-common/src/main/java/org/apache/tajo/type/Type.java
hyunsik: IMO, the main difference between singleton with (typical) ``getInstance`` method and ```static final``` variable is the initialization time. Singleton pattern is preferred if the variable needs to be initialized lazily. Static variable is initialized at class loading. In this case, they are always used throughout a system. So, IMO, static variable is also right choice here. In addition, static variable provides more concise usages.

tajo-common/src/main/java/org/apache/tajo/type/Bit.java
```diff
@@ -0,0 +1,27 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.tajo.type;
+
+import static org.apache.tajo.common.TajoDataTypes.Type.BIT;
+
+public class Bit extends Type {
+  public Bit() {
```
jihoonson: I mean, this constructor is used only in Type. Its modifier doesn't have to be ```public```, and it will also reduce contributors' mistakes.

tajo-common/src/main/java/org/apache/tajo/type/Type.java
jihoonson: Sorry for the confusion. I'm talking about the modifier of the constructor of the classes which are used in only singleton pattern. I left another comment.

tajo-common/src/main/java/org/apache/tajo/type/Bit.java
hyunsik: I got your point. I misunderstood your comment. I'll change them. Thank you!

jihoonson: Thank you for your understanding. When you update your patch, I'll finish my review immediately.

hyunsik:
thank you for your review. I fixed them.

hyunsik:
fixed.

jihoonson:
+1 LGTM! Ship it!

hyunsik:
Thank you very much. I'll commit it soon.

PR #1020, opened 2016-05-16
TAJO-1952: Implement PartitionFileFragment

blrunner:
See https://issues.apache.org/jira/browse/TAJO-1952 and https://github.com/apache/tajo/pull/846.
I recreated a pull request because of a branch deletion mistake.


blrunner:
Not ready to review.  After recent rebase, some unit tests fail. I'm trying to fix it.

blrunner:
At first, I approached to make another fragment for partitioned table. But now, the only difference between ``FileFragment`` and ``PartitionFileFragment`` is the partition key. So, I would prefer to add the variable to ``FileFragment`` instead of making ``PartitionFileFragment``. It will decrease unnecessary modifications and will avoid side effects by ``PartitionFileFragment``. Here are codes by above approach.  https://github.com/apache/tajo/compare/master...blrunner:TAJO-1952-2?expand=1.

PR #1023, opened 2016-05-16
TAJO-2158: The concat_ws function can't support a tab separator.

combineads:
When I executed below SQL in Tajo.
default> select concat_ws('\t','asad','combine');

?concat_ws

asad\tcombine
(1 rows, 0.055 sec, 0 B selected)
When I executed it in Hive.
asad combine
When I executed it in MySQL.
asad combine
I hope Tajo will be support '\t' character is tab.
Thank you.


jihoonson:
+1 LGTM!

PR #1026, opened 2016-05-17
TAJO-2027: Writing Hive UDF integration document

eminency:


tajo-docs/src/main/sphinx/functions/hivefunc.rst
```diff
@@ -0,0 +1,81 @@
...
+
+=============
+Function Name
+=============
+
+Tajo reads hive functions override ``org.apache.hadoop.hive.ql.exec.UDF`` class. Function name is used as specified in
+``@Description`` annotation. If it doesn't exist, Tajo uses full qualified class name as function name. For example,
+it can be like this : ``select com_example_hive_udf_myupper('abcd')``, so it is recommended to use Description annotation.
+
+And if some function conflict occurs, it may throw ``AmbiguousFunctionException``. This conflict means about function signature,
```
hyunsik: Does the 'conflict' mean duplication?

tajo-docs/src/main/sphinx/functions/hivefunc.rst
```diff
@@ -0,0 +1,81 @@
...
+##############
+
+Tajo provides a feature to use Hive functions directly without re-compilation or additional code.
+
+*************
+Configuration
+*************
+
+Only thing to do is registering path to a directory for jar files containing your hive functions.
+You can do this by set ``tajo.function.hive.code-dir`` in ``tajo-site.xml`` like the following.
```
hyunsik: ```hive.jar-dir``` would be more proper for this config. I know this is out of scope of this issue. But, we need to improve it before releasing.

tajo-docs/src/main/sphinx/functions/hivefunc.rst
```diff
@@ -0,0 +1,81 @@
...
+
+And if some function conflict occurs, it may throw ``AmbiguousFunctionException``. This conflict means about function signature,
+not only about function name.
+
+============================
+Parameter type / Return type
+============================
+
+Hive uses *Writable* type of Hadoop in functions, but Tajo uses its internal *Datum* type.
+Because Tajo doesn't support a kind of pluggable type system yet, only some Writable types are supported currently by internal converting.
```
hyunsik: I think the statement 'Because Tajo doesn't support a kind of pluggable type system yet' is unnecessary. IMO, it may cause other misunderstanding.

tajo-docs/src/main/sphinx/functions/hivefunc.rst
eminency: ok

tajo-docs/src/main/sphinx/functions/hivefunc.rst
eminency: Right, I will fix it.

tajo-docs/src/main/sphinx/functions/hivefunc.rst
eminency: Yes, that word would be better.

eminency:
@hyunsik , it's done. Please check.

tajo-docs/src/main/sphinx/functions/hivefunc.rst
hyunsik: If so, DuplicateFunctionException would be proper for this case. Anyway, it is also out of scope of this issue.

hyunsik:
+1 the patch looks good to me. I'll commit it soon.

hyunsik:
I also updated the documentation. But, I missed 'close' tag in the commit log. Could you close this ticket?

hyunsik:
See http://tajo.apache.org/docs/devel/functions/hivefunc.html

eminency:
Thanks, I close this one.

PR #1027, opened 2016-05-17
TAJO-2156: Create GeoIP functions taking various types instead of INET4 type 

eminency:


jinossy:
This changes should be added tests.
If you change GeoIP to GeoIP2, it would be better for tests.
please see the lincese and csv database for testing

http://maxmind.github.io/GeoIP2-java
https://dev.maxmind.com/geoip/geoip2/geoip2-city-country-csv-databases

eminency:
Hi, @jinossy 

Other geoip functions(old ones) have not had a test.
I think it is because of license problem - GeoIP library is set as 'provided' in pom.xml, so test code doesn't work.

Is that wrong?

jinossy:
The Initial code did not add a tests. because geoip is GPL.
Now, geoip function is breaking changes. we should add a tests.
GeoIP2 is Apache 2.0 license. so we can add the module and tests in tajo binary

eminency:
Oh, I got it.

jinossy:
@eminency 
Geoip2 seems to need to add a mock DataReader.
If we add a geoip2 function by new jira issue, It would be better. What do you think?

eminency:
Good suggestion. I will separate it.

eminency:
Separated issues and refined code.

jihoonson:
This patch looks good to me.
@jinossy do you have any other comments?

jinossy:
+1 LGTM

jihoonson:
+1 ship it!

eminency:
Thanks for the review.
I'll commit.

PR #1028, opened 2016-05-18
TAJO-1968: Change 'threshold' in the name of configuration and session variables to 'limit'

seian:


seian:
FYI, I changed a variable name which was `selectivityThreshold` to `selectivityLimit` in `AccessPathRewriter.java` , too. There was a contextual mismatch.

tajo-common/src/main/java/org/apache/tajo/SessionVars.java
```diff
@@ -31,7 +31,7 @@
 
   // Common Suffix Naming Rules:
   //
-  // * LIMIT   - We use the suffix 'LIMIT' if the variable is threshold. So, if some value is greater or less than
+  // * LIMIT   - We use the suffix 'LIMIT' if the variable is limit. So, if some value is greater or less than
```
jihoonson: In this context, ```threshold``` is more appropriate.

tajo-common/src/main/java/org/apache/tajo/SessionVars.java
```diff
@@ -145,7 +145,7 @@
 
   // for index
   INDEX_ENABLED(ConfVars.$INDEX_ENABLED, "index scan enabled", DEFAULT),
-  INDEX_SELECTIVITY_THRESHOLD(ConfVars.$INDEX_SELECTIVITY_THRESHOLD, "the selectivity threshold for index scan",
+  INDEX_SELECTIVITY_LIMIT(ConfVars.$INDEX_SELECTIVITY_LIMIT, "the selectivity limit for index scan",
```
jihoonson: Why did you change the description as well? 
```threshold``` is more appropriate because it has a narrow meaning than ```limit```.

jihoonson:
Hi @seian, thank you for your patch.
It doesn't matter to change variable names if there are no problem for contributors to read code. 
However, documents and comments should provide clear information. So, I left a comment about changes in documents and comments. Please share why you changed them as well.

tajo-common/src/main/java/org/apache/tajo/SessionVars.java
seian: Oops, it seems there was a misunderstanding for me.
According to Hyunsik's comment(please refer to: https://github.com/apache/tajo/pull/848#discussion_r44101976)

> I agreed. Threshold may be too long for human to put on cli. Actually, there mainly are two kinds of suffixes: SIZE and LIMIT. SIZE is literally used to represent size matter. LIMIT has been used to represent some number to determine whether doing or not doing according to the number. It would be better to make them consistent.

I was focusing on the consistency and thought all the terms should be matched using only `LIMIT` and `SIZE`.

seian:
@jihoonson, I understand what your intention is. I will revert the terms in the documentation and upload the patch again. Thanks for your review.

PR #1029, opened 2016-05-20
TAJO-2160: Implement string_agg function.

combineads:
I'll implement to concatenate the strings of a field within a group by query.
I have a table:
ID COMPANY_ID EMPLOYEE
1           1                  Anna
2           1                   Bill
3           2                 Carol
4           2                 Dave

and I wanted to group by company_id to get something like:

SELECT company_id, string_agg(employee, ', ')
FROM mytable
GROUP BY company_id;

COMPANY_ID EMPLOYEE
1                         Anna, Bill
2                      Carol, Dave

PostgreSQL supports string_agg function.

http://www.postgresql.org/docs/current/static/functions-aggregate.html#FUNCTIONS-AGGREGATE-TABLE

Thank you.

### DIFF
--- a/tajo-core/src/main/proto/InternalTypes.proto
+++ b/tajo-core/src/main/proto/InternalTypes.proto
@@ -45,3 +45,8 @@ message CorrProto {
   required double yvar = 5;
   required double covar = 6;
 }
+
+message StringAggProto {
+  required string value = 1;
+  required string delimiter = 2;
+}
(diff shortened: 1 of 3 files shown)


tajo-core/src/main/proto/InternalTypes.proto
```diff
@@ -45,3 +45,8 @@ message CorrProto {
   required double yvar = 5;
   required double covar = 6;
 }
+
+message StringAggProto {
+  required string value = 1;
```
jinossy: IMO, this proto is unnecessary define. the StringAgg is only to aggregate a value.
What do you think about that?

jinossy: Oops. I’m wrong. delimiter should send to context for merge phase.

combineads: It needs to keep the second parameter, delimiter, in the merge method of StringAgg  when the second stages execute to merge the result of fisrt stage of StringAgg.

jihoonson:
This patch looks good to me. @jinossy do you have more comments?

jinossy:
@jihoonson 
No. +1  The patch looks good to me.
@combineads 
Thanks for your contribution!

PR #1030, opened 2016-05-24
TAJO-1968: Change 'threshold' in the name of configuration and session variables to 'limit'

seian:


seian:
I updated the term `threshold` to `limit` but didn't change description on the documentation because it may make users confused.

jihoonson:
Ok. I'll review soon.
BTW, you don't have to close and make a new PR to update your patch. When you commit your patch to your repository, github automatically detects the changes and shows them in the existing PR.
Actually, this is the strongly encouraged way to update contributors' patch because previous discussions can be kept.

seian:
Thanks for the tip. What should I do if I have to change the whole code in my branch in case of some problems being out there and do code from scratch again? Do I have to do just rollback?

jihoonson:
Yes, it doesn't matter whatever you do using git commands. I think that there is only one exception, that is your local branch is accidentally broken, so you cannot continue your work on that branch.

tajo-common/src/main/java/org/apache/tajo/SessionVars.java
```diff
@@ -94,10 +94,10 @@
       Validators.bool()),
 
   // for distributed query strategies
-  BROADCAST_NON_CROSS_JOIN_THRESHOLD(ConfVars.$DIST_QUERY_BROADCAST_NON_CROSS_JOIN_THRESHOLD,
+  BROADCAST_NON_CROSS_JOIN_THRESHOLD(ConfVars.$DIST_QUERY_BROADCAST_NON_CROSS_JOIN_LIMIT,
       "restriction for the total size of broadcasted table for non-cross join (kb)", DEFAULT, Long.class,
       Validators.min("0")),
-  BROADCAST_CROSS_JOIN_THRESHOLD(ConfVars.$DIST_QUERY_BROADCAST_CROSS_JOIN_THRESHOLD,
+  BROADCAST_CROSS_JOIN_THRESHOLD(ConfVars.$DIST_QUERY_BROADCAST_CROSS_JOIN_LIMIT,
```
jihoonson: The name of the above two session variables also need to be changed.

jihoonson:
Hi @seian, sorry for late review.
I left one comment.

PR #1033, opened 2016-05-27
TAJO-2164: SequenceFile print wrong values with TextSerializerDeserializer

blrunner:
See https://issues.apache.org/jira/browse/TAJO-2164

### DIFF
--- a/tajo-storage/tajo-storage-hdfs/src/main/java/org/apache/tajo/storage/sequencefile/SequenceFileScanner.java
+++ b/tajo-storage/tajo-storage-hdfs/src/main/java/org/apache/tajo/storage/sequencefile/SequenceFileScanner.java
@@ -41,6 +41,7 @@ import org.apache.tajo.storage.text.ByteBufLineReader;
 import org.apache.tajo.storage.text.DelimitedTextFile;
 import org.apache.tajo.storage.text.TextLineDeserializer;
 import org.apache.tajo.storage.text.TextLineParsingError;
+import org.apache.tajo.util.KeyValueSet;
 
 import java.io.IOException;
 
@@ -120,6 +121,11 @@ public class SequenceFileScanner extends FileScanner {
     }
 
     outTuple = new VTuple(targets.length);
+
+    KeyValueSet options = new KeyValueSet();
+    options.set(StorageConstants.TEXT_DELIMITER, delim);
+    meta.setPropertySet(options);
+
     deserializer = DelimitedTextFile.getLineSerde(meta).createDeserializer(schema, meta, targets);
     deserializer.init();
 
--- a/tajo-storage/tajo-storage-hdfs/src/main/java/org/apache/tajo/storage/sequencefile/SequenceFileScanner.java
+++ b/tajo-storage/tajo-storage-hdfs/src/main/java/org/apache/tajo/storage/sequencefile/SequenceFileScanner.java
@@ -106,7 +106,7 @@ public class SequenceFileScanner extends FileScanner {
     }
 
 
-    // Set value of non-deprecated key for CSVLineSerDe::getFieldDelimiter.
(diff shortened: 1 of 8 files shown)


tajo-storage/tajo-storage-hdfs/src/main/java/org/apache/tajo/storage/sequencefile/SequenceFileScanner.java
```diff
@@ -105,6 +105,11 @@ public void init() throws IOException {
     }
 
     String delim  = meta.getProperty(StorageConstants.SEQUENCEFILE_DELIMITER, StorageConstants.DEFAULT_FIELD_DELIMITER);
+    if (delim == null || delim.isEmpty()) {
+      delim = meta.getProperty(StorageConstants.TEXT_DELIMITER, StorageConstants.DEFAULT_FIELD_DELIMITER);
+    }
+    meta.getPropertySet().set(StorageConstants.TEXT_DELIMITER, delim);
```
jinossy: The property order should be TEXT_DELIMITER, SEQUENCEFILE_DELIMITER and please use default value of getProperty.
Why you set the property ? is there any reason ?

blrunner: If users only set ``SEQUENCEFILE_DELIMITER``, ``TextLineSerDe`` can't delimit records because ``TextLineSerDe`` use ``TEXT_DELIMITER`` as the delimiter. I modified codes again as follows. The value of ``SEQUENCEFILE_DELIMITER`` will only use if ``TEXT_DELIMITER`` doesn't exit and ``SEQUENCEFILE_DELIMITER`` exists only.

jinossy: I think you can change to simplify code like this 
```
TableMeta tableMeta = meta.clone();
    if (!tableMeta.containsProperty(StorageConstants.TEXT_DELIMITER)) {
      tableMeta.putProperty(StorageConstants.TEXT_DELIMITER, tableMeta.getProperty(StorageConstants.SEQUENCEFILE_DELIMITER);
    }
```
also you should consider the null char and you should remove useless code

tajo-storage/tajo-storage-hdfs/src/main/java/org/apache/tajo/storage/sequencefile/SequenceFileScanner.java
```diff
@@ -104,7 +105,15 @@ public void init() throws IOException {
...
-    String delim  = meta.getProperty(StorageConstants.SEQUENCEFILE_DELIMITER, StorageConstants.DEFAULT_FIELD_DELIMITER);
+
+    // Set value of non-deprecated key for compatibility. It will be used at CSVLineSerDe::getFieldDelimiter.
+    KeyValueSet keyValueSet = meta.getPropertySet();
+    if (!keyValueSet.containsKey(StorageConstants.TEXT_DELIMITER)
+      && keyValueSet.containsKey(StorageConstants.SEQUENCEFILE_DELIMITER)) {
+      keyValueSet.set(StorageConstants.TEXT_DELIMITER, meta.getProperty(StorageConstants.SEQUENCEFILE_DELIMITER));
+    }
+
+    String delim  = meta.getProperty(StorageConstants.TEXT_DELIMITER, StorageConstants.DEFAULT_FIELD_DELIMITER);
```
jinossy: remove unused code

blrunner:
Thanks @jinossy ,
I rebases your comments and added more unit tests.

jinossy:
+1 LGTM
I'm sorry for late review

jihoonson:
Jaehwa looks busy, so I'll commit this patch soon.

blrunner:
Thanks guys, I'll ship it tonight.

PR #1034, opened 2016-05-31
TAJO-2148: Implement an example HTTP tablespace

jihoonson:
You can test like below.

```
default> create table github (*) tablespace http_example using ex_http_json with ('path'='2015-01-01-15.json.gz');
OK
default> \d github

table name: default.github
table uri: http://data.githubarchive.org/2015-01-01-15.json.gz
store type: ex_http_json
Options:
	'path'='2015-01-01-15.json.gz'
	'timezone'='Asia/Seoul'

schema: 


default> select type, repo.name, count(*) from github group by type, repo.name;
[=========================================>] 100%  5.98 sec
type,  repo/name,  ?count
IssuesEvent,  vispy/vispy,  1
PushEvent,  haraldsk/puppet-module-nfs,  2
DeleteEvent,  peterzd/buddy-pix,  1
PushEvent,  jgsme/dev,  3
...
```


jihoonson:
I disabled the example http tablespace for http scheme by default because it is not proper for real appliactions, and thus users need to implement their own tablespace for the real applications.
To use this example tablespace, just add the following lines to ```storage-site.json```.

```
"http": {
      "handler": "org.apache.tajo.storage.http.ExampleHttpFileTablespace",
      "default-format": "json"
    }
```

jinossy:
+1 Looks great to me
Would you add a document to wiki for a developer?

jihoonson:
@jinossy, thank you for your review. I'll add a document to wiki.

PR #1035, opened 2016-06-29
TAJO-2168: NullPointerException occurs when a simple query contains a…

jihoonson:
… python udf.


jihoonson:
Fixed test failure.

jihoonson:
I additionally ArrayOutOfBoundsException which occurs while loading python functions. Our parser used for loading python functions is very naive, so it cannot parse complex functions. However, during unit testing, our python codes like controller.py are unexpectedly loaded, which causes ArrayOutOfBoundsException. So I changed the directory name for python functions.

jinossy:
+1 LGTM!
Ship it!

jihoonson:
Thank you for your review.

PR #1036, opened 2016-07-05
TAJO-2165: Add 'ALTER TABLE UNSET PROPERTY' statement to Tajo DDL

dongjinleekr:


jihoonson:
Thank you for your patch. I'll review soon. Would you please check the test failure?

dongjinleekr:
Updated! Thanks for your immediate response.

Followings are additional comments on this issue:

1. During updating the (potentially) broken test, one question arose: **is there any list of properties that should not be unset?** If there are, it seems like we have to implement an additional feature to make them unset-proof. (for example, show error message when the user tries to unset some reserved properties.) Do you have any opinion on that? Or, is it better to query about this problem on dev mailing list?
2. I omitted implementation for HiveCatalogStore - as you remember, it is a deprecated feature. (TAJO-1775)

jihoonson:
Thank you for the quick update. Here are my answers. 

1. Basically, table properties are used for optional settings per table. So, every table property is optional, and tajo should be able to work with default property values if they are absent. 
However, among table properties, timezone is recently changed as a mandatory property of a table. So, currently I think it is the only one which should not be removed. 
The best solution seems to store the timezone value in the TableMeta or TableDesc instead of in the optional KeyValue set. But this would be better to be done in a separate ticket, and it would be enough to force timezone as unsettable here. 

2. The deprecated one is HCatalogStore, and HiveCatalogStore should be implemented. But, as you see, many functions of HiveCatalogStore are not implemented yet. I think it also would be better to do in a separate ticket.

dongjinleekr:
Oh, my mistake. I confused HCatalogStore and HiveCatalogStore.

Let's sum up. The following is the list of tickets that should be opened. Some of them are from this discussion, and the others are what I found during this work. (I think it is better to implement disabling unset timezone feature on separate issue, not here.)

1. [Major] Move timezone property from KeyValue set and disable to unset it. (Related: [TAJO-1185](https://issues.apache.org/jira/browse/TAJO-1185), [TAJO-1241](https://issues.apache.org/jira/browse/TAJO-1241))
2. [Major, Umbrella] Implement missing HiveCatalogStore features.
3. [Trivial] Remove AlterTableNode#clone - it has been commented out.

If you have any opinion, don't hesitate to comment to me. If you merge and close this request without any comments, I will regard it you accepted my proposal. Thanks again for your guidance! :)

tajo-catalog/tajo-catalog-common/src/main/java/org/apache/tajo/catalog/CatalogUtil.java
```diff
@@ -552,6 +552,16 @@ public static AlterTableDesc setProperty(String tableName, KeyValueSet params, A
     return alterTableDesc;
   }
 
+  public static AlterTableDesc unsetProperty(String tableName, String[] propertyKeys, AlterTableType alterTableType) {
```
jihoonson: ```alterTableType``` must be ```UNSET_PROPERTY```, so the last parameter looks unnecessary.

jihoonson:
Hi @dongjinleekr, I looked over your patch. It looks good to me in overall. I left a single trivial comment.

tajo-catalog/tajo-catalog-common/src/main/java/org/apache/tajo/catalog/CatalogUtil.java
dongjinleekr: I know. But I did it for the consistency with the other AlterTable methods: CatalogUtil#renameColumn, renameTable, addNewColumn and setProperty. How about clear this issue with AlterTableNode#clone? These methods worth to be refined.

tajo-core-tests/src/test/java/org/apache/tajo/engine/query/TestAlterTable.java
```diff
@@ -88,6 +89,36 @@ public final void testAlterTableSetProperty() throws Exception {
...
+    assertNotNull(tableMeta.getProperty("text.null"));
+    assertEquals(tableMeta.getProperty("text.delimiter"), "\\u002b");
+
+    executeDDL("alter_table_unset_property_delimiter.sql", null);
+
+    tableDesc = catalog.getTableDesc(tableName);
+    tableMeta = tableDesc.getMeta();
+    assertEquals(tableMeta.getPropertySet().size(), 2);
+    assertNotNull(tableMeta.getProperty("timezone"));
+    assertNotNull(tableMeta.getProperty("text.null"));
```
jinossy: You should verify the unset property.

tajo-catalog/tajo-catalog-common/src/main/java/org/apache/tajo/catalog/CatalogUtil.java
jihoonson: Right. All those methods need to be refined. But, IMO, they are all internal APIs, and don't have to be consistent. So, I think that the new method has a better signature.

tajo-catalog/tajo-catalog-common/src/main/java/org/apache/tajo/catalog/AlterTableDesc.java
```diff
@@ -51,6 +54,8 @@
   @Expose
   protected KeyValueSet properties;
   @Expose
+  protected Set<String> propertyKeys;
```
jihoonson: Since there is no additional descriptions on this variable, it would be better if it has a more intuitive name like ```unsetPropertyKeys```.

dongjinleekr:
@jihoonson @jinossy Thanks for your comments. I applied all your comments. (You can see the link to the comment in the log message of each commit.) Thanks for reviewing!

dongjinleekr:
As soon as you accept this pull request, I will request another one - the fix for [the glitches discussed in this issue](https://issues.apache.org/jira/browse/TAJO-2175). I already completed that work. :)

tajo-algebra/src/main/java/org/apache/tajo/algebra/AlterTable.java
```diff
@@ -43,6 +44,8 @@
   private AlterTableOpType alterTableOpType;
   @Expose @SerializedName("TableProperties")
   private Map<String, String> params;
+  @Expose @SerializedName("TablePropertyKeys")
+  private List<String> propertyKeys;
```
jihoonson: It would be better to change this variable to be more intuitive as well. I think getter and setter also need to be changed together.

tajo-catalog/tajo-catalog-common/src/main/proto/CatalogProtos.proto
```diff
@@ -321,6 +322,7 @@ message AlterTableDescProto {
   optional PartitionDescProto partition_desc = 6;
   optional KeyValueSetProto params = 7;
   optional string new_table_path = 8;
+  optional StringListProto property_keys = 9;
```
jihoonson: It would be better to change this variable to be more intuitive as well.

tajo-plan/src/main/java/org/apache/tajo/plan/logical/AlterTableNode.java
```diff
@@ -42,6 +42,8 @@
   @Expose
   private KeyValueSet properties = new KeyValueSet();
   @Expose
+  private String[] propertyKeys;
```
jihoonson: It would be better to change this variable to be more intuitive as well. I think getter and setter also need to be changed together.

tajo-plan/src/main/proto/Plan.proto
```diff
@@ -334,6 +335,7 @@ message AlterTableNode {
   optional AddColumn addColumn = 5;
   optional KeyValueSetProto properties = 6;
   optional AlterPartition alterPartition = 7;
+  optional StringListProto propertyKeys = 8;
```
jihoonson: It would be better to change this variable to be more intuitive as well.

jihoonson:
@dongjinleekr thank you for the quick update! I left my last comment. 
Thanks!

jihoonson:
@dongjinleekr thank you for your contribution!

jinossy:
+1
@dongjinleekr Thanks for your contribution!!

PR #1037, opened 2016-07-08
TAJO-2166: Disallow csv file format in DDL Languages

dongjinleekr:


tajo-core-tests/src/test/java/org/apache/tajo/engine/query/TestCreateTable.java
```diff
@@ -97,6 +98,17 @@ public final void testCreateTable2() throws Exception {
     assertDatabaseNotExists("d2");
   }
 
+  @Test
+  public final void testCreateTableWithCsv() throws Exception {
+    try {
+      executeString("CREATE TABLE D1.table1 (age int) using csv;").close();
+      fail("It should not reach here.");
+    } catch (Throwable e) {
+      assertEquals(e.getClass(), SQLSyntaxError.class);
```
jihoonson: How about using JUnit's Expected Exception? You can simply check the expected exception like in https://github.com/apache/tajo/blob/master/tajo-catalog/tajo-catalog-server/src/test/java/org/apache/tajo/catalog/TestCatalogExceptions.java#L65, 
or do more things like described here https://github.com/junit-team/junit4/wiki/Exception-testing.

tajo-core/src/main/java/org/apache/tajo/parser/sql/SQLAnalyzer.java
```diff
@@ -1325,6 +1325,9 @@ public Expr visitCreate_table_statement(Create_table_statementContext ctx) {
 
       if (checkIfExist(ctx.USING())) {
         String fileType = ctx.storage_type.getText();
+        if (fileType.equals("csv")) {
+          throw new TajoRuntimeException(new SQLSyntaxError("Creating csv storage type table is disallowed from 0.12."));
```
jihoonson: It would be better to let users know that they use ```text``` instead of ```csv```. In addition, I think it would be better if a specific version name is not included in the error message for maintenance issue.

dongjinleekr:
I applied all your comments! :)

jihoonson:
+1 Thanks.

PR #1045, opened 2016-08-13
TAJO-2170: Disable unsetting timezone property

dongjinleekr:
@jihoonson Hello. I just completed [TAJO-2170, Disable unsetting timezone property](https://issues.apache.org/jira/browse/TAJO-2170). However, It would be good to explain my decisions made on this issue. In short, it became a little bit different from our initial expectation.

A little bit different from [your comment in TAJO-2165](https://github.com/apache/tajo/pull/1036#issuecomment-230673810), which became the starting point of this feature, I didn't move the timezone property from KeyValueSet to TableMeta or TableDesc. Here is why:

- Our goal is to disable the removal of the timezone property from the catalog. So, it is sufficient to keep this property from removal by raising Error when the user attempts to remove it from the catalog storage; However, KeyValueSet, TableMeta and TableDesc are all components in alter table message, not storage. So, they don't need to be modified.
- If we modify KeyValueSet, TableMeta or TableDesc, we have to reconduct similar task every time some new unremovable properties are added. This repetitive task does not worth.

With above reasons, I decided not to modify KeyValueSet, TableMeta or TableDesc.

### DIFF
--- a/tajo-catalog/tajo-catalog-server/src/main/java/org/apache/tajo/catalog/store/AbstractDBStore.java
+++ b/tajo-catalog/tajo-catalog-server/src/main/java/org/apache/tajo/catalog/store/AbstractDBStore.java
@@ -21,7 +21,9 @@
  */
 package org.apache.tajo.catalog.store;
 
+import com.google.common.collect.Sets;
 import com.google.protobuf.InvalidProtocolBufferException;
+
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
 import org.apache.hadoop.conf.Configuration;
@@ -62,6 +64,7 @@ public abstract class AbstractDBStore extends CatalogConstants implements Catalo
   protected final String connectionPassword;
   protected final String catalogUri;
 
+  protected final Set<String> unremovablePropertySet = Sets.newHashSet("timezone");
   protected final String insertPartitionSql = "INSERT INTO " + TB_PARTTIONS
     + "(" + COL_TABLES_PK + ", PARTITION_NAME, PATH, " + COL_PARTITION_BYTES
     + ") VALUES (?, ? , ?, ?)";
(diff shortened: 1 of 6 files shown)


dongjinleekr:
Add to this, I have another question: is there any reason you separated CreateTableExecutor from DDLExecutor? For all DDL operations except 'Create Table' are executed in DDLExecutor, which invokes RPC operations via CatalogService object. However, 'Create Table' operation is executed in an indirect way, via CreateTableExecutor. I thought there must be some design intention or development history I don't know. If it's no inconvenience to you, I would like to know.

tajo-catalog/tajo-catalog-server/src/main/java/org/apache/tajo/catalog/store/AbstractDBStore.java
```diff
@@ -1110,12 +1113,20 @@ private void setProperties(final int tableId, final KeyValueSetProto properties)
...
 
-  private void unsetProperties(final int tableId, final PrimitiveProtos.StringListProto propertyKeys) {
+  private void unsetProperties(final int tableId, final PrimitiveProtos.StringListProto propertyKeys)
+      throws UnremovableTablePropertyException {
     final String deleteSql = "DELETE FROM " + TB_OPTIONS + " WHERE TID=? AND KEY_=?";
 
     Connection conn;
     PreparedStatement pstmt = null;
 
+	  Set<String> keys = Sets.newHashSet(propertyKeys.getValuesList());
```
jinossy: Please fix indentation in this code block
https://cwiki.apache.org/confluence/display/TAJO/Code+Style
Thanks

dongjinleekr: Fixed!! :)

dongjinleekr:
Fixed!! :)

jinossy:
+1 LGTM!

PR #1046, opened 2016-08-15
TAJO-2179: Add a regular expression scanner and appender.

jinossy:
I refer to the hive for compatibility.


jinossy:
Could anyone review this?

tajo-storage/tajo-storage-hdfs/src/main/java/org/apache/tajo/storage/regex/RegexLineDeserializer.java
```diff
@@ -0,0 +1,167 @@
...
+    int currentTarget = 0;
+    int currentIndex = 0;
+    Matcher m = inputPattern.matcher(line);
+
+    if (!m.matches()) {
+      unmatchedRows++;
+      if (unmatchedRows >= nextUnmatchedRows) {
+        nextUnmatchedRows *= 100;
+        // Report the row
+        LOG.warn("" + unmatchedRows + " unmatched rows are found: " + line);
```
jihoonson: This also looks good, and I'm just curious why this log is printed repeatedly when 100 unmatched rows are found instead of just printing the total number of unmatched rows when the deserializer is released. Is it more helpful?

jihoonson:
+1 looks good to me. Sorry for late review.

tajo-storage/tajo-storage-hdfs/src/main/java/org/apache/tajo/storage/regex/RegexLineDeserializer.java
jihoonson: Never mind. Printing which line is not matched will be much helpful.

jinossy:
Thank you for your review!

PR #1051, opened 2016-10-18
TAJO-2184: Can not run Tajo with non-default $TAJO_CONF_DIR

dongjinleekr:
start-tajo.sh was updated.


eminency:
Hi, @dongjinleekr. Thanks for the contribution.

Though I reviewed briefly, it seems **--config** parameter is passed only to tajo-daemon.sh, not to tajo-daemons.sh.
Could you check it out?

dongjinleekr:
Sure. I already checked it out. As you can see from tajo-daemon.sh and tajo-daemons.sh, only the first one takes --config parameter; the second one works as a ssh-proxy to the worker hosts. By this reason, it is sufficient to pass the parameter only to tajo-daemon.sh. (In detail, tajo-daemons.sh passes the execution command of tajo-daemon.sh to the workers via its parameters, including --config.)

hyunsik:
FYI, as Dongjin said, tajo-daemon.sh is an actual script to be executed at each node. tajo-daemons is a script to call iteratively a given command in all remote nodes.

PR #1052, opened 2017-12-02
[TAJO-2185] Eliminate protoc binary dependency

hyunsik:
See https://issues.apache.org/jira/browse/TAJO-2185.

Building Tajo project depends on protoc binary installed on a local machine. It is very inconvenient because users must install protoc in their local machine. It would be great if we eliminate this dependency. This patch uses [1] to replace the dependency of local binary by protoc-jar.

### DIFF
--- a/tajo-catalog/tajo-catalog-common/pom.xml
+++ b/tajo-catalog/tajo-catalog-common/pom.xml
@@ -75,25 +75,27 @@
         </executions>
       </plugin>
       <plugin>
-        <groupId>org.codehaus.mojo</groupId>
-        <artifactId>exec-maven-plugin</artifactId>
-        <version>1.2</version>
+        <groupId>com.github.os72</groupId>
+        <artifactId>protoc-jar-maven-plugin</artifactId>
+        <version>3.5.0</version>
         <executions>
           <execution>
-            <id>generate-sources</id>
             <phase>generate-sources</phase>
-            <configuration>
-              <executable>protoc</executable>
-              <arguments>
-                <argument>-Isrc/main/proto/</argument>
-                <argument>--proto_path=../../tajo-common/src/main/proto</argument>
-                <argument>--java_out=target/generated-sources/proto</argument>
-                <argument>src/main/proto/CatalogProtos.proto</argument>
-              </arguments>
-            </configuration>
             <goals>
-              <goal>exec</goal>
+              <goal>run</goal>
             </goals>
+            <configuration>
+              <protocVersion>2.5.0</protocVersion>
+              <includeStdTypes>true</includeStdTypes>
+              <inputDirectories>
+                <include>src/main/proto/</include>
+              </inputDirectories>
+              <includeDirectories>
+                <includeDirectory>../../tajo-common/src/main/proto</includeDirectory>
+                <includeDirectory>../../tajo-catalog/tajo-catalog-common/src/main/proto</includeDirectory>
+              </includeDirectories>
+              <outputDirectory>target/generated-sources/proto</outputDirectory>
+            </configuration>
           </execution>
         </executions>
       </plugin>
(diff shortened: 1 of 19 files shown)


hyunsik:
There was travis build error because of the changed travis environment. Travis seems to use busybox where ulimit is not an external command. So, I changed travis.yml file too.

hyunsik:
I found a new bug filed at https://issues.apache.org/jira/browse/TAJO-2186. IMO, It seems not critical. So, I temporarily fixed this unit test issue.

tajo-catalog/tajo-catalog-common/pom.xml
```diff
@@ -75,25 +75,27 @@
...
             </goals>
+            <configuration>
+              <protocVersion>2.5.0</protocVersion>
+              <includeStdTypes>true</includeStdTypes>
+              <inputDirectories>
+                <include>src/main/proto/</include>
+              </inputDirectories>
+              <includeDirectories>
+                <includeDirectory>../../tajo-common/src/main/proto</includeDirectory>
+                <includeDirectory>../../tajo-catalog/tajo-catalog-common/src/main/proto</includeDirectory>
```
jihoonson: Don't have to include the input directory. Guess it's a copy-paste error?

tajo-core/pom.xml
```diff
@@ -93,33 +93,29 @@
...
+            <configuration>
+              <protocVersion>2.5.0</protocVersion>
+              <includeStdTypes>true</includeStdTypes>
+              <inputDirectories>
+                <include>src/main/proto/</include>
+              </inputDirectories>
+              <includeDirectories>
+                <includeDirectory>../tajo-common/src/main/proto</includeDirectory>
+                <includeDirectory>../tajo-catalog/tajo-catalog-common/src/main/proto</includeDirectory>
+                <includeDirectory>../tajo-core/src/main/proto</includeDirectory>
```
jihoonson: Has to be ```../tajo-client/src/main/proto```?

tajo-storage/tajo-storage-hbase/pom.xml
```diff
@@ -112,26 +112,28 @@
...
+            <configuration>
+              <protocVersion>2.5.0</protocVersion>
+              <includeStdTypes>true</includeStdTypes>
+              <inputDirectories>
+                <include>src/main/proto/</include>
+              </inputDirectories>
+              <includeDirectories>
+                <includeDirectory>../../tajo-common/src/main/proto</includeDirectory>
+                <includeDirectory>../../tajo-catalog/tajo-catalog-common/src/main/proto</includeDirectory>
+                <includeDirectory>../../tajo-core/src/main/proto</includeDirectory>
```
jihoonson: This new directory should be included?

tajo-catalog/tajo-catalog-common/pom.xml
hyunsik: Right, I'll remove tajo-catalog-common.

tajo-core/pom.xml
hyunsik: Interestingly, tajo-client/src/main/proto was not necessary. The protobuf files of tajo-client depend on only those of tajo-core, but it wasn't vise versa.

tajo-storage/tajo-storage-hbase/pom.xml
hyunsik: I'll remove the tajo-core.

tajo-core/pom.xml
jihoonson: Ok, then we don't have include the input directory here as well.

hyunsik:
@jihoonson Thanks for your quick review. I reflected your comments.

jihoonson:
No problem. Would you address [this comment](https://github.com/apache/tajo/pull/1052#discussion_r154543103) as well?

tajo-core/pom.xml
hyunsik: I removed that too.

hyunsik:
Thanks. I'll merge it.

PR #1053, opened 2018-03-17
TAJO-2187: Update KEY file and signatures of release artifacts

hyunsik:
This patch updates the signatures of download page.


hyunsik:
I already pushed this change to our web when I submitted this patch. But, this change haven't been committed to our git repository.

hyunsik:
Thank you for your quick review.

PR #1055, opened 2018-07-13
Dictionary encoded text in ORC scanner may cause incorrect result.

jinossy:
See the issue:
https://issues.apache.org/jira/browse/TAJO-2189


hyunsik:
+1
The patch looks good to me.

jinossy:
Thank you for your quick review!

